PR #291, opened 2013-04-23
Added digest package to imports

christophergandrud:
See my previous pull request on _run-source.R_.


ConorIA:
Thanks @chrtophergandrud. I'll make an explicit selection here and hopefully that will fix the problem.

Issue #394, opened 2013-12-11
Error: contains a blank line


kevinushey:
When attempting to install one of our packages with `install_github`, I get the error:

```
Error: contains a blank line
```

From what I can see, it occurs because `devtools` adds some extra `GitHub`-specific information to the `DESCRIPTION` file, but that `DESCRIPTION` file may have more than one newline ending the file. Hence, a blank line is left between the old `DESCRIPTION` fields and the new material added by `devtools`.

I think excessive final newlines should be stripped from the `DESCRIPTION` before modifying it, just to protect users from this error.

PR #439, opened 2014-03-19
Fix for #394

kevinushey:
Hi Hadley,

A simple fix for #394. We read in the `DESCRIPTION` and strip out any blank lines. Not so sure if you want to give a warning though.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # devtools 1.4.1.99
 
+* `install_github()` now removes blank lines found in a package `DESCRIPTION`
+  file, protecting users from the vague `error: contains a blank line` error. 
+  (#394)
+
 * `document()` now requires at least version 3.0.0 of roxygen2.
 
 * `install_github()` now defaults to `dependencies = TRUE` so you definitely


jacopone:
Hi all, apologies if this is not the right way to report this. 

I've installed Rcompression to solve this problem http://r.789695.n4.nabble.com/Decompressing-raw-vectors-in-memory-td4603202.html via install_github("omegahat/Rcompression"). 

It seems that the problem reported here also happens here https://github.com/omegahat/Rcompression/blob/master/DESCRIPTION, and I expected, maybe wrongly, this fix to solve it. 

Or should I remove blank lines from https://github.com/jacopone/Rcompression/blob/master/DESCRIPTION ? 

Thanks

kevinushey:
I was just able to run `install_github("omegahat/Rcompression")` without issue, so at least it seems like the `devtools` side of things are in good shape (using the latest CRAN version of `devtools` + underlying packages).

That said, removing the trailing spaces as you have is an appropriate workaround if you're still seeing the issue.

jacopone:
Yes that works well. The problem really happens when in a project using packrat:
* install_github("omegahat/Rcompression")
* packrat::snapshot()

Then when I restore() it breaks. The workaround I'm using so far is forking Rcompression, removing the spaces and using that version on my project. 

Thanks!

kevinushey:
That seems like a fair workaround. Ultimately it seems like the bug would lie in Packrat though, not devtools (it has its own version of `install_github()` used during restore)

PR #622, opened 2014-10-15
CRAN now requires `Title` field to be title case

bpbond:
Updated defaults in build_description to reflect this. This is a new requirement; see https://github.com/wch/r-source/commit/1f0dcb55368d7509c8123ce0bff5c6ef95a48607


hadley:
Seems to already have been done. Thanks though!

PR #624, opened 2014-10-21
add NO_AUTO_CLEAN to needs_clean

luckyrandom:
Since the default Makefile in R doesn't handle dependencies of header
files, devtools run a clean compiling when a header or Makevars is newer
than the dll.

Based on (http://bruno.defraine.net/techtips/makefile-auto-dependencies-with-gcc/),
I use the following Makevars to handle dependencies in development
environment. It may improve compiling speed significantly when a header
file is modified.

``` make
# DEVTOOLS_NO_AUTO_CLEAN
PKG_CXXFLAGS=-MMD -MP

all: $(SHLIB)

SRC=$(wildcard *.cpp)
DEP=$(SRC:.cpp=.d)
-include $(DEP)
```

To work with the above or similar Makevars, we need a method to suppress
the auto cleaning mechanism of devtools. With this commit, needs_clean
will not trigger cleaning and let the Makevars handle dependencies of
header files, when it see a commented line `# DEVTOOLS_NO_AUTO_CLEAN` in
Makevars.

### DIFF
--- a/R/compile-dll.r
+++ b/R/compile-dll.r
@@ -108,12 +108,33 @@ needs_compile <- function(pkg = ".") {
   source > dll
 }
 
+makevars <- function(pkg = ".") {
+  pkg <- as.package(pkg)
+  srcdir <- file.path(pkg$path, "src")
+  dir(srcdir, "^Makevars.*$", recursive = TRUE, full.names = TRUE)
+}
+
 # Does the package need a clean compile?
 # (i.e. is there a header or Makevars newer than the dll)
+# If the file Makevars includes a commented line like `# DEVTOOLS_NO_AUTO_CLEAN`,
+# then return FALSE and trust the Makevars to handle dependencies of
+# header files.
 needs_clean <- function(pkg = ".") {
   pkg <- as.package(pkg)
 
-  headers <- mtime(headers(pkg))
+  makevars <- makevars(pkg)
+  no_auto_clean <- FALSE
+  for (f in makevars) {
+    if (length(grep("^#+\\s*DEVTOOLS_NO_AUTO_CLEAN\\s*$", readLines(f, warn = FALSE))))
+    no_auto_clean <- TRUE
+  }
+
+  if (no_auto_clean) {
+    # Still check a Makevars newer than the dll
+    headers <- mtime(makevars)
+  } else {
+    headers <- mtime(headers(pkg))
+  }
   # no headers, so never needs clean compile
   if (is.null(headers)) return(FALSE)
 
--- a/R/compile-dll.r
+++ b/R/compile-dll.r
@@ -130,7 +130,7 @@ needs_clean <- function(pkg = ".") {
   }
 
   if (no_auto_clean) {
-    # Still check a Makevars newer than the dll
+    # Run clean compiling when `Makevars` file is modified
     headers <- mtime(makevars)
   } else {
     headers <- mtime(headers(pkg))


R/compile-dll.r
```diff
@@ -108,12 +108,33 @@ needs_compile <- function(pkg = ".") {
   source > dll
 }
 
+makevars <- function(pkg = ".") {
+  pkg <- as.package(pkg)
+  srcdir <- file.path(pkg$path, "src")
+  dir(srcdir, "^Makevars.*$", recursive = TRUE, full.names = TRUE)
```
hadley: I think this would be more convenient if it returned the lines (or `character()`) if no makefile

R/compile-dll.r
```diff
@@ -108,12 +108,33 @@ needs_compile <- function(pkg = ".") {
...
+  makevars <- makevars(pkg)
+  no_auto_clean <- FALSE
+  for (f in makevars) {
+    if (length(grep("^#+\\s*DEVTOOLS_NO_AUTO_CLEAN\\s*$", readLines(f, warn = FALSE))))
+    no_auto_clean <- TRUE
+  }
+
+  if (no_auto_clean) {
+    # Still check a Makevars newer than the dll
+    headers <- mtime(makevars)
```
hadley: Why do we need this?

R/compile-dll.r
luckyrandom: I modify function `headers` to get this function. I thought the returned value should be consistent with `headers`.

R/compile-dll.r
luckyrandom: We should clean and compile all the source files, when Makevars file is modified.

R/compile-dll.r
```diff
@@ -108,12 +108,33 @@ needs_compile <- function(pkg = ".") {
...
+# then return FALSE and trust the Makevars to handle dependencies of
+# header files.
 needs_clean <- function(pkg = ".") {
   pkg <- as.package(pkg)
 
-  headers <- mtime(headers(pkg))
+  makevars <- makevars(pkg)
+  no_auto_clean <- FALSE
+  for (f in makevars) {
+    if (length(grep("^#+\\s*DEVTOOLS_NO_AUTO_CLEAN\\s*$", readLines(f, warn = FALSE))))
```
hadley: Can you please extract out a `has_no_auto_clean()` function that takes a vector of paths, and returns TRUE or FALSE? (It's also slightly nicer to use `grepl()` rather than `grep()`)

hadley:
Could you please add a bullet point to `NEWS.md`.  It might be worth thinking about where else in the docs we should highlight this - maybe in `compile_dll()`?

luckyrandom:
When review my own code, this implementation feels like a back door, as it change the behavior of `compile_dll` by adding comment to `makevars`. To make it more user friendly, we may add a function `use_detect_ dependencies` that generates the `Makevars` file.

An alternative implementation is adding arg `auto_clean` to `compile_dll` and `load_all`. Which way do you prefer?

hadley:
@jjallaire any thoughts on a clean way to turn off devtool's auto-cleaning when a proper makefile is present?

jjallaire:
I'd say the most explicit/direct way to do this would be to:

1. Add the `auto_clean` argument as suggested.

2. Have it default to `getOption("devtools.autoclean', TRUE)` so it can be externally specified on a global or per-project basis.

hadley:
No one else seems to experienced this problem, and the PR has been open for over two years, so I'm just going to close it.

PR #628, opened 2014-10-28
Documentation of load_data() to .RData, not .Rdata

singmann:
According to the R-extension manual data-files should have the extension
.RData and not .Rdata
(http://cran.r-project.org/doc/manuals/R-exts.html#Data-in-packages).
Consistent with that, load_data() only loads .RData files (at least on
windows), but no .Rdata files. I only changed the documentation to be
consistent with that, i.e., specifically refer to .RData files.


hadley:
Thanks!

PR #641, opened 2014-11-07
Execute scripts with the package's root as working directory in load_code()

### LINKED ISSUE #640: Working directory when preparing package for testing in test()
When testing a package using `devtools::test()`, the current working directory for executing the *package's* code ("installing" the package into a temporary environment) seems to be the `R` subdirectory of the package, whereas it's the package's root directory when using `devtools::check()`.

Test package: https://github.com/krlmlr/test.getwd

Output with `devtools::test()` (copied from RStudio's console):
```
In package: /home/muelleki/git/R/test.getwd/R
```

Output with `devtools::check()`: https://github.com/krlmlr/test.getwd/blob/master/test.getwd.Rcheck/00install.out#L4

**EDIT**: It's the same with `devtools::document()`.

Is this by design? Otherwise I'd like to take a look at this, so that the CWD is always the package's root.

This is important when the package wants to access stuff in `inst/` during installation: This seems to work with R, but doesn't when testing using `devtools::test()`.

krlmlr:
- Includes a test

Fixes #640.


hadley:
Can you merge/rebase please?

krlmlr:
Merged.

hadley:
Thanks!

Issue #623, opened 2014-10-19
devtools::load_all() may fail when S4 class definitions are in different files


Geoff99:
Hi
This is similar to something I tried to raise in #550, but in the previous issue I confused two different aspects of my problem when I try to use the devtools package development sequence, so I am trying again, just concentrating on explaining the aspect that causes devtools to fail for me.

If I create a simple example project and add two files in the R directory as follows
a.r is 

```r
#' @include b.r
Class2 <- setClass('Class2',contains='Class1')
```

and b.r is 

```
Class1 <- setClass('Class1')
```

then when I try to run load_all() on this package I get the following error messages

```
> library(devtools)
> load_all()
Loading loadallExample
Error in reconcilePropertiesAndPrototype(name, slots, prototype, superClasses,  (from a.r#2) : 
  no definition was found for superclass “Class1” in the specification of class “Class2”
> 
```

The problem arises because although I have the @include directive in my file a.r, it has not yet been processed, so load_all() attempts to read the code from the files in alphabetical order.

I still get an error if I try and run devtools::document('.'), as seen here

```
> document('.')
Updating loadallExample documentation
Loading loadallExample
Error in reconcilePropertiesAndPrototype(name, slots, prototype, superClasses,  (from a.r#2) : 
  no definition was found for superclass “Class1” in the specification of class “Class2”
> 
```

And because of the error while loading, the document('.') does not finish, and does not update the collate entry in the package DESCRIPTION file.

A solution is to run roxygen2::update_collate() before trying to load_all() or document() my package

```
> roxygen2::update_collate('.')
Updating collate directive in  ./DESCRIPTION 
> load_all()
Loading loadallExample
> 
```

The updated collate entry in the DESCRIPTION file created by roxygen2::update_collate tells load_all() (and all the other devtools machinery) the correct order to process the files in.

I can avoid this problem myself now I know what causes it, but as an absolute novice package creator it really confused me for ages. In my real life example the alphabetic order of the file names was not so obvious, and when I tried to pin the error down by commenting out bits of code, it enabled devtools::document to run successfully.  Then when I removed the commenting out, the original code would behave nicely, making the problem appear only intermittently and very frustratingly.  It would be nice if some fix of the type I have illustrated could be included in the official devtools.

Thanks

Geoff

 PS I'm still pretty much a novice package creator (waiting keenly for the book to be finished and published), just a little more informed about this particular problem.

[...]

PR #649, opened 2014-11-16
Fix and associated tests for issue #623 (S4 and load_all) by calling roxygen2::update_collate

Geoff99:
This proposal is to add a testthat `load_all` is aware of and uses any `@include` directives in the code for the package being loaded. 

**(Update 16/12/14** It also includes the fix itself, I won't bother with the separate pull request for that now)**
It is related to issue #623 and a proposed fix in pull request #648.

Due to my inexperience with git the code from pull request #648 is also included.  If that causes problems, please let me know and I will sort it out.  

Pls see the caveat in PR #648 which should not be applied until issue klutometis/roxygen#302 has been resolved (see PR klutometis/roxygen#303 and PR klutometis/roxygen#304 for proposed fixes). 
***Updated 9 Jan 2015 - the necessary fix to roxygen2 was included in a recent release of roxygen2, so this caveat is now obsolete).***

The testthat uses a copy of the template package (`testCollateOrder`) in `tempdir()` created using a utility function `temp_copy_pkg`.  Working on a copy insulates the test package template from any permanent side effects of `roxygen2::update_collate` (viz changing the DESCRIPTION file). 

I contemplated placing  `temp_copy_pkg` in utils.r, and amending a number of the other testthats in the devtools test suite to work off copies of their test package templates as a safety measure, but decided I was getting (way) too far ahead of myself, and should seek comment and feedback on the pull requests I have submitted before doing anything else.

Geoff


Geoff99:
Temporarily close this pull request while I bring my forked copy of devtools up to date with the latest official development version of devtools

Geoff99:
Hi @hadley 

As promised yesterday.

This pull request contains a `test_that`  `load_all` processes files in the correct collate order.

It also contains a fix for `load_all` itself, so that after `roxygen2::update_collate` has created/refreshed the collate order in the DESCRIPTION file, the `pkg$collate` entry in the `pkg` structure (list) is also refreshed - and hence `load_all` passes the test  :-).

Finally, I have taken the liberty of adding a check that the version of roxygen2 is up to date, since if it is not, the older version of update_collate messes up packages which use a manually created Collate entry.  I found out the hard way that this is particularly *nasty* since the place where I have found most manually generated collate entries is in dummy packages in `test_that` s - and if the collate entry is altered unintentionally, the test no longer does what it did when it was first coded :-( .  The version check is a bit ugly, since there is a bug in the current version of R (3.1.2) which I have been told will be fixed in the next release.  If you'd like me to remove or adjust this version check, just say.

PS The explicit version check seems to be necessary since R 3.1.2 (on my Windows machine) only does version checks on imports packages when it comes across a reference to them in the NAMESPACE file, and `roxygen2` does not appear in the `devtools` NAMESPACE file, since `devtools` uses the `roxygen2::etc   `  syntax.

Geoff

Geoff99:
Correction to   
~~The version check is a bit ugly, since there is a bug in the current version of R (3.1.2) which I have been told will be fixed in the next release.~~ 

I have tidied up the versionCheck code after reading through the bug fix in R itself, and figuring out that the code I had originally written was more elaborate than needed.

R/load.r
```diff
@@ -92,6 +92,18 @@ load_all <- function(pkg = ".", reset = TRUE, recompile = FALSE,
 
   if (!quiet) message("Loading ", pkg$package)
 
+  # Check that the available roxygen2 version has update_collate fixed so it
+  # really does do nothing if there are no @includes
+  roxvc <- list(name = "roxygen2", op = '>=', version = '4.1.0')
```
hadley: This check is unnecessary since devtools now imports roxygen2

tests/testthat/test-load-collate.r
```diff
@@ -24,3 +24,23 @@ test_that("Extra files in collate don't error, but warn", {
...
+  # since update_collate (in load_all) may have permanent side effects,
+  # namely changing the collate field in the DESCRIPTION file
+  test_pkg <- temp_copy_pkg('testCollateOrder')
+  on.exit(unlink(test_pkg, recursive = TRUE))
+
+  expect_message(load_all(test_pkg), "Loading testCollateOrder")
+
+  expect_equal(a, 1) #even though b.r set it to 2
+
+  message("Cleaning up after testCollateOrder")
```
hadley: Please remove this message

Geoff99:
Hi @hadley 

Many thanks for looking over this `testCollateOrder` pull request so quickly.

I've removed the message from the `test_that`, and deleted the explicit version check on `roxygen2`.  

I am still nervous about deleting the roxygen2 version check, because although I cannot build devtools with an outdated version of roxygen2, I can install it in a library which has an outdated version of roxygen2 and run it with that outdated version.  I can supply an example if that would be useful.  It is because of a bug (OK perhaps a design feature) in how base R (as at 3.1.2) does (or doesn't do) dependency checks on Imports packages (dependency checks are **only** applied to DESCRIPTION file Imports packages which are also mentioned in the NAMESPACE file).  I'll accept whatever you decide is best, of course :-)

Geoff

hadley:
I'm not too keen on working around bugs in base R. I think this is unlikely to cause problems in practice.

Geoff99:
Sure, no worries.  Makes sense as a strategy to me.

I have tried raising the issue on the R-Bugs site, but I don't think I'll get a reaction.  I need (lots!) more skill to write **concise** but helpfully precise bug reports.  I'll get better.

Keep up the great work on devtools (and all the other packages).  I really appreciate it (and am learning heaps from seeing your 'style')

Geoff

PS The pull request already has the workaround removed

PR #652, opened 2014-11-18
use explicit qualification for getPackageName

krlmlr:
so that it works in Rscript when methods is not loaded


hadley:
Already done in another PR

PR #702, opened 2015-01-20
Log system information after installing dependencies

krlmlr:


hadley:
What's the use case for this?

krlmlr:
It's nice to have this information when you need it -- to double-check if your local configuration matches that on Travis if test results differ. It's just a few lines in the log that are collapsed by default.

hadley:
Seems like this would be better pushed upstream.

PR #708, opened 2015-01-27
Add default cache argument for lint

jimhester:
Combined with the bugfix in 2fbfee237c39c47cda15e93e96ef74a16b4bc40d this allows use of the caching functionality in lintr.

Reference timings (lints were previously cached)
```r
system.time(devtools::lint('../../../devtools', cache = FALSE))
#> Linting devtools.....................................................................................................................................
#>    user  system elapsed
#>  85.842   0.660  86.692

system.time(devtools::lint('../../../devtools', cache = TRUE))
#> Linting devtools.....................................................................................................................................
#>    user  system elapsed
#>   3.990   0.139   4.187

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -33,5 +33,5 @@ Suggests:
     MASS,
     rmarkdown,
     knitr,
-    lintr
+    lintr (>= 0.2.1)
 License: GPL (>= 2)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # devtools 1.7.1.9000
 
+* `lint()` gains a `cache` argument (@jimhester, #708).
+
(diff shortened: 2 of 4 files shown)


hadley:
Does this need a bump in the version req for lintr?

jimhester:
That is a good point, I will update this

jimhester:
Ok lintr version bumped and requirement added to devtools.

hadley:
And would you mind adding a bit more info about the cache? i.e. where it's saved and how to get rid of it.  Does the cache file needed to be added to `.gitignore` and/or `.Rbuildignore`?

jimhester:
Ok this should be ok to merge now, let me know if the documentation is unclear.

hadley:
Thanks!

PR #712, opened 2015-02-03
Add argument to `install()` and `install_deps()`

jimhester:
This allows you to pass all arguments to `install.packages` if needed.

Motivation for this is to be able to set a alternate repo argument for `install_deps`, for instance to use `devtools::install()` to automatically install dependencies for a bioconductor package.

```r
install(repos = BiocInstaller::biocinstallRepos(), dependencies = TRUE)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # devtools 1.7.1.9000
 
+* `install()` and `install_deps()` gain a `...` argument, so additional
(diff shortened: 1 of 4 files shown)


hadley:
Could you add a note to `NEWS.md` please?

hadley:
(FWIW github only notifies on comments, not commits, so it's useful to add a comment like "PTAL" when done so I see the changes sooner)

PR #720, opened 2015-02-15
Clean up some partial matches.

craigcitro:
(I don't know a sane way to test this, which may make it futile?)


hadley:
There are a few options you can set - I probably should set them somewhere

craigcitro:
i tried doing combinations of `options(warnPartialMatchDollar = TRUE)` for detection, and `options(warn = 2)` to turn those into errors -- but the issue is that there are existing errors in dependencies. i didn't know a way to set "only turn this warning into an error for *this* package".

hadley:
Fix all the dependent packages for me too? :wink:

craigcitro:
well, **all** is a big commitment. :stuck_out_tongue:

PR #724, opened 2015-02-23
adding doxygen in document()

cbarbu:
I wanted to manage doxygen and roxygen together so I wrote this and wondered if you'd be interested.


hadley:
I think this would be better integrated into roxygen as a new roclet. Would you mind moving the issue there?

cbarbu:
That sounds right. Sure!
Le 23 févr. 2015 20:20, "Hadley Wickham" <notifications@github.com> a écrit
:

PR #726, opened 2015-02-26
Update .travis.yml for native R support.

craigcitro:
This modifies the old config to use the native R support.

Note that the old config was careful about ordering of installation in order to ensure that we were using the cloned devtools. I've been more naive about that, but IIUC we're still testing the right thing: https://travis-ci.org/craigcitro/devtools/builds/52301857

If I'm missing something, maybe the right thing to do is to just run a manual `remove.packages` in the `before_script`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,25 +1,27 @@
-language: c
-script: ./travis-tool.sh run_tests
+language: r
+warnings_are_errors: true
 
-after_failure:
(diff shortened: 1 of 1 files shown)


wch:
Very nice! How about also updating the template in `inst/templates`?

craigcitro:
@wch see #727 :wink:

PR #744, opened 2015-03-20
Use complete C replacements for shim functions

jimhester:
This allows them to be replaced globally everywhere they are referenced
rather than just within the package being developed.

The implementation borrows the C functions used in `testthat::with_mock()` so the shim functions are transparently replaced everywhere.

The use case that motivated this pull request was in https://github.com/jimhester/covr/issues/67.  Briefly `BiocGenerics:::testPackage()` uses `system.file(package=x)` to locate the tests for a given package, however the test were not able to be found if the package was loaded with `devtools::load_all()`.  Bioconductor tests are put in `inst/unitTests` by convention, so it would need to use the shim system.file in order to find them when the package is loaded with `load_all()`.  `covr` needs to use `load_all()` so that source code compilation is done with the proper flags.  This change allows `BiocGenerics:::testPackage()` to work transparently regardless of whether the package is installed normally or simply loaded with `load_all()`.

`covr()` could probably be changed to do a regular install of the package in a temporary location, but this change seems like it would make the shim functions behave more consistently throughout, so I think it is a worthwhile change.


R/zzz.r
```diff
@@ -49,5 +49,12 @@ NULL
 
   find_rtools()
 
+  shims <<- list(
```
wch: Does it actually work to assign a value to `shims` when devtools is loaded the normal way (i.e. not with `load_all`)? I would expect that it would be a locked binding, and you would have to use an environment instead of a list.

wch:
It feels unsafe to me to be replacing objects in base packages. A less invasive way would be to attach a new environment as the parent of the global env, with these things in it. Even that feels a bit weird... but if you were to go about it, you could do it like:

```R
shims <- new.env()
base::attach(shims, name = "devtools_shims", warn.conflicts = FALSE)
```


Also, is the C code supposed to be part of this PR?

R/zzz.r
jimhester: It looks like the namespace is not locked until after `.onLoad()` is
called, so it does work with a regular load as well.  I get no errors after
installing the package in this pull request and (in a new session) calling
`library(devtools)`.

My initial pull request used an environment as you describe and populated
the environment when `load_all` was called for the first time, but this
method seems simpler as there is no need for these objects to change after
the package is loaded.

C code missing was a silly mistake, thank you for pointing it out.

The situation this pull request is trying to remedy is when one package
("test_package") is loaded with `load_all()`.  And a second package calls
one of the shim'd functions, e.g. `system.file("unitTests", package =
"test_package")`.  Would what you proposed work in that case?

jimhester:
The warning that is failing the build now is baffling me, https://travis-ci.org/hadley/devtools/builds/55557865#L1559-L1562

It mentions `tools:::httpdPort`, but that function does not seem to appear anywhere in devtools, and definitely wasn't part of my changes.  I will try to figure out what is going on.

jimhester:
Ok so I have finally figured out what is going on with the `tools:::httpdPort` warning.  This occurs in R-3.1.3 because there is a call to `tools:::httpdPort` in the definition to `utils::?` and `utils::help`.  We copy these existing definitions to `shims$help$orig_value` and `shims$?$orig_value` and these definitions are causing the warning.

However the definitions for `utils::help` and `utils::?` in R-devel no longer call `tools:::httpdPort` so this warning does not happen in R-devel, and would not happen in CRANs checks either.

So this code should actually be ok to merge, but it will continue to cause a WARNING on travis because we are not able to test against R-devel.

If you want I can modify `travis.yml` so that  `warnings_are_errors: false` so this builds without failing.

jimhester:
As I mentioned now that travis is using `R 3.2` this now builds without warnings.

I can fix the merge conflicts once you are done with this round of merging.

hadley:
I'd really rather avoid this - I think it's fundamentally dangerous, and it's something that CRAN disapproves of.  I think you're better off doing a regular install in covr.

jimhester:
I did end up doing a regular install as you suggest, I was running into issues using `load_all` with `S4` classes as well trying to use `covr` on Bioconductor packages.

PR #746, opened 2015-03-23
Use bioc_required so installing BiocInstaller works

jimhester:


hadley:
Would you mind adding `sudo: required` while you're in there?

hadley:
Thanks!

PR #747, opened 2015-03-24
Test dots

jimhester:
Pass ellipsis to testthat::test_dir to allow filtering options

As enabled by hadley/testthat#188

This allows you to use any of the options for `grepl()` from `devtools::test()`, `perl = TRUE`, `inverse = TRUE`, ect.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 * `system_check()` combines arguments with ` `, not `, `. (#753)
 
 * `lint()` gains a `cache` argument (@jimhester, #708).
+* `test()` gains an `...` argument so that additional arguments can be passed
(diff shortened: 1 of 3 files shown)


jimhester:
`invert = TRUE` is not an option for `grepl()` (which I forgot), so to get that functionality you would also need https://github.com/hadley/testthat/pull/239 in addition to this pull request.

hadley:
Could you please add a note to NEWS?

jimhester:
I have added a note to NEWS, however the build seems to have stalled.  It should be ok to merge, all I did was rebase on the latest master.

hadley:
Looks good - thanks!

PR #758, opened 2015-04-14
use_testthat adds example test file

krlmlr:


hadley:
Motivation?

krlmlr:
- Example file that can be extended without having to look up `testthat` syntay
- `R CMD check` doesn't start failing

hadley:
From my perspective, it's just a file that I always have to remember to delete. That's why `create()` no longer makes a package doc template.

krlmlr:
How about

```
use_test <- function(name, pkg = ".")
```

similar to `use_vignette`?

hadley:
I like it! I'll close this PR since I don't think there will be much overlap.

PR #760, opened 2015-04-15
install revdep packages manually

krlmlr:
because revdep returns a sorted list, and the packages may depend upon each other if recursive = TRUE


hadley:
I think this needs to be fixed in `check_cran()`, otherwise it'll break the nice reporting provided by `check_cran()`.

Issue #761, opened 2015-04-15
install_github fails if purrr is loaded


robertzk:
As purrr exports `unzip`, which overwrites `utils::unzip`. A full stack trace using [bettertrace](http://github.com/robertzk/bettertrace):

```
> install_github('rstudio/DT')
Downloading github repo rstudio/DT@master
        Error in unzip(src, list = TRUE) : unused argument (list = TRUE)
In package devtools: install_github("rstudio/DT")
In package devtools: install_remotes(remotes, ...)
In package base: vapply(remotes, install_remote, ..., FUN.VALUE = logical(1))
In package devtools: FUN(X[[1L]], ...)
In package devtools: source_pkg(bundle, subdir = remote$subdir)
In package devtools: decompress(path, outdir)
In package devtools: getrootdir(as.vector(unzip(src, list = TRUE)$Name))
In package base: nchar(gsub("[^/]", "", file_list))
In package base: gsub("[^/]", "", file_list)
In package base: as.vector(unzip(src, list = TRUE)$Name)
In package purrr: unzip(src, list = TRUE)

Error: unused argument (list = TRUE)
```

robertzk:
```
> Sys.info()
                                                                                           sysname                                                                                            release
                                                                                          "Darwin"                                                                                           "13.4.0"
                                                                                           version                                                                                           nodename
"Darwin Kernel Version 13.4.0: Sun Aug 17 19:50:11 PDT 2014; root:xnu-2422.115.4~1/RELEASE_X86_64"                                                                      "Roberts-MacBook-Air-3.local"
                                                                                           machine                                                                                              login
                                                                                          "x86_64"                                                                                          "robertk"
                                                                                              user                                                                                     effective_user
                                                                                         "robertk"                                                                                          "robertk"

**Comment by robertzk** [comment]
```
R version 3.0.2 (2013-09-25)
Platform: x86_64-apple-darwin10.8.0 (64-bit)

robertzk:
https://github.com/hadley/devtools/pull/762

robertzk:
My devtools was out of date.

PR #762, opened 2015-04-15
fix unzip ambiguity

robertzk:
An attempt to address https://github.com/hadley/devtools/issues/761 (edit: turns out my devtools was out of date, but the `unzip` utility is gone from devtools so this should be necessary regardless)


hadley:
Thanks!

Issue #763, opened 2015-04-19
Fails to compile on travi.ci...


linearregression:
Travis.ci complains about not able to compile because of sudo privileged.
.travis.yml already set sudo; required.
language: r
warnings_are_errors: true
sudo: required

compile log from travisci:
R for Travis-CI is not officially supported, but is community maintained.
Please file any issues using the following link
  https://github.com/travis-ci/travis-ci/issues/new?labels=community:r
and mention @craigcitro, @eddelbuettel and @hadley in the issue
Installing R
0.05s$ sudo add-apt-repository "deb http://cran.rstudio.com/bin/linux/ubuntu $(lsb_release -cs)/"
sudo: must be setuid root
The command "sudo add-apt-repository "deb http://cran.rstudio.com/bin/linux/ubuntu $(lsb_release -cs)/"" failed and exited with 1 during .
Your build has been stopped.

craigcitro:
@linearregression do you have a link to the build?

linearregression:
First seen the failure:
https://travis-ci.org/linearregression/devtools/builds/59164984 
form just forking
Then after moving sudo required up
you will see  it fails on bioinstaller:
https://travis-ci.org/linearregression/devtools/builds/59165654
Then after setting 
bioc_required: true
it will pass
Look at pull requests:
https://github.com/hadley/devtools/pull/764

jimhester:
FYI https://github.com/hadley/devtools/pull/746 already sets `bioc_required=true`, it just needs to be merged.

hadley:
Fixed now.

PR #764, opened 2015-04-19
Fix https://github.com/hadley/devtools/issues/763

linearregression:
Fix Move sudo: required right after language:r
seem to solve the problem.

but somewhere in devtools forces for bioconductor depdendency.


hadley:
Dup of #746

PR #767, opened 2015-04-21
use custom .Rprofile for calling external instance of R

### LINKED ISSUE #602: "Cyclic dependency check skipped" in release()
When I call `devtools::release()`, at some point during the check process the following message appears:

```
Package has help file(s) containing install/render-stage \Sexpr{} expresssons but no prebuilt PDF manual.
* checking package namespace information ... OK
* checking package dependencies ... NOTE
  No repository set, so cyclic dependency check skipped
* checking if this is a source package ... OK
* checking if there is a namespace ... OK
```

I'm not sure if the `NOTE` about package dependencies is related to the note about the `\Sexpr`-s. With `check(cran = TRUE)` these notes are not shown.

krlmlr:
Worked for me, but a temporary .Rprofile is created for each invocation of R. Perhaps use a file in `inst/`?

Fixes #602.


R/R.r
```diff
@@ -9,7 +9,12 @@ R <- function(options, path = tempdir(), env_vars = NULL, ...) {
     on.exit(set_path(old))
   }
 
-  in_dir(path, system_check(r_path, options, c(r_env_vars(), env_vars), ...))
+  tmp_user_profile <- tempfile("Rprofile")
+  writeLines(text = 'options(repos=c(CRAN="http://cran.rstudio.com"))',
```
hadley: This needs to be pulled from whatever the current option is. I think you'll need `dput()` to handle multiple repos.

R/R.r
```diff
@@ -9,7 +9,12 @@ R <- function(options, path = tempdir(), env_vars = NULL, ...) {
...
+             con = tmp_user_profile)
+
+  in_dir(path, system_check(r_path, options, c(R_PROFILE_USER=tmp_user_profile,
```
hadley: Maybe that should go in `r_env_vars()`?

R/R.r
```diff
@@ -9,7 +9,12 @@ R <- function(options, path = tempdir(), env_vars = NULL, ...) {
     on.exit(set_path(old))
   }
 
-  in_dir(path, system_check(r_path, options, c(r_env_vars(), env_vars), ...))
+  tmp_user_profile <- tempfile("Rprofile")
```
hadley: `file.path(tempdir(), ".Rprofile")` would avoid creating multiple files

R/R.r
krlmlr: Should `r_env_vars()` be responsible for creating the file, too? A faint smell...

krlmlr:
Thanks for the review, I've updated the code and also fixed a minor unrelated style issue in 12e6414.

R/R.r
```diff
@@ -50,10 +51,24 @@ r_env_vars <- function() {
     "TAR" = auto_tar())
 
   if(is.na(Sys.getenv("NOT_CRAN", unset = NA))) {
-    c(vars, "NOT_CRAN" = "true")
-  } else {
-    vars
+    vars <- c(vars, "NOT_CRAN" = "true")
```
hadley: It just occurred to me that this would be nicer as `vars[["NOT_CRAN"]] <- "true"`

R/R.r
```diff
@@ -50,10 +51,24 @@ r_env_vars <- function() {
...
+
+  vars
+}
+
+# Create a temporary .Rprofile based on the current "repos" option
+# and return a named vector that corresponds to environment variables
+# that need to be set to use this .Rprofile
+r_create_repos_profile <- function() {
+  tmp_user_profile <- file.path(tempdir(), "Rprofile-devtools")
+  tmp_user_profile_con <- file(tmp_user_profile, "w")
```
hadley: I'd do `on.exit(close(tmp_user_profile))` here to be a bit safer.

R/R.r
```diff
@@ -50,10 +51,24 @@ r_env_vars <- function() {
...
+    vars <- c(vars, "NOT_CRAN" = "true")
   }
+
+  vars
+}
+
+# Create a temporary .Rprofile based on the current "repos" option
+# and return a named vector that corresponds to environment variables
+# that need to be set to use this .Rprofile
+r_create_repos_profile <- function() {
```
hadley: Maybe this would be better called `r_profile()`?  We might expand it to include other options in the future.

R/R.r
```diff
@@ -50,10 +51,24 @@ r_env_vars <- function() {
...
+  tmp_user_profile <- file.path(tempdir(), "Rprofile-devtools")
+  tmp_user_profile_con <- file(tmp_user_profile, "w")
+  writeLines("options(repos =", tmp_user_profile_con)
+  dput(getOption("repos"), tmp_user_profile_con)
+  writeLines(")", tmp_user_profile_con)
+  close(tmp_user_profile_con)
+
+  c(R_PROFILE_USER=tmp_user_profile)
```
hadley: Space around =

hadley:
Thanks - a few more minor suggestions. I want to make sure that this code is easy to understand, because I think it's quite important.

I think the NEW needs a bit more. Something like: Previously, devtools ran all external R processes with `R --vanilla`. Now it only suppresses user profiles, and constructs a custom Rprofile to override the default.

R/R.r
```diff
@@ -1,6 +1,6 @@
 # R("-e 'str(as.list(Sys.getenv()))' --slave")
 R <- function(options, path = tempdir(), env_vars = NULL, ...) {
-  options <- paste("--vanilla", options)
+  options <- paste("--no-site-file", "--no-environ", options)
```
hadley: I think this also needs `--no-save` and `--no-restore`?

krlmlr: Hm... `--vanilla` doesn't imply `--no-save`, but I agree it's better to add it anyway.

hadley: Are you sure? `R --help` gives me :  `--vanilla	Combine --no-save,
--no-restore, --no-site-file, --no-init-file and --no-environ`

On 4/22/15, Kirill Müller <notifications@github.com> wrote:
>> @@ -1,6 +1,6 @@
>>  # R("-e 'str(as.list(Sys.getenv()))' --slave")
>>  R <- function(options, path = tempdir(), env_vars = NULL, ...) {
>> -  options <- paste("--vanilla", options)
>> +  options <- paste("--no-site-file", "--no-environ", options)
>
> Hm... `--vanilla` doesn't imply `--no-save`, but I agree it's better to add
> it anyway.
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/hadley/devtools/pull/767/files#r28900125


-- 
http://had.co.nz/

krlmlr: Not at all. I was relying on `?Startup` which reads

> The command-line option --vanilla implies --no-site-file, --no-init-file, --no-environ and (except for R CMD) --no-restore

but this refers to startup only and does not seem to list what *else* is implied by --vanilla. Thanks.

PR #768, opened 2015-04-22
Show file size before submitting to CRAN

### LINKED ISSUE #683: Show size of uploaded archive before submission in release()
This just hit me -- some test artifacts were ignored in `.gitignore` but not in `.Rbuildignore` and unnecessarily increased the file size by a large factor.

Perhaps a listing of the directories, too?

(On the other hand, you can always rerun `release()` without clicking on the confirmation link, that arrives by e-mail, so this is probably of minor importance.)

krlmlr:
- extract functions `build_cran` and `upload_cran`

Fixes #683.


R/release.r
```diff
@@ -243,13 +245,21 @@ cran_submission_url <- "http://xmpalantir.wu.ac.at/cransubmit/index2.php"
...
-  maint <- maintainer(pkg)
-  comments <- cran_comments(pkg)
+  built_path <- build_cran(pkg)
+  upload_cran(pkg, built_path)
+}
 
+build_cran <- function(pkg) {
   message("Building")
   built_path <- build(pkg, tempdir(), manual = TRUE)
   message("File size: ", file.info(built_path)$size, " bytes")
```
hadley: I think it would be better to show the size in either kb or (maybe) mb

gaborcsardi: Or make it always pretty: https://github.com/gaborcsardi/prettyunits (sorry, again, self-promotion). If you don't want another dependency, the function is very small, so you can just take it: https://github.com/gaborcsardi/prettyunits/blob/master/R/sizes.r

hadley:
Thanks! 

FWIW, can you please add a comment when you've complete a round of changes? (just PTAL is fine) That way github pings me

PR #773, opened 2015-04-27
add complete function

bquast:
complete a development session by checking if NEWS(.md), version and date are up-to-date, the following steps are included

* the latest git commit message is displayed at the same time as the NEWS(.md)
* the version no. is displayed
* if there is a date, it is checked against today's, if different the date can automatically be updated

the function uses a style similar to release()


R/complete.R
```diff
@@ -0,0 +1,48 @@
...
+    if (yesno("Is NEWS file up-to-date?"))
+      return(invisible())
+  }
+
+  rule("Version number")
+  print(pkg$version)
+  if (yesno("Is the version number up to date?"))
+    return(invisible())
+
+  if ( exists("date", envir = 1) & pkg$date != Sys.Date() ) {
```
hadley: I'm not a big fan of including the `Date` field in the DESCRIPTION, just because it's something you need to keep up-to-date, and doesn't really seem necessary (since a built date is automatically added when you build the package). Do you have a reason you like to use it?

bquast:
I noticed that andere have since stopped including it myself. My reason fort including it was simply that many othet maintainers still do.

I could replace the `yesno`  menu with menu which suggests removing it, with updating it as an alternative. Or would you rather leave it out altogether?

gaborcsardi:
@bquast IMHO it is better to leave it out. It is simply not needed, so why bother?

bquast:
@gaborcsardi so my reason was more that some maintainers are perhaps not aware of it being optional, and in that case and up-to-date field is better than an outdated one. the `if(exists("date", envir = 1)` condition should make sure that anyone who doesn't use this field isn't bothered with this question.

However, if you think that this generally accepted practice, then perhaps it is indeed redundant.

hadley:
I think it might be better to file this as an issue first - I'm not convinced this should be a separate function to `release()`, but I'm happy to discuss more.

PR #776, opened 2015-04-28
Hotfix: demand text from github api

ashander:
Alternative approach to dee7b7

Write an extra utils function that demands text from the github API even with 404. This integrates more easily with testing for this enhancement.

Conflicts:
	R/install-github.r

### DIFF
--- a/R/install-github.r
+++ b/R/install-github.r
@@ -106,10 +106,8 @@ remote_download.github_remote <- function(x, quiet = FALSE) {
   } else {
     auth <- NULL
   }
-
-  x <- httr::HEAD(src_submodules, auth)
-  has_submods <- identical(httr::status_code(x), 200L)
-  if (has_submods)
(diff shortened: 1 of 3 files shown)


hadley:
I think a better approach would be to:

* Write `github_has_remotes()` that returns TRUE or FALSE
* Add positive and negative test on real repo - no mocking

ashander:
+1 seems less brittle, assuming no issue to have real repo hits in the tests and I can think of reliably TRUE and FALSE repos.

I'll make this change shortly unless I hear otherwise. Thanks,




 - Jaime

hadley:
Do you think you could do this by Monday? I'd like to submit to CRAN then

ashander:
Yes, no problem



- Jaime

ashander:
#786 implements this suggestion, so closing

PR #777, opened 2015-04-28
Implement use_wercker

jimhester:
I think I got everything, let me know if you think it needs any changes.


hadley:
Looks good! I'll probably merge once werker responds about requiring full read/write permissions for all public/private repos.

jimhester:
You may want to open an issue about it at https://github.com/wercker/support/issues, not sure which method would get an answer faster.

I agree it seems excessive to ask for that loose of permission up front.

inst/templates/wercker.yml
```diff
@@ -0,0 +1,5 @@
+box: rocker/hadleyverse
```
krlmlr: Do you really want to use a box this large? Wercker saves the box after each build, for hadleyverse this contributes to 1:30 minutes of build time. I've played with creating a smaller base box tailored to wercker: https://github.com/krlmlr/r-wercker-base; the box is based on drd and includes devtools, lintr and covr.

jimhester: This is a fair point, I would actually prefer to just use `rocker/r-base`, but it seems incapable of installing `devtools` because of a missing `curl` dependency.

You think it would be possible to get a working R on https://registry.hub.docker.com/_/busybox/.  2.5 Mb base image is pretty appealing :smile:.

jimhester: Also @krlmlr I wrote a blog post about this setup if you haven't seen it, http://jimhester.com/post/wercker-and-rocker-finally-performant-continuous-integration-for-r.  Your post and repositories were part of my inspiration for working on it, so thanks for blazing the trail!

krlmlr: So far I've built upon Dirk and Carl's efforts, but we might be able to squeeze it just a bit further. Note that my image includes `devtools` and all system deps necessary for it. (It also includes R-devel.)

krlmlr: @jimhester: No, I haven't seen your post until now, but somehow you have found mine ;-) Thanks for the credit. -- Wercker steps are a clean solution, it's great you've found the time to implement them.

hadley:
I'm going to hang off on merging this for this version of devtools because I can't recommend people use it until they fix the overly OAuth scope request

hadley:
I'm going to close. We can re-open once wercker has a less ridiculous policy.

PR #780, opened 2015-04-30
Wrong path

hmalmedal:
The function `use_test()` used the wrong path and would overwrite the test file.


hadley:
Good catch - thanks!

PR #783, opened 2015-05-01
Use full path when checking if file exists

hmalmedal:
When using the functions `use_test()`, `use_package_doc()` and `use_revdep()` from outside the project root, files were overwritten.

Also fix small typo.


hadley:
Thanks!

PR #784, opened 2015-05-03
spell-check package description

krlmlr:
Is active only with `check(cran = TRUE, check_version = TRUE)`. Emits a warning if no spell checker is found (checked via `utils::aspell(NULL)` because the worker function `utils:::aspell_find_program` is not exported).

R source reference: https://github.com/wch/r-source/blob/deb67e31fd0106dae7ab0d274ab874e56ab1b7b5/src/library/tools/R/QC.R#L6517

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -97,6 +97,9 @@
 
 * `NOT_CRAN` envvar no longer overrides externally set variable.
 
+* `check(check_version = TRUE)` also checks spelling of the `DESCRIPTION`; if no
+  spell checker is installed, a warning is given (#784, @krlmlr).
+
 # devtools 1.7.0
 
 ## Improve reverse dependency checking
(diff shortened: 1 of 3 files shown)


hadley:
Hmmmmmmmm, this doesn't seem like it's going to help a lot of people given that aspell isn't usually installed?  (And also maybe it should be used just when `cran = TRUE`?)

krlmlr:
I think it's pretty standard on Debian/Ubuntu, at least it's straightforward to install it (via packages `aspell` and `aspell-en`, tested on rocker). I guess it can be a pain on Windows or even OS X, that's why I made it optional (with warning).

The spell check is part of the "incoming" check (see [R source reference](https://github.com/wch/r-source/blob/deb67e31fd0106dae7ab0d274ab874e56ab1b7b5/src/library/tools/R/QC.R#L6517)), which is activated only with `check_version = TRUE`. Only when R-core chooses to split this logic, it will make sense to include it with `cran = TRUE`.

hadley:
Ah, got it. Can you please add a note to NEWS and to the check docs?

krlmlr:
Updated docs and NEWS. Please `document()`.

hadley:
Thanks!

PR #785, opened 2015-05-03
rename pkg_deps to dev_package_deps in function call

krlmlr:
otherwise `release_checks()` throws an error

### DIFF
--- a/R/check-devtools.r
+++ b/R/check-devtools.r
@@ -21,7 +21,7 @@ check_dev_versions <- function(pkg = ".") {
   message("Checking for dependencies on development versions... ",
     appendLF = FALSE)
 
-  deps <- pkg_deps(pkg, NA)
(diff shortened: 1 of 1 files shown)


R/check-devtools.r
```diff
@@ -21,7 +21,7 @@ check_dev_versions <- function(pkg = ".") {
   message("Checking for dependencies on development versions... ",
     appendLF = FALSE)
 
-  deps <- pkg_deps(pkg, NA)
+  deps <- dev_package_deps(pkg, NA)
   deps <- deps[!is.na(deps$version), , drop = FALSE]
```
hadley: Doesn't this line also need a fix?

krlmlr: Right, and also the line below. The two functions are not interchangeable. I'm closing this pull.

PR #786, opened 2015-05-04
Slight refactor submodule warning to add tests

ashander:
Testing with hadley/devtools as known submodule-free and armstrtw/rzmq
as known submodule-using.

Implements and tests suggestion in  #776


hadley:
Thanks!

PR #787, opened 2015-05-05
Minor update to use_cran_badge

karthik:
@gaborcsardi noted that gh repo name might not always match package name. This small change should fix that.


hadley:
Thanks!

PR #788, opened 2015-05-05
Use with_makevars instead of with_env for with_debug

jimhester:
Previously with_debug would not actually change the compilation flags
because the variables were set in the environment rather than in a
Makevars file.  This change adds a `with_makevars` function which sets
the compilation flags in a temporary Makevars file.

A simple test to see if the previous and new behavior is to look at the compilation flags used when installing devtools.
```r
# in devtools directory
clean_dll()
with_debug(install())
```
In the current master the compilation flags will be the defaults, including `-O3`.

Using this pull the flags will be correctly set to the debug flags.


hadley:
Thanks!

PR #813, opened 2015-05-24
A small format modification

XD-DENG:
In the /argument part of the check.Rd file (in /man folder), I made a small modification on the item "check_version".
The previous code will cause display error in the Page 7 of the PDF manual generated. It was truncated automatically as below:
image

To confirm, you may want to have a check at the online PDF manual: http://cran.r-project.org/web/packages/devtools/devtools.pdf


hadley:
Can you please modify the source in `check.R`?

hadley:
Hmmm, looking at this again, this seems more like a pdf layout problem. I'd rather have it be semantically correct in the sources.

PR #816, opened 2015-05-26
Fix failing test

krlmlr:
Follow-up to #803.

I relied too much on Travis-CI, which didn't to report a now failing test -- and I didn't test locally either.

The test seems to be wrong here, the alternative (call `check_dir` before `create_description` in `load_all`) would imply that you "sometimes" can use a package subdirectory in `load_all` (which you can't if `create_description` is called as fallback and which is now explicitly tested).

Failing test: https://github.com/hadley/devtools/blob/57c93d43663ff411c964712377787fbb72f29014/tests/testthat/test-package.R#L5

Travis-CI issue: https://github.com/travis-ci/travis-ci/issues/3697


hadley:
Thanks!

tests/testthat/test-load.r
```diff
@@ -0,0 +1,10 @@
+context("Loading")
+
+test_that("Package root is working directory when loading", {
+  expect_message(load_all("testLoadDir"), "[|].*/testLoadDir[|]")
+})
+
+test_that("Loading a package subdirectory throws an error", {
```
krlmlr: @hadley What is the point of calling `create_description()` in `load_all()` if the `DESCRIPTION` file is missing? What was the original intention? The `load_all()` fails anyway if `R` and/or `man` is missing.

hadley: Hmmm, I think the goal was to make it easy to bootstrap an existing directory into a package - i.e. you could just drag all your files into an R directory and start using devtools immediately.

krlmlr: But why in `load_all()`?

hadley: I think the goal was to make `load_all()` just work for people, even before they'd formally made a package.

PR #825, opened 2015-05-29
build_vignettes() gains dependencies argument

krlmlr:
Passed to `install_deps()`.

Use case: run `build_vignettes(dependencies = FALSE)` in tests


hadley:
Thanks!

PR #829, opened 2015-06-01
use_data also works with data from the parent frame

krlmlr:
Includes test for overall functionality and for this particular feature.

Closes #816 (=includes it).


krlmlr:
The failing test is fixed in #816.

hadley:
Thanks!

PR #830, opened 2015-06-01
Speed up loading of devtools

krlmlr:
Avoid loading methods package and other heavy dependencies by default.

Loading time: Currently about 1 second, with this pull request: 0.2 seconds.

Git functionality (head method) and `devtools:::r_release()` still work correctly from Rscript.


hadley:
Thanks! The timings were a bit different for me - 0.4s before, and 0.07 after.

PR #831, opened 2015-06-02
export uses_testthat

### LINKED ISSUE #824: Conditionally run tests if there is a testthat infrastructure
I haven't found an "official" way to detect from outside if testing infrastructure is present: `uses_testthat()` is not exported. On the other hand, `test()` fails in non-interactive mode if no testing infrastructure is present. So there seems to be no way to run tests if they exist, and do nothing if they don't. Am I missing anything?

I see two solutions:

1. Export `uses_testthat()`
2. Don't fail in `test()` in non-interactive mode if no infrastructure is present.

Which is better?

krlmlr:
Fixes #824.

The docs are a bit brief (included in the docs for `test()`), let me know if you prefer separate documentation.


hadley:
Thanks!

PR #833, opened 2015-06-04
new functions with_temp_lib and with_temp_libpaths

### LINKED ISSUE #826: with_temp_lib()
- [x] Add helper function
- [x] Document and export
- [x] Refactor uses of `on.exit(.libPaths(...))`

Per https://github.com/hadley/devtools/pull/820#discussion_r31356439.

krlmlr:
- including tests
- tests now use `with_temp_libpaths()` instead of manual `.libPaths()` and `on.exit()`
- fixes related issue with `with_lib()` -- library path was not reset

Fixes #826.


krlmlr:
Merged `master`, checks pass now.

R/with.r
```diff
@@ -128,9 +143,13 @@ set_libpaths <- function(paths) {
   invisible(old)
 }
 
+reset_libpaths <- function(paths) {
```
hadley: Maybe it would be to handle this consistently with the other setter functions, i.e. add a `add` argument?

krlmlr: Do we deprecate `with_libpaths` then?

hadley: That would make sense

krlmlr: There is an inconsistency: with_path will *append* the new path to `PATH` if `add` is set, but with_lib should *prepend*. This inconsistency will be difficult to communicate. Options:

- Prepend to `PATH` if `add` is set
- Use parameter `action = c("replace", "prefix", "suffix")` instead of `add` for with_path and with_lib

Please advise.

While we're at it: The `with` functions are tremendously useful on their own, how about extracting them to a separate package (with less dependencies ;-) )?

hadley: I like the `action` idea. Another `with_` function does the same thing right?

I have no objections to moving these functions into a separate package.

krlmlr:
Closing for now, feature should be part of a new package to which the existing `with_*` functions will be migrated. (What should it be called? `withr`? `scoper`? `set.unset`?)

R/with.r
jimhester: @krlmrlr I started moving the `with_` functions a few months ago, https://github.com/jimhester/robustr, made a few improvements (including more useful function definitions when printing), but I haven't had time to give it a final polish and convert devtools to use it.

krlmlr: @jimhester Thanks for letting me know. New development should continue there (https://github.com/jimhester/robustr/issues/1, https://github.com/jimhester/robustr/issues/2).

PR #835, opened 2015-06-05
work around bogus NOTE

krlmlr:
introduced in #830: https://travis-ci.org/hadley/devtools/builds/64963102#L1720


hadley:
Thanks! I wondered why I had that `@importFrom rversions`

PR #836, opened 2015-06-06
with_lib resets library path

krlmlr:
- Reset library path after `with_lib()`
- Includes test that fails without this modification

Closes #833, I'm planning to implement the `with_temp_lib()` function largely from scratch.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -48,6 +48,8 @@
   reads compilation variables from the Makevars rather than the environment
   (@jimhester, #788).
 
+* Properly reset library path after `with_lib()` (#836, @krlmlr).
+
 # devtools 1.8.0
  
 ## Helpers
--- a/R/with.r
+++ b/R/with.r
@@ -130,9 +130,13 @@ set_libpaths <- function(paths) {
   invisible(old)
 }
 
+reset_libpaths <- function(paths) {
+  .libPaths(paths)
+}
+
 #' @rdname with_something
 #' @export
-with_libpaths <- with_something(set_libpaths)
+with_libpaths <- with_something(set_libpaths, reset_libpaths)
 
 # lib ------------------------------------------------------------------------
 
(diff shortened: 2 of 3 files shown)


hadley:
Could you please squash? (And close #833?)

krlmlr:
Should be good to merge now. I have no idea what's wrong with the coverage -- I've added explicit tests. @jimhester Any idea?

jimhester:
It is failing the "patch coverage" test.  Patch coverage refers to
> The lines changed in each commit must meet this minimum coverage. Recommended value of 90%+. 

https://github.com/codecov/support/wiki/Patch-Coverage-Status

Presumably because you did not add any new tests for the changed lines from this pull request.  Looks like @hadley set it so that 90% of the changed lines need to have coverage to pass.

krlmlr:
Thanks @jimhester for the quick reply. My point is:

- I have added a [test for the new code](https://github.com/hadley/devtools/pull/836/files#diff-5c28a8e2e00a665df9b1f5938388e7c0R44)
- The test calls [`with_lib`](https://github.com/hadley/devtools/pull/836/files#diff-7793392c33c95f57fb11771ab7b11f2aR153) and [`with_libpaths`](https://github.com/hadley/devtools/pull/836/files#diff-7793392c33c95f57fb11771ab7b11f2aR139)
- These call (albeit internally) `set_lib` et al., which are [not covered](https://codecov.io/github/hadley/devtools/R/with.r?ref=20bcb8662ccfdf949d2aa71ed62220513b0a7974#l-144)

jimhester:
Ah I misunderstood the issue, I will try to figure out what is going on
then.

krlmlr:
Thanks. I probably should have been a bit more specific right from the start.

jimhester:
@krlmlr So the reason `set_lib()` was not being covered was because you were forcing the set and reset functions in `with_something()`.  This forces a copy of the functions to be created, and only this copy was being run, not the original function.  Removing those lines shows the proper coverage.

krlmlr:
Removed the `force()` call as suggested by @jimhester and added a test to satisfy codecov.

PR #844, opened 2015-06-10
Improve rtools registry path detection

Kevin-Jin:
First search for the key in HKCU before attempting HKLM in case the user installed Rtools while on a non-admin account.


hadley:
cc @jjallaire

jjallaire:
Thanks! I just made the same change: https://github.com/rstudio/rstudio/commit/be6ba0954062775f13eda69290fb56f9f0608569

One difference is that I scan HKLM first so that the detected version won't change from under existing users (i.e. we'll always do exactly what we did before and then only if we fail entirely will we scan HKCU).

Issue #845, opened 2015-06-11
release() assumes package root is git repository root


mbjones:
Current checks using `git2r` ensure that a package that is in a git repository doesn't have uncommitted changes and that the checkout is clean.  However, the checks in git.R in `git_uncommited()` and similar functions assume that the package is rooted in the top-level repository directory.  For packages that are in a subdirectory contained within a larger git repository, the call to `r <- git2r::repository(path)` returns an error that the directory is not a valid repository:

```
Error in validObject(.Object) : 
  invalid class “git_repository” object: Invalid repository
```

It seems to me that devtools should not require that the package root directory be the same as the git repository root directory. This is easily fixed by adding `discover=T` to the repository check in functions like `git_uncommitted()`, like this:

```r
 r <- git2r::repository(path, discover=T)
```

Does this seem reasonable? I will add a pull request shortly.

PR #846, opened 2015-06-11
Support packages that are not at the root of a git repository.

mbjones:
This fix adds 'discover=T' to git2r:repository() calls in order to allow devtools to locate the repository root directory when the package is in a subdirectory of the overall git repository.  This change should address issue #845.  I have tested this on Mac OS X under R 3.2.0.

I encountered this problem when trying to release() a package that was not at the root of its git repository, and the git_uncommitted() function threw an error as described in #845. After the fix, git_uncommitted() correctly detects the repository root.


R/git.R
```diff
@@ -3,19 +3,19 @@ uses_git <- function(path = ".") {
 }
 
 git_sha1 <- function(n = 10, path = ".") {
-  r <- git2r::repository(path)
+  r <- git2r::repository(path, discover=T)
```
hadley: Could you please use `discover = TRUE`?

hadley:
Thanks - could you please also add a bullet point to NEWS?

mbjones:
Yes, I'll change to use 'TRUE', and add a bullet to news in a revised PR.  More soon...

mbjones:
@hadley I made the changes you requested, and tests passed on my local Mac deployment, as well as in Travis.  I'm not clear on your development workflow -- do you want PRs to fix the codecov issues?  Or does this look good now?

hadley:
codecov is mostly informatively - I'm still experimenting with it.

hadley:
Thanks!

PR #849, opened 2015-06-11
remove_s4_classes() respecting inheritance

### LINKED ISSUE #848: load_all() error when using setOldClass() with inheritance
I have prepared a [MRE](https://github.com/famuvie/testpkg) for this issue, as a dummy package that can be installed and tested quickly.

In summary, I need to register some S3 classes with inheritance structure using `setOldClass()`.
```R
#' @importFrom methods setOldClass
methods::setOldClass(c('derived_class', 'my_class'))
```
When I `load_all()` for the first time, everything goes fine. But for successive times, it fails as follows:
```
Loading testpkg
Error in methods::setOldClass(c("derived_class", "my_class")) (from testmethod.R#3) : 
  inconsistent old-style class information for “derived_class”; the class is defined but does not extend “my_class” and is not valid as the data part
```

**Variations:**
- The error **does not** occur if the class to be registered is single (i.e. `setOldClass('my_class')`).
- Incomprehensibly, it **does not occur** either  if the derived class name is `'splines'` (!?!?)
- Finally, the error **still occurs** if I register `'my_class'` beforehand in a separate sentence.

```
> session_info()
Session info ---------------------------------------------------------------------------------------------------------------
 setting  value                       
 version  R version 3.2.0 (2015-04-16)
 system   x86_64, linux-gnu           
 ui       RStudio (0.99.441)          
 language (EN)                        
 collate  en_US.UTF-8                 
 tz       <NA>                        

Packages -------------------------------------------------------------------------------------------------------------------
 package    * version date       source        
 devtools     1.7.0   2015-01-17 CRAN (R 3.1.2)
 digest     * 0.6.8   2014-12-31 CRAN (R 3.2.0)
 memoise    * 0.2.1   2014-04-22 CRAN (R 3.1.2)
 Rcpp       * 0.11.5  2015-03-06 CRAN (R 3.1.3)
 roxygen2   * 4.1.1   2015-04-15 CRAN (R 3.2.0)
 rstudioapi * 0.3.1   2015-04-07 CRAN (R 3.2.0)
 stringr    * 0.6.2   2012-12-06 CRAN (R 3.1.2)
```

famuvie:
prevents derived classes to remain orphan
Fixes #848

### DIFF
--- a/R/remove-s4-class.r
+++ b/R/remove-s4-class.r
@@ -6,7 +6,7 @@ remove_s4_classes <- function(pkg = ".") {
   pkg <- as.package(pkg)
 
   classes <- methods::getClasses(ns_env(pkg))
-  lapply(classes, remove_s4_class, pkg)
+  lapply(rev(classes), remove_s4_class, pkg)
 }
 
 # Remove an s4 class from a package loaded by devtools
--- a/R/remove-s4-class.r
+++ b/R/remove-s4-class.r
@@ -6,7 +6,28 @@ remove_s4_classes <- function(pkg = ".") {
   pkg <- as.package(pkg)
 
   classes <- methods::getClasses(ns_env(pkg))
-  lapply(rev(classes), remove_s4_class, pkg)
+  lapply(sort_s4classes(classes), remove_s4_class, pkg)
+}
+
+# Sort S4 classes for hierarchical removal
+# Derived classes must be removed **after** their parents.
+sort_s4classes <- function(classes) {
+  for (x in classes) {
+
+    ## classes extended by x
+    ext <- extends(x)
+
+    ## from those, take only defined by the package (i.e. in classes)
+    ## (i.e. leaves out classes like 'oldClass', which are not to be removed)
+    own_ext <- ext[ext %in% classes]
+
+    ## make sure the extended classes are first in the classes vector
+    ## by reversing the order of the extended classes
+    if (length(own_ext) > 1)
+      classes[classes %in% ext] <- rev(own_ext)
+  }
+
+  return(classes)
 }
 
 # Remove an s4 class from a package loaded by devtools
--- a/R/remove-s4-class.r
+++ b/R/remove-s4-class.r
@@ -17,42 +17,44 @@ sort_s4classes <- function(classes) {
 
   sorted_classes <- vector(mode = 'character', length = 0)
 
-  ## classes extended by x which are within the domestic set of classes
-  ## and are not x itself
-  extends_domestic <- function(x, classes) {
+  ## Return the parent class, if any within domestic classes
+  extends_first <- function(x, classes) {
     ext <- extends(x)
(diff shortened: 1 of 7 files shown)


famuvie:
Mmm... contrary to #848, with this reverse removing of classes the error arises when the derived class is named 'splines'. What is it special with this name?

famuvie:
Ok. There is nothing particular with the class name `splines` except for the fact that alphabetically comes after `my_class`. The current `devtools` version removes classes by alphabetical order which happen to coincide with inheritance hierarchy (i.e. `my_class` is extended by `splines`).

I needed to write a small routine to sort the classes defined in the package, so that extended classes are before their respective extensions. With this approach I think the problem is solved.

hadley:
@wch happen to remember an elegant way to do this?

hadley:
Ideally you'd do a topological sort on the S4 dependency graph before deleting

famuvie:
Right. My solution worked for me, but not necessarily in general.
Implemented a topological sorting. Not very fast in R, but graphs of classes should not be huge.

hadley:
Could you please add some unit tests for `sort_s4classes()`?

famuvie:
sure... in a couple of weeks.

famuvie:
Done, with some modifications thanks to the testing.
Incidentally, this fixes one previously failing check that was commented out in `test-s4-unload.R`.

R/remove-s4-class.r
```diff
@@ -6,7 +6,60 @@ remove_s4_classes <- function(pkg = ".") {
...
+
+  ## Return the parent class, if any within domestic classes
+  extends_first <- function(x, classes) {
+    ext <- extends(getClass(x, where = nsenv))
+    parent <- ext[2]
+    classes %in% parent
+  }
+
+  ## Matrix of classes in columns, extending classes in rows
+  extended_classes <- vapply(classes,
```
hadley: Please indent arguments by only two spaces

tests/testthat/test-s4-unload.r
```diff
@@ -46,7 +46,10 @@ test_that("loading and reloading s4 classes", {
   # Check that classes are unregistered
   # This test on A fails for some bizarre reason - bug in R? But it doesn't
```
hadley: Can you please just delete this whole comment block? i.e. 47-51

hadley:
Could you please also add a bullet to NEWS?

tests/testthat/testS4sort/DESCRIPTION
```diff
@@ -0,0 +1,11 @@
+Package: testS4sort
+Title: Test package for sorting S4 classes
+License: GPL-3
```
wch: I think it would be best to just take out the `License` field here. Or it should be the same license as devtools itself.

famuvie:
Done.

hadley:
Thanks!

Issue #810, opened 2015-05-22
build() should call clean_vignettes by default if vignettes = TRUE


krlmlr:
## TL;DR
- [ ] new parameter `clean_vignettes = TRUE` to `build()`
- [ ] update docs for `build_vignettes()`, perhaps mention that adding `inst/doc` to `.Rbuildignore` doesn't work
## Details

The documentation of `build_vignette` [(code ref)](https://github.com/hadley/devtools/blob/d2bdbada3f8921978ab13dbcad5897a60cf5f965/R/vignettes.r#L5) mentions this:

> You need to ensure that these files are not included in the built package...

It would be best if `devtools` just did this:

```
build <- function(..., vignettes = TRUE, clean_vignettes = vignettes) {
  if (clean_vignettes)
    clean_vignettes()
  ...
}
```

Or, even better, move the stuff away during the `build` process and put it back when done.

Otherwise, if `build_vignettes` is called just once, all subsequent updates to the original vignette source will be ignored. (If `inst/doc` is in `.Rbuildignore`, vignettes are built but not included in the `.tar.gz`.) Build artifacts, if ignored by the VCS, should have no impact on the result of subsequent builds!

Per https://github.com/hadley/devtools/commit/9bb78f10ce750d2bd24ca655329e3cbf18cb3f37#commitcomment-11326492.

hadley:
Makes sense to me. Want to do a PR?

PR #855, opened 2015-06-19
RFC: enhanced "vignettes" parameter for build()

krlmlr:
Related to #810. The `vignettes` parameter specifies in detail what should happen with existing files in `inst/doc`.

Not yet implemented, not yet ready to merge. The "stash" would be implemented by moving the files to `inst/doc/.devtools_stash` and restoring afterwards (hoping that the directory doesn't exist yet).

### DIFF
--- a/R/build.r
+++ b/R/build.r
@@ -15,8 +15,18 @@
 #'   the parent directory of the package.
 #' @param binary Produce a binary (\code{--binary}) or source (
 #'   \code{--no-manual --no-resave-data}) version of the package.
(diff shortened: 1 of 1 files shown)


hadley:
I don't think you need quite so many options. I think just implementing update would be fine.

hadley:
I'm going to close because this seems a bit premature to me - I really need to think through how vignettes should behave (and who should be responsible for building - roxygen2 or devtools)

PR #856, opened 2015-06-19
as.package and load_all gain new argument 'create'

### LINKED ISSUE #852: New argument create_if_missing to load_all
If `load_all()` doesn't find a `DESCRIPTION` file in the package directory, it is created. As of #803, parent directories will not be scanned for a `DESCRIPTION` file. #816 explicitly documents and tests this behavior.

Now the behavior of `load_all()` is difficult to understand and communicate: If a directory happens to contain `R`, `src` or `data`, a `DESCRIPTION` is created -- if not (and no `DESCRIPTION` exists), an error is thrown:

``` r
> pkg <- tempfile(); dir.create(pkg); devtools::load_all(pkg)
Error: '/tmp/RtmpqMuf1p/file498b5f0c4c48' does not look like a package: no R/, src/ or data directories
No traceback available 
> pkg <- tempfile(); dir.create(pkg); dir.create(file.path(pkg, "R")); devtools::load_all(pkg)
No DESCRIPTION found. Creating with values:
```

I suggest a new argument `create_if_missing` that governs this behavior. If `TRUE`, always create (perhaps via `devtools::setup()`). If `FALSE`, never create (but use the usual `pkg <- as.package(pkg)` idiom). If `NULL` (the default), the user will be asked if in interactive mode (otherwise same as `FALSE`). 

@hadley: Is this sufficient for the use case you described in  [your comment](https://github.com/hadley/devtools/pull/816/files#r32617415)?

krlmlr:
Governs if a package structure should be created if it is missing. If `NA`, the user is queried (in interactive mode only).

This also changes the semantics of `load_all()`: Now, a package is created only if no `DESCRIPTION` is found in the directory hierarchy, and only if requested or approved by the user. Any `DESCRIPTION` in a parent directoy will be used if found.

All relevant code paths tested.

Fixes #852.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -34,7 +34,9 @@
 
 * `uses_testthat()` and `check_failures()` are now exported (#824, #839, @krlmlr).
 
-* Remove explicit `library(testthat)` call in `test()` (#798, @krlmlr)
+* Remove explicit `library(testthat)` call in `test()` (#798, @krlmlr).
+
+* `as.package` and `load_all` gain new argument `create`. Like other functions with a `pkg` argument, `load_all` looks for a `DESCRIPTION` file in parent directories (#852, @krlmlr).
 
 # devtools 1.8.0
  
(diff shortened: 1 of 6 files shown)


hadley:
I think `setup()` now needs a bit of extra checking to be safe, but I can add that.

Issue #798, opened 2015-05-11
Figure out how to eliminate require(testthat)


hadley:
as requested by CRAN

PR #860, opened 2015-06-23
Fix regressions on win-builder

krlmlr:
Caused by too restrictive tests. Passes the previously failing test on win-builder now:

- R-devel: http://win-builder.r-project.org/B43aMx2dhNzO/
- R-stable: http://win-builder.r-project.org/Dgi7yJvgEZww/

Follow-up to #798.


hadley:
Thanks!

PR #867, opened 2015-07-04
Prepare for CRAN release

krlmlr:
- silent test of test() function
- skip problematic test of check() on CRAN

### DIFF
--- a/tests/testthat/test-test.r
+++ b/tests/testthat/test-test.r
@@ -7,6 +7,8 @@ test_that("Package can be tested with testthat not on search path", {
     on.exit(attach(testthat_env, testthat_pos), add = TRUE)
   }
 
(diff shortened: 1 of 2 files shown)


tests/testthat/test-test.r
```diff
@@ -7,6 +7,8 @@ test_that("Package can be tested with testthat not on search path", {
     on.exit(attach(testthat_env, testthat_pos), add = TRUE)
   }
 
-  test("testTest")
-  test("testTestWithDepends")
+  test("testTest", reporter = "stop")
+  expect_true(TRUE)
```
krlmlr: @hadley Is there anything nicer than this? For some reason, `expect_true(!is.null(test("...", ...)))` does not seem to work in this particular case.

PR #871, opened 2015-07-08
Also suggest rmarkdown package when creating vignette

krlmlr:
This seems to be necessary: https://github.com/yihui/knitr/blob/f38d038561938145fb084c6370ab09b291a87025/man/vignette_engines.Rd#L12-L14


hadley:
Thanks!

Issue #887, opened 2015-08-03
interface `make` [feature-request]


cbeleites:
Would a tiny interface function to call `make` be welcomed?

I mostly use devtools by now, but  I have `make` targets for some tasks such as 

- preparing vignettes where the source is too large and/or calculations are too time consuming to be allowed on CRAN. These are prepared offline; pdfs go with the package but the real source & data is distributed via the project web site.
- the same applies to some tests
- `jekyll` for updating the project web site

famuvie:
I would certainly appreciate it.

hadley:
As I mentioned in the pull request, devtools is already too big, so I think this belongs elsewhere.

PR #891, opened 2015-08-10
Feature: functions to call `make`

cbeleites:
I put up a discussion of this feature as issue [#887](https://github.com/hadley/devtools/issues/887) and concluded there is interest in this functionality.

The feature has 3 functions:

- `make ()` to call make
- `make_gettargets ()` and `make_geteasytargets ()` to retrieve the targets of a `Makefile`.

Please consider adding it to `devtools`.


gaborcsardi:
IMHO, this could be in another package. devtools is already huge and has a lot of functionality that could be refactored into other smaller packages.

hadley:
I'm with @gaborcsardi on this one. I think they're useful functions, but since devtools is already so big, and these functions are not tightly coupled to devtools, I think another package is the way to go.

PR #894, opened 2015-08-13
Add env_var argument to check()

jimhester:
This allows adding or overriding the default environment variables set
by check().

This was needed when I was doing reverse dependencies for Bioconductor packages, `cran = TRUE` sets `_R_CHECK_LIMIT_CORES_ = TRUE`, which fails the check whenever a package uses more than 2 cores in examples or vignettes.

Bioconductor however allows packages to use many cores in examples and vignettes, so I needed a way to set `_R_CHECK_LIMIT_CORES_ = warn` instead.

I added a test to check for the proper behavior, but it required a fairly hacky use of `tryCatch`, `with_mock()` and `stop()` to abort from the function prior to the normal end.  If there is a cleaner way to do this test I would be happy to make the necessary changes.

### DIFF
--- a/R/check.r
+++ b/R/check.r
@@ -116,8 +116,6 @@ check_r_cmd <- function(built_path = NULL, cran = TRUE, check_version = FALSE,
 
   env_vars <- c(check_env_vars(cran, check_version, force_suggests), env_vars)
 
-  # remove duplicated vars, keeping only the last instance
-  env_vars <- env_vars[!duplicated(names(env_vars), fromLast = TRUE)]
   if (!quiet)
     show_env_vars(env_vars)
 
--- a/R/check.r
+++ b/R/check.r
@@ -96,16 +96,11 @@ check <- function(pkg = ".", document = TRUE, cleanup = TRUE, cran = TRUE,
 # Run R CMD check and return the path for the check
 # @param built_path The path to the built .tar.gz source package.
 # @param check_dir The directory to unpack the .tar.gz file to
-<<<<<<< HEAD
-check_r_cmd <- function(name, built_path = NULL, cran = TRUE,
-                        check_version = FALSE, force_suggests = FALSE,
-                        args = NULL, check_dir = tempdir(), quiet = FALSE, ...) {
-=======
 # @param env_vars Optional additional environment variables to set.  Values
 #   set here will override the defaults set by \code{\link{cran_env_vars}()}.
 check_r_cmd <- function(built_path = NULL, cran = TRUE, check_version = FALSE,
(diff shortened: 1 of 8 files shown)


R/check.r
```diff
@@ -109,16 +113,18 @@ check_r_cmd <- function(built_path = NULL, cran = TRUE, check_version = FALSE,
...
   # not set, R CMD check will use the defaults as described in R Internals.
-  env_vars <- c(
+  env_vars <- c(env_vars,
     if (cran) cran_env_vars(),
     if (check_version) c("_R_CHECK_CRAN_INCOMING_" = "TRUE", aspell_env_var()),
     if (!force_suggests) c("_R_CHECK_FORCE_SUGGESTS_" = "FALSE")
   )
 
+  # remove duplicated vars, keeping only the first instance
+  env_vars <- env_vars[!duplicated(names(env_vars))]
```
hadley: This seems backwards to me - normally you'd keep the last instance. Can you please flip around with `fromLast`?

tests/testthat/test-check.r
```diff
@@ -29,3 +29,16 @@ test_that("aspell environment variables", {
     expect_warning(aspell_env_var(), "Skipping spell check")
   )
 })
+
+test_that("setting env_vars overrides defaults", {
```
hadley: I'd rather you test this by pulling `rcmd_check_envars()` out into a separate function.

hadley:
Also needs bullet point for NEWS

hadley:
I just refactor `check_env_vars()` out, so this should be a bit easier now

R/check.r
hadley: In fact, I don't think you even need this anymore - `Sys.setenv() ` will take care of it (and the chances of duplication are now basically zero with new `cran = TRUE` strategy)

hadley:
Do you have time for this? I need it by tomorrow if you want in this release

jimhester:
I can look at it, I think most of it can be removed with your changes though.

hadley:
Yeah, I think it's probably easiest to start from scratch. But threading the arguments through all the function calls is tricky, and I'm likely to make a mistake :/

jimhester:
2a161ae should be good to go I think.  This whole issue is much less of a problem now with the change to setting `--as-cran`.  The user can just do something like
```r
with_env(list(_R_CHECK_LIMIT_CORES_" = "warn"),  check())
```
Whereas previously the above would have been overwritten by devtools internally.

Still it is at least slightly useful to provide an argument to the function, so I guess it can't hurt to merge this.

hadley:
Hmmm, I think I'm swayed by your argument that this is of limited utility since it's now easy to do elsewhere.

PR #895, opened 2015-08-13
Use BiocInstaller repositories if installing a Bioconductor package

jimhester:
If the user specified repositories take precedence over the default
Bioconductor repositories.

This allows one to use `install_github()` with Bioconductor packages and have the dependencies work like any other package. e.g.
```r
install_github("Bioconductor-mirror/S4Vectors")
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # devtools 1.9.0
 
+* Detect if `install_` commands are called on a Bioconductor package and
+  include the Bioconductor repositories if they are not already set (#895,
+  @jimhester).
+
 * Avoid importing heavy dependencies to speed up loading (#830, @krlmlr).
 
 * Remove explicit `library(testthat)` call in `test()` (#798, @krlmlr).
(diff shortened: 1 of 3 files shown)


PeteHaitch:
Very happy to see this being worked on! (And sorry I haven't made any progress myself) 

One question, Jim: are you proposing that dependencies that are from the Bioconductor devel branch are installed from the Bioconductor-mirror github repo rather than using `BiocInstaller::useDevel(TRUE); BiocInstaller::biocLite("pkg")`? The issue I see there is that those packages available from the Bioconductor-mirror repo aren't guaranteed to pass R CMD build/check since they haven't been run through the BioC build system. Apologies if I've misunderstood.

jimhester:
@PeteHaitch The dependencies work with this patch the same way they do with `biocLite()`, they pull from the `BiocInstaller::biocinstallRepos()`, so it will only be packages that have successfully passed through the Bioconductor build system, not from the git mirrors.

PeteHaitch:
Awesome! I thought so (given the code) but your example confused me. Cheers.

hadley:

jimhester:
Done https://github.com/hadley/devtools/commit/406f3dd09410340b391f960f466a367868da77b7

hadley:
Thanks!

Issue #642, opened 2014-11-10
use_github("hadley/abc")


hadley:
* Create new repo on github
* `git init`
* `git push`
* Add URL and BugReports to DESCRIPTION

hadley:
Why did I comment this code out?

```R
  # git2r::branch_set_upstream(git2r::head(r), "origin/master")
  # git2r::push(r, "origin", "refs/heads/master")
```

jennybc:
The recent conversation probably already reminded you ...

You probably commented this out because pushing and/or setting tracking branches with `git2r` is not completely sorted out yet: <https://github.com/ropensci/git2r/issues/133>, <https://github.com/ropensci/git2r/issues/116>, <https://github.com/ropensci/git2r/issues/152>. FWIW I am using v0.11.0 of `git2r` and have still not managed to set the upstream remote with code like you've got above. Caveat: I'm using https. Maybe the alleged fix only applies to ssh?

stewid:
I have managed to add an existing project to GitHub using SSH and https with v0.11.0 of git2r. 

```r
library(git2r)

## Add an existing project to GitHub using SSH and git2r. Create a new
## repository on GitHub before running the following R code.

## 1) Create a new repository
repo <- init("test-01")

## 2) Add 'README.md' to local repository
add(repo, "README.md")

## 3) Commit staged 'README.md'. Add sessionInfo to commit message
commit(repo, message = "Initial commit", session = TRUE)

## 4) Add URL for the remote repository
remote_add(repo, "origin", "git@github.com:stewid/test-01.git")

## 5) Create SSH credentials
cred <- cred_ssh_key("~/.ssh/id_rsa.pub", "~/.ssh/id_rsa")

## 6) Push changes in local repository to GitHub
push(repo, "origin", "refs/heads/master", credentials = cred)
```

```r
library(git2r)

## Add an existing project to GitHub using https and git2r. Create a
## new repository on GitHub before running the following R code.

## 1) Create a new repository
repo <- init("test-02")

## 2) Add 'README.md' to local repository
add(repo, "README.md")

## 3) Commit staged 'README.md'. Add sessionInfo to commit message
commit(repo, message = "Initial commit", session = TRUE)

## 4) Add URL for the remote repository
remote_add(repo, "origin", "https://github.com/stewid/test-02.git")

## 5) Create credentials from environmental variables
cred <- cred_env("GITHUB_USER", "GITHUB_PAT")

## 6) Push changes in local repository to GitHub
[...]

PR #896, opened 2015-08-17
Extend github functionality

jennybc:
This PR works on #642 and adds a few other bits re: GitHub. It is a just a test balloon. If you think this looks promising, I will refine further and add tests for everything I can (but much of this requires auth).

  * Adds unexported function uses_github() to create an explicit test for use in other functions. Previously, github usage was assessed by looking for a remote named "origin". Now the test is this: is there at least one remote with a URL that contains "github"? It is possible there should also be a requirement re: branch name and/or existence of a remote tracking branch.

  * Existing unexported function `github_info()` becomes more flexible about the name of the github remote. "origin" is still the top choice, with "github" and "upstream" being accepted alternatives, in that order.

  * Adds unexported function to delete a github repo. This was simply handy for development and informal testing.

  * Extends `use_github()` to support https protocol, in addition to ssh.

  * Extends `use_github()` to push the repo to GitHub and set a remote tracking branch. CAVEAT: I have not personally tested this with ssh -- only with https.

  * Adds exported function `use_github_links()` that populates URL and BugReports fields of DESCRIPTION with the correct GitHub links. It is called from `use_github()`.


R/git.R
```diff
@@ -63,18 +63,52 @@ git_path <- function(git_binary_name = NULL) {
...
-    return(github_dummy)
+    return(FALSE)
 
   r <- git2r::repository(path, discover = TRUE)
-  if (!("origin" %in% git2r::remotes(r)))
+  r_remote_urls <- remote_urls(r)
+
+  any(grepl("github", r_remote_urls))
+
+  ## or require ALSO that name of matching remote be in github_remote_names?
```
hadley: I don't think that's necessary

R/git.R
```diff
@@ -63,18 +63,52 @@ git_path <- function(git_binary_name = NULL) {
...
+github_info <- function(path = ".", remote_name = NULL) {
+  if (!uses_git(path))
+    return(github_dummy)
+
+  if (!uses_github(path))
     return(github_dummy)
 
-  github_remote_parse(git2r::remote_url(r, "origin"))
+  r <- git2r::repository(path, discover = TRUE)
+  r_remote_urls <- grep("github", remote_urls(r), value = TRUE)
```
hadley: I think the worry here is that if you have both origin and something else that points to github (e.g. upstream), this will pick the wrong one.

R/git.R
```diff
@@ -63,18 +63,52 @@ git_path <- function(git_binary_name = NULL) {
...
+
+  any(grepl("github", r_remote_urls))
+
+  ## or require ALSO that name of matching remote be in github_remote_names?
+  #any(grepl("github", r_remote_urls) &
+  #      names(r_remote_urls) %in% github_remote_names)
+}
+
+github_info <- function(path = ".", remote_name = NULL) {
+  if (!uses_git(path))
```
hadley: This can be eliminated since it's already covered by `uses_github()`

R/git.R
```diff
@@ -63,18 +63,52 @@ git_path <- function(git_binary_name = NULL) {
...
+    return(github_dummy)
+
+  if (!uses_github(path))
     return(github_dummy)
 
-  github_remote_parse(git2r::remote_url(r, "origin"))
+  r <- git2r::repository(path, discover = TRUE)
+  r_remote_urls <- grep("github", remote_urls(r), value = TRUE)
+
+  x <- r_remote_urls[remote_name]
```
hadley: I don't find the logic here easy to follow. I think you should be testing `r_remote_urls` rather than `r_remote_urls[NULL]` (as in the default case)

R/github.R
```diff
@@ -8,7 +8,9 @@ github_auth <- function(token) {
 
 github_response <- function(req) {
   text <- httr::content(req, as = "text")
-  parsed <- jsonlite::fromJSON(text, simplifyVector = FALSE)
+  parsed <-
+    if (identical(text, '')) text else
```
hadley: This doesn't seem like the right test to me. I think it `httr::has_content()` on `req` would be better. (I presume this was added to handle the delete repo case)

R/git.R
jennybc: The way I select the single remote URL to send to `github_remote_parse()` will always select the one named "origin" if such exists. Among remote URLs that contain "github", we prioritize remotes according to name like so: take "origin" ... or "github" ... or "upstream" ... or whatever's left.

Would you just rather keep the hard requirement that the primary github remote be named "origin"?

R/git.R
jennybc: ok I removed this comment and revert to plain old git2r::remote_url() since names will not be vetted

R/git.R
jennybc: good point
done

R/github.R
jennybc: done (and yes, reponse after successful deletion is empty)

jennybc:
I'll work on your remaining comments and add examples/tests when back from camping.

R/git.R
hadley: That's fine - but that strategy is not currently obvious in the code, so it needs a little refactoring.

jennybc:
Do you have a policy that you only test exported functions? Context: want to know if appropriate to write tests for `uses_git()`, `uses_github()`, `github_info()`, all of which are not exported. This is tempting because they can be tested without calling or mocking the GitHub API.

jennybc:
I will make a new PR shortly, where the above commits and more are squashed.

PR #901, opened 2015-08-22
Extend GitHub functionality

jennybc:
This PR might polish off #642.

  * Adds unexported function `uses_github()` to provide an explicit test for use in other functions.

  * Existing unexported function `github_info()` becomes more flexible about the name of the github remote. "origin" is still the top choice, but if it doesn't exist, other names are acceptable

  * Adds unexported function to delete a github repo. This was simply handy for development and informal testing.

  * Extends `use_github()` to support https protocol, in addition to ssh.

  * Extends `use_github()` to push the repo to GitHub and set a remote tracking branch. CAVEAT: I have not personally tested this with ssh -- only with https.

  * Adds exported function `use_github_links()` that populates URL and BugReports fields of DESCRIPTION with the correct GitHub links. It is called from `use_github()`.

  * Extends `dr_github()` to check for a GitHub connection and for dummy GitHub links in DESCRIPTION.

  * Adds a new test file, `test-github-connections.R`. You may not like it? It tests some unexported functions and many tests are disabled if we can't count on GITHUB_PAT being in place. It was helpful to me during development, but I can delete it from this PR if you wish.


jennybc:
BTW GitHub recommends https over ssh: [Which remote URL should I use?](https://help.github.com/articles/which-remote-url-should-i-use/). I made "ssh" the default in `use_github()` because that's what you had before but would you consider making "https" the default?

I would only need to invert order of default args for protocol in `use_github()`. Optionally, could also tweak `dr_github()`, which currently suggests that ssh > https.

jennybc:
`use_travis()` and `use_appveyor()` should probably call `uses_github()` early on to make a "go" / "no go" decision.

gaborcsardi:
@jennybc AFAIK both Travis and Appveyor can be used with GitHub alternatives, e.g. BitBucket, Kiln, etc.

jennybc:
Yes, you're right @gaborcsardi. I was being sort of myopic: I just looked for instances of `github_info()` that weren't preceded by a test for GitHub usage. As it is, the *messaging* in these functions presupposes GitHub usage.

Looking closer, `use_coverage()` is in the same boat.

So maybe the thing to do is to adjust the messaging if GitHub usage can or cannot be detected. Or just do nothing. Dummy values `<USERNAME>` and `<REPO>` aren't going to hurt anyone.

jimhester:
FWIW I would preserve the existing behavior (filling in user/repo) for the common case (GitHub), and just use the placeholders if `!isTRUE(uses_github())`.

Having to replace the dummy values really reduces the utility of these setup functions IMHO.

R/doctor.R
```diff
@@ -91,7 +93,21 @@ dr_github <- function(path = ".") {
...
   if (identical(Sys.getenv("GITHUB_PAT"), ""))
     msg[["PAT"]] <- paste("* GITHUB_PAT environment variable not set",
-      "(this is not critical unless you want to install private repos)")
+      "(this is not necessary unless you want to install private repos",
+      "or connect local repos to GitHub)")
+
+  pkg <- as.package(path)
+  desc_path <- file.path(pkg$path, "DESCRIPTION")
+  desc <- read_dcf(desc_path)
+  desc <- compact(desc[c("URL", "BugReports")])
```
hadley: I think it's more important to warn that Url and BugReports fields are empty. 

Do you think that people mindlessly copying-and-pasting `<USERNAME>/<REPO>` is a common failure mode?

R/git.R
```diff
@@ -63,19 +63,40 @@ git_path <- function(git_binary_name = NULL) {
...
+github_info <- function(path = ".", remote_name = NULL) {
+  if (!uses_github(path))
     return(github_dummy)
 
-  github_remote_parse(git2r::remote_url(r, "origin"))
+  r <- git2r::repository(path, discover = TRUE)
+  r_remote_urls <- grep("github", remote_urls(r), value = TRUE)
+
+  if (!is.null(remote_name) && !remote_name %in% names(r_remote_urls))
+    stop("no github-related remote named ", remote_name, " exists")
```
hadley: Maybe "found" is a little more accurate than "exists"?

R/github.R
```diff
@@ -53,6 +57,17 @@ github_tag <- function(username, repo, ref = "master") {
   github_GET(file.path("repos", username, repo, "tags", ref))
 }
 
+github_delete_repo <- function(username, repo, pat = github_pat()) {
```
hadley: Just having this function in devtools scares me a bit. Could you at least at an interactive confirmation prompt?

R/infrastructure-git.R
```diff
@@ -29,38 +29,90 @@ use_git <- function(message = "Initial commit", pkg = ".") {
...
 #' @param private If \code{TRUE}, creates a private repository.
+#' @param protocol transfer protocol, either "ssh" (the default) or "https"
 #' @family git infrastructure
 #' @keywords internal
 #' @export
-use_github <- function(auth_token = github_pat(), private = FALSE, pkg = ".") {
+#' @examples
+#' \dontrun{
+#' create("testpkg")
+#' use_github(pkg = "testpkg", protocol = "https", private = TRUE)
```
hadley: I think it's better to default to ssh in the R ecosystem.

R/infrastructure-git.R
```diff
@@ -29,38 +29,90 @@ use_git <- function(message = "Initial commit", pkg = ".") {
...
+  if (uses_github(pkg$path))
     return(invisible())
 
-  message("Creating GitHub repository")
-
+  message("* Creating GitHub repository")
   create <- github_POST("user/repos", pat = auth_token, body = list(
     name = jsonlite::unbox(pkg$package),
-    description = jsonlite::unbox(pkg$description),
+    #description = jsonlite::unbox(gsub("\n", " ", pkg$description)),
```
hadley: That seems like a reasonable change to me, so you can delete the comment

R/infrastructure-git.R
```diff
@@ -29,38 +29,90 @@ use_git <- function(message = "Initial commit", pkg = ".") {
...
+  if (git_uncommitted(pkg$path)) {
+    git2r::add(r, "DESCRIPTION")
+    git2r::commit(r, "Add GitHub links to DESCRIPTION")
+  }
+
+  message("* Pushing to GitHub and setting remote tracking branch")
+  ## in https case, when GITHUB_PAT is passed as password,
+  ## the username is immaterial, but git2r doesn't know that
+  cred <- git2r::cred_env("DUMMY_EMAIL", "PAT")
+  with_envvar(c("DUMMY_EMAIL" = "whatever", "PAT" = auth_token),
```
hadley: I think `EMAIL = ""` would be a bit nicer (it mimics http auth elsewhere)

R/infrastructure-git.R
```diff
@@ -29,38 +29,90 @@ use_git <- function(message = "Initial commit", pkg = ".") {
...
+              git2r::push(r, "origin", "refs/heads/master", credentials = cred))
+  ## now change remote URL if user requested ssh
+  if(protocol == "ssh") {
+    git2r::remote_remove(r, "origin") # git2r doesn't expose URL modification :(
+    git2r::remote_add(r, "origin", create$ssh_url)
+  }
+
+  git2r::branch_set_upstream(git2r::head(r), "origin/master")
```
hadley: Does this need a git2r version bump in DESCRIPTION?

R/infrastructure-git.R
```diff
@@ -98,3 +150,36 @@ add_git_ignore <- function(pkg = ".", ignores) {
...
+#' does not already use GitHub, the above links will be used quite literally.
+#'
+#' @inheritParams use_git
+#' @family git infrastructure
+#' @keywords internal
+#' @export
+use_github_links <- function(pkg = ".") {
+
+  if (!uses_github(pkg)) {
+    message(paste("Cannot detect that package already uses GitHub.",
```
hadley: Seems like this should just `stop()`. (Also `message()` and `stop()` take `...` so you don't need the `paste()`)

R/infrastructure-git.R
```diff
@@ -98,3 +150,36 @@ add_git_ignore <- function(pkg = ".", ignores) {
...
+                  "Adding the links to DESCRIPTION anyway."))
+  }
+
+  gh_info <- github_info(pkg)
+  pkg <- as.package(pkg)
+
+  desc_path <- file.path(pkg$path, "DESCRIPTION")
+  desc <- read_dcf(desc_path)
+
+  desc[["URL"]] <-
```
hadley: Would be nicer if this didn't clobber existing URL

R/infrastructure-git.R
```diff
@@ -98,3 +150,36 @@ add_git_ignore <- function(pkg = ".", ignores) {
...
+#' Populates the URL and BugReports fields of DESCRIPTION with
+#' \code{https://github.com/<USERNAME>/<REPO>} AND
+#' \code{https://github.com/<USERNAME>/<REPO>/issues}, respectively. If package
+#' does not already use GitHub, the above links will be used quite literally.
+#'
+#' @inheritParams use_git
+#' @family git infrastructure
+#' @keywords internal
+#' @export
+use_github_links <- function(pkg = ".") {
```
hadley: The uses functions are normally designed to fail instead of overriding existing fields

hadley:
Also, I agree with @jimhester WRT how the placeholders should work

R/doctor.R
jennybc: I check for dummy info because `use_github_links()` will dutifully insert placeholder links, even if GitHub remote can't be detected. It felt like good citizenship to build in a check for this weird state. No I cannot imagine anyone doing this on purpose!

Good point re: also checking for empty Url and BugReports. Will add.

jennybc:
Thanks for the comments! Will act on them.

R/doctor.R
hadley: Ah ok - I think it would be better to just barf in `use_github_links()`

R/infrastructure-git.R
jennybc: Yes, I imagine so. Currently DESCRIPTION say `git2r (>= 0.10.1)` but I am working against `0.11.0` from CRAN and suspect this actually matters.

In fact, you might want `git2r` to have a CRAN release before you merge this. There are at least 2 changes in v0.11.0.9000 that were prompted by issues I opened and that, if exploited, would make this code cleaner. Notes mostly for me:

  * Ability to edit a remote's URL (https://github.com/ropensci/git2r/commit/26276693ab80d7cce4171786f04c467c66f7d079) vs. delete it then add it back
  * New credential method that works with PAT only (https://github.com/ropensci/git2r/commit/a91a716fc6d7bcc7177e7bdbd942ff646f960f72) vs dummy email/username and PAT as password

R/github.R
jennybc: I'm moving it into the test file, where it is used and makes it reasonably available for development.

R/infrastructure-git.R
jennybc: I just learned, with some pain, that if you send an empty email, `git2r::push()` hangs the R process (and RStudio somehow can't recover). I'll bring that up over in `git2r`. But for now I'm going with `EMAIL = "EMAIL"` here. Once `git2r` updates, we can use the new `git2r::cred_token()` and will no longer need to provide dummy email.

jimhester: Wouldn't it be more clear to just lookup the user's configured git email and use that here?
```r
c <- git2r::config(r)
email <- c$local$user.email %||% c$global$user.email
```
Would do the job I think.

jennybc:
Thank you for your patience. I did as asked in all the comments. Highlights of changes and departures from initial PR message:

  * The scary `github_delete_repo()` function had been moved into `test-github-connections.R`. Has an improved check for repo existence and a commented-out interactive confirmation menu.

  * `devtools` preference for ssh protocol over https faithfully upheld.

  * New exported function `use_github_links()` does nothing if no GitHub connection detected and will not clobber existing entries in URL and BugReports fields of DESCRIPTION.

  * `dr_github()` diagnoses the lack of a GitHub connection, empty URL or BugReports, and lack of GitHub repo and issue links.

  * Functions `use_travis()`, `use_appveyor()`, and `use_coverage()` were not touched.

  * `git2r` version bumped to v0.11.0 and comments left where things can be made nicer when > v0.11.0 is on CRAN.

jennybc:
This last travis failure appears to just be bad luck, but I can't restart it. Successful build and test with previous commit.

DESCRIPTION
```diff
@@ -29,10 +29,10 @@ Imports:
     rversions,
     stats,
     git2r (>= 0.10.1)
-Suggests: 
+Suggests:
     testthat (>= 0.7),
     BiocInstaller,
-    Rcpp (>= 0.10.0),
+    Rcpp (>= 0.11.0),
```
hadley: Was there a reason for this change?

R/github.R
```diff
@@ -7,8 +7,12 @@ github_auth <- function(token) {
 }
 
 github_response <- function(req) {
-  text <- httr::content(req, as = "text")
-  parsed <- jsonlite::fromJSON(text, simplifyVector = FALSE)
+  if (length(httr::content(req)) > 0) {
+    text <- httr::content(req, as = "text")
+    parsed <- jsonlite::fromJSON(text, simplifyVector = FALSE)
+  } else {
+    parsed <- ''
```
hadley: I think that should probably be `parsed <- list()`

tests/testthat/test-github-connections.R
```diff
@@ -0,0 +1,187 @@
+context("GitHub connections")
+
+## set auth to TRUE if env var GITHUB_PAT is available (eg local testing)
+## set auth to FALSE otherwise (eg on travis, at this point)
+auth <- FALSE
```
hadley: How does `auth` get set?

tests/testthat/test-github-connections.R
```diff
@@ -0,0 +1,187 @@
...
+      path <- file.path(path, pkg_name)
+      create(path, description = list(NULL))
+    })
+  )
+  path
+}
+
+erase_test_pkg <- function(pkg_path) unlink(pkg_path, recursive = TRUE)
+
+github_delete_repo <- function(pkg_path) {
```
hadley: This still feels risky to me - I'd rather not test adding repos if we have to delete them afterwards.

DESCRIPTION
jennybc: I cannot *promise* that v0.11.0 is necessary, though suspect it is. I will check with v0.10.0.

hadley: Did you mean to change the version of git2r?

R/github.R
jennybc: Now that I've moved `github_delete_repo()` out of the package, I could actually put this function back as I found it. This was only necessary because the content of the response after successful deletion is empty and `jsonlite:;fromJSON("")` errors. But apparently that has never been a problem before.

hadley: Sounds good.

DESCRIPTION
jennybc: YES `git2r` must be >= 0.11.0. Otherwise `cred_env()` is not exported and we are unable to establish push to Github. And if you can't push, you can't set the remote tracking branch.

hadley: Do you realise that you accidentally modified the version of Rcpp instead of git2r?

tests/testthat/test-github-connections.R
jennybc: I set it to `TRUE` when I check locally, because I'm happy for my `GITHUB_PAT` to be used for tests of GitHub connectivity.

In theory, I change it back to `FALSE` every time I push, because `devtools` doesn't have an encrypted `GITHUB_PAT` env var on Travis yet.

In reality, I forget and have to make even more commits :(

tests/testthat/test-github-connections.R
jennybc: Should I write a version of that file that has zero ambitions to actually do anything on GitHub?

DESCRIPTION
jennybc: :flushed:

jennybc:
Please see the revised tests. No more calls to the GitHub API and no `git2r` usage that actually hits a remote.

hadley:
Oops, forgot to ask for a bullet point in NEWS. Could you please add in a new PR?

jennybc:
Yes re: NEWS but will be much later today. First day classes.

PR #910, opened 2015-09-02
Fix with_lib tests on Windows

krlmlr:
Something was weird with the normalization of the paths, perhaps some 8.3 problem. Now the tests also normalize the return value of .libPaths(), this seems to fix the issue.

Works according to WinBuilder:
http://win-builder.r-project.org/4NAgZD9sPrem.

~~~This should now work also on Windows, if not please ping me again.~~~


hadley:
Thanks!

PR #914, opened 2015-09-04
bugfix when calling install_dev_remotes

jimhester:
The last refactoring of `install_dev_remotes` had a bug which made it non-functional.

This should fix things so it actually works, sorry I missed this earlier!


hadley:
np

PR #921, opened 2015-09-10
convert errors to warnings when processing NAMESPACE

krlmlr:
in `process_imports`

Reason: Otherwise, `devtools::document` cannot fix a `NAMESPACE` file that became invalid due to changes in external packages

(I'm not sure if this worked before.)


hadley:
Thanks!

PR #922, opened 2015-09-10
Allow for ssh keys with passphrase

jennybc:
Solves difficulties setting remote tracking branch when protocol is ssh and the ssh keys have a passphrase. See https://github.com/hadley/devtools/issues/642.


hadley:
Thanks!

Issue #800, opened 2015-05-14
accept alternative maintainer email for build_win


maxheld83:
I'm guessing this is a common scenario:

1. I'm working on a fork of some maintainer's R project
2. I'm ready to open a PR and want to make sure that it passes `R CMD Check` on all platforms ...
3. ... so I use the great `build_win`
4. ... and realize a couple of seconds too late that the resulting email will be sent to the *maintainer* of the project.

Would be nice to specify an alternative email address to `build_win` as an argument.

wch:
Win-builder simply uses the maintainer's email address, so this would require modification to the DESCRIPTION file before uploading.

I do agree that in general that something like this would be useful - perhaps a confirmation prompt that says  "message will be sent to person@foo.com", before uploading the package?

krlmlr:
We could use `as.person(eval(parse(text=getOption("devtools.desc.author"))))$email` as e-mail address, and perhaps a prompt (in interactive mode) or an argument that allows overriding the maintainer's mail address (for scripting). Not sure if this will make everybody happy, though.

hadley:
I think it would be fine as an argument to `build_win()`. But modifying the DESCRIPTION file is a pain.

hadley:
I'd be happy to accept a PR that implemented this.

PR #924, opened 2015-09-11
Explicitly request consent for sending e-mail to maintainer

krlmlr:
`build_win()`, only in interactive mode. A cheap solution to #800.


hadley:
Thanks!

PR #925, opened 2015-09-13
Remove with_ functions from devtools and use those from withr

jimhester:
I am planning on submitting withr to CRAN in the next week or so.  But I wanted to send a PR to iron out any bugs (and try the new `Remotes:` feature)


jimhester:
This now works, all of the calls to `with_` functions _within_ devtools are replaced by calls to the corresponding withr functions.

However we need to decide how to handle calls to the old `with_` function from _other_ packages. There are three possibilities I see

1. Do nothing (this will break packages that use the functions, but will force them to update to withr functions immediately).
2. Continue to supply the functions and with a `.Deprecated` warning, remove them at a later date.
3. Export the existing functions as aliases to the corresponding withr functions.

1 isn't going to happen so 2 seems to be the best option, as part of the reason for splitting the functions out was so fewer packages would need to depend on devtools directly.

hadley:
I think 1 would be ok if you did the revdep checks and informed the revdep maintainers

jimhester:
Oh well... already did the deprecation work!

https://github.com/jimhester/devtools/commit/9fd20d284225a610e7a3397a02932e06471a13cb should be good (assuming it passes).

hadley:
Thanks!

PR #927, opened 2015-09-14
use_coverage() suggest installing dependencies of covr as binary packages

krlmlr:
to save roughly 5 minutes in each CI run.

See also kevinushey/rex#43.


hadley:
This is just an interim problem with travis. I don't think it's a good idea to add these binary deps.

krlmlr:
Just to clarify: The way it is now, all deps of the `covr` package will be installed from source. Using

```
r_binary_packages:
  - covr
```

doesn't work, because `rex` is not yet available on Debian. Once `covr` is on Debian, we might want to use the declaration shown above.

PR #929, opened 2015-09-16
Degrade gracefully if summary cannot be retrieved for one of the checked packages

krlmlr:
- example: if the DESCRIPTION file in one of the checked packages is missing


R/revdep-summarise.R
```diff
@@ -63,6 +65,14 @@ revdep_check_summary <- function(res) {
   )
 }
 
+try_check_summary_package <- function(path) {
+  res <- try(check_summary_package(path), silent = TRUE)
```
hadley: Can you please use `tryCatch()` instead? I don't like the way `try()` works

hadley:
Thanks!

PR #932, opened 2015-09-17
Optionally recode non-ASCII DESCRIPTION in read_dcf

krlmlr:
- recode DESCRIPTION files for revdep check, to ensure their contents can be written when creating the summary
- also, use all = FALSE for read.dcf() to always return a matrix, even for malformed DESCRIPTION files

### DIFF
--- a/R/revdep-summarise.R
+++ b/R/revdep-summarise.R
@@ -109,7 +109,7 @@ check_dirs <- function(path) {
 
 check_description <- function(path) {
   pkgname <- gsub("\\.Rcheck$", "", basename(path))
-  read_dcf(file.path(path, "00_pkg_src", pkgname, "DESCRIPTION"))
+  read_dcf(file.path(path, "00_pkg_src", pkgname, "DESCRIPTION"), recode = TRUE)
 }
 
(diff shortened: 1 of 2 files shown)


hadley:
I haven't had any problems with this so would you mind adding a test case?

krlmlr:
I remember that this was useful when revdep-checking memoise -- the surveillance package [uses Latin1 encoding](https://github.com/cran/surveillance/blob/c93f52c3d54e4e42a7a9db89251fa9e52e8e6dea/DESCRIPTION#L84), and this messes up the summary without this PR. Does that count as test case? :-)

hadley:
I think it's probably better to push this into https://github.com/metacran/description and then eventually using description in devtools.

krlmlr:
@gaborcsardi Does "description" have equivalent code for the case of non-ASCII encoding?

gaborcsardi:
@krlmlr I don't think so. 

But I would not recode the fields when reading the file. Then you would need to recode it back when exporting, copying, etc.

I would rather change it when printing, or when the problem actually comes up.

PR #935, opened 2015-09-22
Fix error message in release()

jimhester:
It was missing a newline

```
Error: Can't find cran-comments.md in withr.
This file gives CRAN volunteers comments about the submission,
and it must exist.  Please create it using this guide:
http://r-pkgs.had.co.nz/release.html#release-checkThen run use_build_ignore('cran-comments.md')
```


hadley:
Thanks!  (It also really needs to come much early in the check)

PR #936, opened 2015-09-23
Use get directly instead of match.fun

jimhester:
The previous implementation used `match.fun()`, which has the undesirable effect that the function cannot be found if the devtools namespace is not attached.

So `devtools::install_deps()` would fail, which is what was happening on travis (https://travis-ci.org/hadley/devtools/builds/81765159#L2108-L2110) when I tried to use `Remotes:`

This fix uses `get()` directly in the `devtools` namespace, so works regardless if devtools is attached.

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -152,7 +152,9 @@ dev_remote_type <- function(remotes = "") {
     } else {
       stop("Malformed remote specification '", x, "'", call. = FALSE)
     }
-    tryCatch(fun <- get(paste0("install_", tolower(type)), envir=asNamespace("devtools")),
+    tryCatch(fun <- get(x = paste0("install_", tolower(type)),
+                        envir = asNamespace("devtools"),
+                        mode = "function"),
       error = function(e) {
         stop("Malformed remote specification '", x, "'", call. = FALSE)
       })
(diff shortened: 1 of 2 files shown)


R/deps.R
```diff
@@ -152,7 +152,9 @@ dev_remote_type <- function(remotes = "") {
     } else {
       stop("Malformed remote specification '", x, "'", call. = FALSE)
     }
-    tryCatch(fun <- match.fun(paste0("install_", tolower(type))),
+    tryCatch(fun <- get(x = paste0("install_", tolower(type)),
```
hadley: I'm not terribly consistent about enforcing this, but I'd prefer subsequent arguments to be indented just one extra level

jimhester: Should be fixed by 4903b51

wch:
Is it desirable to have it search in non-devtools environments? (That's what happens because of the default `inherits=TRUE`).

jimhester:
@wch It is probably not desirable.  Doing so opens the possibility for another package to export an `install_*` function to support a new backend, but I am not sure that is worth the extra search cost and safety concerns.

wch:
I think you mean that commit uses `inherits = FALSE`. :)

jimhester:
I edited the comment, no one will ever know!

jimhester:
https://github.com/jimhester/devtools/commit/d21c892ae490385e6c8891a6df77ec3603ddc481 should be ready to merge whenever. This will allow use of the `Remotes: ` feature on travis, so it is fairly useful.

hadley:
Thanks!

PR #937, opened 2015-09-24
Use the proper Bioconductor URL for both release and devel

jimhester:
Depending on if the user has ran BiocInstaller::useDevel() the
repository URL differs. The previous behavior was hard-coded to look only
at Bioconductor package in the last release. Making URL a parameter
ensures that if the user changes from looking at devel to release in the
same session the previous cached result will not be used.


jimhester:
This should now be ready to merge.

hadley:
Thanks!

PR #945, opened 2015-10-04
Unclutter AppVeyor output

krlmlr:
- in case of failure, save check directory as artifact

Per @HenrikBengtsson's suggestion in krlmlr/r-appveyor#37.


hadley:
Thanks!

PR #951, opened 2015-10-16
Typo correction in infrastructure.R

sebastian-c:
already exist -> already exists


sebastian-c:
This needs a better fix than what I did. It's mostly likely that there will be multiple files.

PR #956, opened 2015-10-26
Move curl dependency to Suggests

jimhester:
It is only used in 1 function, (upload_ftp()) which is only used in
build_win() so not needed most of the time.


jimhester:
I am just going to add this to https://github.com/hadley/devtools/pull/962 rather than using a separate pull request for it.

PR #964, opened 2015-10-28
update 'package_deps' documentation

kevinushey:
Fills in some of the holes in documentation on the `package_deps()` function.


hadley:
Thanks!

PR #967, opened 2015-11-03
Fix devtools::load_all() in noninteractive mode if no DESCRIPTION exists

krlmlr:
Currently fails in this case.


hadley:
Thanks!

PR #941, opened 2015-09-30
don't insert unnecessary trailing whitespace (closes #940)

kevinushey:
This PR fixes an issue where an unnecessary single space would be inserted after fields that already began with whitespace.

PR #968, opened 2015-11-03
Strip space at EOL in write_dcf()

krlmlr:
to avoid bogus changes.


hadley:
This is a different problem to #941, right?

krlmlr:
I think it's the same problem, the solution in #941 is more subtle. Let's shelve this for now.

PR #976, opened 2015-11-10
use_git(): if git is already initialized, tell the user that

daattali:
if you already have git in your project but don't realize it and run `use_git()`, it can be confusing to not see any output and think "why isn't it doing anything?". This isn't just a hypothetical, it happened today and it took us a few minutes to debug in @jennybc class

### DIFF
--- a/R/infrastructure-git.R
+++ b/R/infrastructure-git.R
@@ -10,8 +10,10 @@
 use_git <- function(message = "Initial commit", pkg = ".") {
   pkg <- as.package(pkg)
 
-  if (uses_git(pkg$path))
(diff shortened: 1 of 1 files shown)


jennybc:
@daattali Maybe you should do similar for `use_github()`? It's also silent when it does nothing.

Right here:
https://github.com/hadley/devtools/blob/81dd313e1d143a3bdaaa2e1f09c887801db66694/R/infrastructure-git.R#L84-L85

daattali:
@jennybc done https://github.com/hadley/devtools/pull/977

hadley:
Thanks!

PR #980, opened 2015-11-16
Modify pre-commit hook re: README.Rmd; fixes #979

### LINKED ISSUE #979: make README pre-commit hook more selective
The hook in [readme-rmd-pre-commit.sh](https://github.com/hadley/devtools/blob/master/inst/templates/readme-rmd-pre-commit.sh) prevents a commit if `README.md` is older than `README.Rmd`, even when those files aren't staged. Might be nice to make it a bit less trigger happy? [This superuser post](http://superuser.com/questions/552617/execute-git-pre-commit-hook-only-if-files-in-certain-directory-modified) looks easy to adapt.

jennybc:
* do nothing if neither README.Rmd nor README.md staged
  * keep checking that .md is newer than .Rmd
  * check that both are staged
  * tell user how to override


hadley:
Thanks!

PR #981, opened 2015-11-16
release() and build_win() use --compact-vignettes=both

krlmlr:
to avoid WARNINGs with the built package.


hadley:
I've never seen these notes. Can you provide more details?  What system requirements are necessary to be able to compact vignettes?

krlmlr:
It's even a warning, and happens with PDF vignettes.

http://win-builder.r-project.org/48l69H500MSZ/00check.log was created with 3835a87f49 and contains:

```
* checking sizes of PDF files under 'inst/doc' ... WARNING
  'gs+qpdf' made some significant size reductions:
     compacted 'tikzDevice.pdf' from 1137Kb to 493Kb
```

hadley:
Hmmm, if it needs other software, we definitely shouldn't do this unconditionally.  Could you please add a check for the presence of the needed software?

krlmlr:
Didn't think about that. Adding an `args` argument to `release()` would be simpler and potentially more useful, would you prefer that? (`build_win()` already has one.)

hadley:
Yeah, that sounds good.

PR #982, opened 2015-11-18
Support passthrough of credentials to git2r

onlymee:
git2r has very sensible defaults for credentials, eg. for ssh it defaults to looking for ssh-agent, but there are times where it is useful to specify specific credentials.

This modification simply allows for a git2r credentials object to passet through to the underlying clone call.  This is useful if you wish to use devtools non-interactively (e.g. a script in a CI solution)


R/install-git.r
```diff
@@ -18,20 +20,21 @@
 #' install_git("git://github.com/hadley/stringr.git")
 #' install_git("git://github.com/hadley/stringr.git", branch = "stringr-0.2")
 #'}
-install_git <- function(url, subdir = NULL, branch = NULL, args = character(0),
+install_git <- function(url, subdir = NULL, branch = NULL, credentials=NULL, args = character(0),
```
hadley: Please put spaces around `=` to match existing style

hadley: Bump

jennybc: @hadley would you like same thing done for `use_github()`?

hadley: Yes, if it would be helpful

hadley:
If you can do this by Friday it'll make it in 1.10, otherwise it will have to wait for 1.11 (probably in about a months time)

onlymee:
Also added the credentials passthrough to use_github as suggested.  I checked for other git2r clone/push/fetch but missed that one.

jimhester:
@onlymee you can squash your commits with `git rebase -i` to combine them. See https://help.github.com/articles/about-git-rebase/.

You should also rebase this PR against the current master first which would remove a lot of the above commits. If you have `hadley/devtools' as the remote called `upstream` then `git rebase upstream master` when you are on your master branch should do it.

hadley:
Can you please rebase & squash?

jimhester:
I rebased, squashed and merged at https://github.com/hadley/devtools/commit/91fddbf5d3376faf76527db51e4d212913b1ef21. Thanks @onlymee!

PR #997, opened 2015-12-16
More permissive syntax for DCF

cmacrander:
Allows blank lines and comments in DESCRIPTION files. Both whitespace and comments are essential for readability and maintainability. Although not strictly DCF format, it is the modern norm.


hadley:
This is for devtools internal use only, and I'm pretty sure that R's `read.dcf()` doesn't support whitespace or comments.

Issue #942, opened 2015-09-30
devtools::install_github("hadley/devtools") failure


csrvermaak:
Hi Hadley

I'm reading your book “R packages”, but unfortunately got stuck at the Getting Started section with the 
devtools::install_github("hadley/devtools") line. 

I've installed the packages:
install.packages(c("devtools", "roxygen2", "testthat", "knitr"))

I've got the RStudio API:
> rstudioapi::isAvailable("0.99.149")
[1] TRUE

But, when I run the following:
devtools::install_github("hadley/devtools")

I get this:

> devtools::install_github("hadley/devtools")
Downloading github repo hadley/devtools@master
Installing devtools
"C:/PROGRA~1/RRO/R-32~1.2/bin/x64/R" --no-site-file --no-environ --no-save --no-restore CMD INSTALL  \
  "C:/Users/Rvermaak/AppData/Local/Temp/Rtmpwnsawu/devtools2e2455761c23/hadley-devtools-81dd313"  \
  --library="C:/Users/Rvermaak/Documents/R/win-library/3.2" --install-tests 

* installing *source* package 'devtools' ...
** libs
gcc -m64 -I"C:/PROGRA~1/RRO/R-32~1.2/include" -DNDEBUG     -I"c:/applications/extsoft/include"     -O2 -Wall  -std=gnu99 -mtune=core2 -c devtools.c -o devtools.o
gcc -m64 -shared -s -static-libgcc -o devtools.dll tmp.def devtools.o -Lc:/applications/extsoft/lib/x64 -Lc:/applications/extsoft/lib -LC:/PROGRA~1/RRO/R-32~1.2/bin/x64 -lR
installing to C:/Users/Rvermaak/Documents/R/win-library/3.2/devtools/libs/x64
Warning in file.copy(files, dest, overwrite = TRUE) :
  problem copying .\devtools.dll to C:\Users\Rvermaak\Documents\R\win-library\3.2\devtools\libs\x64\devtools.dll: Permission denied
** R
** inst
** tests
** preparing package for lazy loading
** help
*** installing help indices
** building package indices
** installing vignettes
** testing if installed package can be loaded
* DONE (devtools)

(NOTE: There is a file copy warning, but it seems to have installed fine?)

When I try to subsequently load the library, things go a bit pear-shaped:

> library(devtools)
Error in get(method, envir = home) : 
[...]

Issue #631, opened 2014-11-02
install_github doesnot work but R CMD INSTALL works


zhilongjia:
While I install a package (https://github.com/zhilongjia/infotheo , include cpp code) in local, it works. When I install it via install_package() via nectwork, it does not work. So there seems to be some different operation between R CMD INSTALL ang install_github(). Maybe it's related the cpp code.

similar error in https://github.com/hadley/devtools/issues/503, but I cannot catch answer from there.

To install the package https://github.com/zhilongjia/infotheo
the verbose infor of install_github() is 
    *\* libs
    make: Nothing to be done for `all'.

while when I use R CMD INSTALL in local. these are
    *\* libs
    g++ -I/usr/share/R/include -DNDEBUG      -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -g  -c discretize.cpp -o discretize.o
    g++ -I/usr/share/R/include -DNDEBUG      -fpic  -g -O2 -fstack-protector-strong -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2 -g  -c entropy.cpp -o entropy.o
    g++ -shared -Wl,-z,relro -o infotheo.so discretize.o entropy.o -L/usr/lib/R/lib -lR

Other lines are the same.
About the help. it shows "Error in fetch(key) : lazy-load database '�' is corrupt"

Package devtools version 1.6.1, in Debian 8.  
Thank you.

Zhilong

Issue #503, opened 2014-06-24
install_github() with custom URL


ghuiber:
I'm filing an issue because I want Markdown, but this is actually a request:

I would like to write packages for coworkers that they could adopt with `install_github()`. For that, this function must be able to query a custom enterprise repo whose URL is of the form `https://github.scm.corp[...].com`.

So I forked `devtools` and tried to simply add an extra parameter to `install_github()`, called `repo_path` and with a default value set to `="https://github.com"`. This was to be used by `install_github_single()` via its own call to `github_get_conn()`. 

I updated the `@param` set in the header of `install_github.r` accordingly, then ran

```
> load_all(); document(); test(); install();
```

It should have worked, I thought, but `install()` protested as shown below:

```
> load_all()
Loading devtools
> getwd()
[1] "C:/Users/ghuiber/Documents/GitHub/devtools"
> document(); test(); install()
Updating devtools documentation
Loading devtools
Writing NAMESPACE
Writing r_env_vars.Rd
Writing bash.Rd
Writing build_github_devtools.Rd
Writing add_build_ignore.Rd
Writing build.Rd
Writing build_win.Rd
Writing check_cran.Rd
Writing check_doc.Rd
Writing check.Rd
Writing cran_env_vars.Rd
Writing clean_source.Rd
Writing eval_clean.Rd
Writing compile_dll.Rd
Writing clean_dll.Rd
Writing create.Rd
Writing dev_example.Rd
Writing dev_help.Rd
Writing help.Rd
Writing dev_meta.Rd
Writing dev_mode.Rd
Writing document.Rd
Writing has_devel.Rd
Writing has_tests.Rd
Writing imports_env.Rd
Writing load_imports.Rd
Writing infrastructure.Rd
Writing inst.Rd
Writing install_bitbucket.Rd
Writing install_git.Rd
Writing install_git_single.Rd
Writing git_path.Rd
Writing install_github.Rd
Writing github_pat.Rd
Writing install_gitorious.Rd
Writing install_local.Rd
Writing install_url.Rd
Writing install_version.Rd
Writing install.Rd
Writing install_deps.Rd
Writing load_code.Rd
Writing load_data.Rd
Writing load_dll.Rd
Writing load_all.Rd
Writing create_description.Rd
Writing missing_s3.Rd
Writing ns_env.Rd
Writing parse_ns_file.Rd
Writing parse_deps.Rd
Writing check_dep_version.Rd
Writing pkg_env.Rd
[...]

PR #1001, opened 2015-12-23
Unregister S3 methods explicitly

### LINKED ISSUE #419: Error in fetch(key) : lazy-load database '�' is corrupt
The documentation is not reloaded properly when changes are made.

For example:

```
library(devtools)
dev_mode()
document("mypackage")
install("mypackage")
?some_function # shows the documentation correctly

# I keep working and make modifications to the documentation of "some_function".
# I want to update the documentation and the package so I issue:

detach("package:mypackage", unload = TRUE)
document("mypackage")
install("mypackage")
?some_function
```

Now the documentation shows 

```
"Error in fetch(key) : lazy-load database '�' is corrupt"
```

However if I restart my R session the documentation is visualized correctly.

jimhester:
This is a workaround for a bug in base R. unloadNamespace does not
unregister any S3 methods, so the promises are invalid if you install a
new version of a package, try to unload and re-load the package.

See https://stat.ethz.ch/pipermail/r-devel/2015-December/072150.html for
more details and a reproducible example.

This fixes the common lazy-load errors people often run into when trying
to use devtools' install functions.

Fixes #419, #503, #942, #631

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # devtools 1.9.1.9000
 
+* Fix longstanding lazy load database corruption issues when reloading packages
+  which define S3 methods on generics from base or other packages (#1001, @jimhester).
+
 * `document()` now only runs `update_collate()` once.
 
 * Bugfix for `Remotes: ` feature that prevented it from working if devtools was
--- a/R/install.r
+++ b/R/install.r
@@ -63,6 +63,10 @@ install <- function(pkg = ".", reload = TRUE, quick = FALSE, local = TRUE,
   pkg <- as.package(pkg)
   check_build_tools(pkg)
 
+  if (is_loaded(pkg)) {
+    eapply(ns_env(pkg), force, all.names = TRUE)
+  }
+
   if (!quiet) {
     message("Installing ", pkg$package)
   }
--- a/R/install.r
+++ b/R/install.r
@@ -63,6 +63,9 @@ install <- function(pkg = ".", reload = TRUE, quick = FALSE, local = TRUE,
   pkg <- as.package(pkg)
   check_build_tools(pkg)
 
+  # Forcing all of the promises for the current namespace now will avoid lazy-load
+  # errors when the new package is installed overtop the old one.
+  # https://stat.ethz.ch/pipermail/r-devel/2015-December/072150.html
   if (is_loaded(pkg)) {
     eapply(ns_env(pkg), force, all.names = TRUE)
   }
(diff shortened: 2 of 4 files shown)


jimhester:
While this is working for the test package and for devtools itself I have not tested it with any S4 packages to ensure it works there. I wanted to put it out there to get feedback on the implementation and see if there are any changes that need to be made.

This error has long been a frustrating one for me, as it seemed to happen randomly, so finally figuring out what was going on and fixing it was pretty satisfying!

wch:
Nice! If this is indeed a reliable fix that's compatible with S4 stuff, it'll be great to have those errors go away.

jimhester:
https://github.com/hadley/devtools/commit/200b631d42192c966dfacff5a48b51edd837f70f uses a different approach to fixing the issue by simply forcing all of the promises explicitly before trying to unload rather than removing them. I also converted all of the previous uses of `as.list()` to use `eapply(ns, force)` to make it a little more explicit what was going on.

The previous commit was running into problems when trying to unload or reinstall devtools that do not occur using this method.

I also tested this version on some S4 classes from Bioconductor and they seem to work without error.

hadley:
I like it. I think the only thing that's missing is a comment with the code explaining your reasoning above

jimhester:
I actually had a comment written (https://github.com/hadley/devtools/commit/6d760c5bf903294bee4aaf3dfc6f01f2e2cac51e) just neglected to push it. I also added a note to the NEWS about this change https://github.com/hadley/devtools/commit/d96101f163916162a81a4f12f820f4849917f4e4.

Let me know if you want me to squash the commits or leave them as-is.

hadley:
Thanks!

jimhester:
For reference I opened a bug report about this (https://bugs.r-project.org/bugzilla3/show_bug.cgi?id=16644). BDR points out it is technically documented behavior, so there is unlikely to be a fix upstream. 

Personally I think the current behavior of `detach()` and `unloadNamespace()` is hostile to users. You should be able to first load and unload a package and have your environment be identical to when you started. It doesn't even seem that difficult to add this support, so I wonder why it wasn't implemented (I am probably missing some subtlety).

Anyway this pull should fix the majority of the errors people experience when using devtools at least, thank you for merging it!

hadley:
@jimhester thanks for digging into this - it's been a hassle for ages!

gaborcsardi:
:100:

gaborcsardi:
Btw. this makes me want to create an `uploader` package. :)

brodieG:
@jimhester bloody awesome.  Best X-mas present this year.  So sick and tired of the install-test-install-corrupt-db-quit-restart-crap-didn't-save-workspace-cycle.

sanjmeh:
I am still facing the same frustrating install-test-install-corrupt-db-quit-restart-crap-didn't-save-workspace cycle. Can one of the experts here document the solution or attach a link to Stack overflow where this is addressed to your satisfaction? Apologies if this is already resolved and documented somewhere else.

PR #1005, opened 2016-01-03
fix message pluralization in package install

leeper:
This fixes a small grammatical error in the package installation messages, such as the last line of the following:

```R
> devtools::install_github("leeper/rio")
Downloading GitHub repo leeper/rio@master
Installing rio
Installing 1 packages: BH
```

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -221,14 +221,18 @@ print.package_deps <- function(x, show_ok = FALSE, ...) {
 update.package_deps <- function(object, ..., quiet = FALSE, upgrade = TRUE) {
   ahead <- object$package[object$diff == 2L]
   if (length(ahead) > 0 && !quiet) {
-    message("Skipping ", length(ahead), " packages not available: ",
-      paste(ahead, collapse = ", "))
+    message(sprintf(ngettext(length(ahead), 
+      "Skipping %d unavailable package: %s", 
+      "Skipping %d unavailable packages: %s"
+    ), length(ahead), paste(ahead, collapse = ", ")))
   }
 
   missing <- object$package[object$diff == 1L]
   if (length(missing) > 0 && !quiet) {
-    message("Skipping ", length(missing), " packages ahead of CRAN: ",
-      paste(missing, collapse = ", "))
+    message(sprintf(ngettext(length(missing), 
+      "Skipping %d package ahead of CRAN: %s", 
+      "Skipping %d packages ahead of CRAN: %s"
+    ), length(missing), paste(missing, collapse = ", ")))
   }
 
   if (upgrade) {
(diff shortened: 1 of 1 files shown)


hadley:
How does this handle languages with different pluralisation rules? (http://unicode.org/repos/cldr-tmp/trunk/diff/supplemental/language_plural_rules.html)

leeper:
`ngettext()` is the general mechanism for internationalization of these types of messages. devtools doesn't have any message translations, so before and after this change all messages will be printed in English regardless of locale. If one were to contribute translations (.po files) with the correct pluralizations, they would be appropriately accessed after the change (but not under devtool's current behavior). So, this is an i18n step, l10n could come next.

hadley:
But how can that work with just two options?

leeper:
Take an example from [`print.data.frame()`](https://github.com/wch/r-source/blob/eb4605682aa287f106151d64d4d173378a3d1b91/src/library/base/R/dataframe.R#L1410-L1413): 

```R
    if(length(x) == 0L) {
	cat(sprintf(ngettext(n, "data frame with 0 columns and %d row",
			     "data frame with 0 columns and %d rows"),
		    n), "\n", sep = "")
    }
```

What happens here is that in English there are only two possible pluralizations (for `n==1` and for `n != 1`). They're provided as arguments to the function. The translations, [say in Polish](https://github.com/wch/r-source/blob/trunk/src/library/base/po/R-pl.po#L1736-L1740), are provided in [a .po file](https://github.com/wch/r-source/blob/trunk/src/library/base/po/R-pl.po) in the package source:

```
msgid "data frame with 0 columns and %d row"
msgid_plural "data frame with 0 columns and %d rows"
msgstr[0] "ramka danych z zerową liczbą kolumn oraz %d wierszem"
msgstr[1] "ramka danych z zerową liczbą kolumn oraz %d wierszami"
msgstr[2] "ramka danych z zerową liczbą kolumn oraz %d wierszami"
```

When R is running in English, the arguments to `ngettext()` are used directly. When running in a different locale, the gettext catalog (if present and installed) is queried for the appropriate translation based on the value of `n`. The format above shows pluralizations for a three-plural language. 

(That's not a great example because the plural forms for `n==1` and `n>=2` are the same, but there apparently aren't any examples of easily-accessible messages in base R where the messages are actually different in those two forms in any translated language ([there's an example of one here](https://github.com/wch/r-source/blob/trunk/src/library/base/po/R-pl.po#L1678-L1682), but that message can't be triggered to display all of its possible values).)

In another (non-English), two-plural language like [French](https://github.com/wch/r-source/blob/trunk/src/library/base/po/R-fr.po#L1721-L1724), the file is similar but with only two translations:

```
msgid "data frame with 0 columns and %d row"
msgid_plural "data frame with 0 columns and %d rows"
msgstr[0] "tableau de données (data frame) avec 0 colonnes et %d lignes"
msgstr[1] "tableau de données (data frame) avec 0 colonnes et %d lignes"
```

So, in short, the value of the 2nd argument to `ngettext()` is used to look up translations in the catalog and the appropriate, pluralized translation is selected from among those available based on the value of `n`.

hadley:
Got it - thanks!

Wouldn't it be better style to do (e.g.)

```R
message(ngettext(length(ahead), 
  "Skipping %d package not available: ", 
  "Skipping %d packages not available: "
)
```

Which as I re-read it seems rather contorted. Maybe "Skipping %d unavailable packages:" ?

leeper:
Just updated the commit to more closely mimic that style.

hadley:
Thanks!

PR #1009, opened 2016-01-08
Always force the promises when loading using load_all

jimhester:
As a continuation of #1001 you can still get lazy load errors if you use `load_all()` on a already loaded package for the same reason (promises never forced and then invalided when new code is loaded).

This was special cased for devtools so the functions would continue to work, but actually needs to be done for any package that is already loaded to avoid lazy load errors.

### DIFF
--- a/R/load.r
+++ b/R/load.r
@@ -94,11 +94,14 @@ load_all <- function(pkg = ".", reset = TRUE, recompile = FALSE,
   # in the DESCRIPTION file
   pkg$collate <- as.package(pkg$path)$collate
 
+  # Forcing all of the promises for the loaded namespace now will avoid lazy-load
+  # errors when the new package is installed overtop the old one.
+  #
   # Reloading devtools is a special case. Normally, objects in the
   # namespace become inaccessible if the namespace is unloaded before the
   # object has been accessed. Instead we force the object so they will still be
   # accessible.
-  if (pkg$package == "devtools") {
+  if (is_loaded(pkg)) {
     eapply(ns_env(pkg), force, all.names = TRUE)
   }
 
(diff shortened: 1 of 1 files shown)


R/load.r
```diff
@@ -94,11 +94,14 @@ load_all <- function(pkg = ".", reset = TRUE, recompile = FALSE,
   # in the DESCRIPTION file
   pkg$collate <- as.package(pkg$path)$collate
 
+  # Forcing all of the promises for the loaded namespace now will avoid lazy-load
+  # errors when the new package is installed overtop the old one.
```
wch: I think it should it be "loaded" instead of "installed"?

jimhester: You are correct, this comment was partially copied from `install()` and I missed that. Fixed with https://github.com/jimhester/devtools/commit/1ca9e288079fef5dabcc804d747ba6c2c12a8b84 Thanks!

R/load.r
```diff
@@ -94,11 +94,14 @@ load_all <- function(pkg = ".", reset = TRUE, recompile = FALSE,
   # in the DESCRIPTION file
   pkg$collate <- as.package(pkg$path)$collate
 
+  # Forcing all of the promises for the loaded namespace now will avoid lazy-load
+  # errors when the new package is loaded overtop the old one.
+  #
   # Reloading devtools is a special case. Normally, objects in the
```
hadley: Maybe delete this old comment?

jimhester: I kept that comment because for the devtools case we are forcing the promises to avoid the lazy load errors _and_ because it lets us continue to use the functions when devtools is being reloaded.

But I am fine removing it if it seems confusing.

hadley:
Thanks!

PR #1012, opened 2016-01-12
Show github API URL for information

Jaimyoung:
It helps to see the actual github API URL. Without this, it took me a while to find why my install_github() didn't work. It turned out to be due to the trailing space in, e.g. host="github.hostname.com/api/v3/" argument that leads to a double slashes in the final URL, e.g. "github.hostname.com/api/v3//jaimyoung/repo/zipball/master" that leads to rather uninformative error message:

client error: (403) Forbidden 

This function nicely shows "Source: https://github.hostname.com/api/v3//jaimyoung/repo/zipball/master" in the message to better reveal what went wrong.

### DIFF
--- a/R/install-github.r
+++ b/R/install-github.r
@@ -88,13 +88,14 @@ github_remote <- function(repo, username = NULL, ref = NULL, subdir = NULL,
 
 #' @export
 remote_download.github_remote <- function(x, quiet = FALSE) {
-  if (!quiet) {
-    message("Downloading GitHub repo ", x$username, "/", x$repo, "@", x$ref)
-  }
-
   dest <- tempfile(fileext = paste0(".zip"))
   src_root <- paste0("https://", x$host, "/repos/", x$username, "/", x$repo)
   src <- paste0(src_root, "/zipball/", x$ref)
+
+  if (!quiet) {
+    message("Downloading GitHub repo ", x$username, "/", x$repo, "@", x$ref,
+            "\nfrom URL ", src)
+  }
 
   if (!is.null(x$auth_token)) {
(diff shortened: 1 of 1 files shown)


R/install-github.r
```diff
@@ -110,6 +110,10 @@ remote_download.github_remote <- function(x, quiet = FALSE) {
     warning("GitHub repo contains submodules, may not function as expected!",
       call. = FALSE)
 
+  if (!quiet) {
```
hadley: Why not put everything in one message block?

Jaimyoung: That's absolutely fine. 
Then this message (a few lines above) then needs to happen after "src" string is assembled.

  if (!quiet) {
    message("Downloading GitHub repo ", x$username, "/", x$repo, "@", x$ref)
  }

That should work.

Jaimyoung: Actually, looking at the codes again, it might be better to keep this as a separate message block.
Between the first message and the new message, enough number of things happen, 1) dest tempfile created, 2) src URL built, 3) authenticated, 4) checking submodules. 
So it would be good to have the two messages separate.
Let me know what you think.

hadley: I'd prefer to have one message containing all the information

Jaimyoung:
Combine into a single message block per suggestion.
The message looks like:

> devtools::install_github(repo = "user/repo", host="github.host.com/api/v3/")
...
Downloading GitHub repo user/repo@master
from URL https://github.host.com/api/v3//repos/user/repo/zipball/master
...

hadley:
Thanks!

PR #1013, opened 2016-01-12
Fix dependency install when not upgrading

gaborcsardi:
It was broken. Correct me please if I am mistaken.

As I see it, `NA` means that the package is not installed / available, 
and if `upgrade=FALSE` holds, then we only want to install the
ones that are not installed locally. I.e. the ones with `is.na(installed)`.

(Btw. it seems that `devtools` ignores version requirements here, but 
that is another issue.)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,7 @@
 
 * `use_appveyor()` template now creates `failure.zip` artifact instead of
   polluting the logs with `R CMD check` output (#1017, @krlmlr, @HenrikBengtsson).
+
 * Fix a bug in the installers (`install_github`, etc.), when upgrades are
   not requested (#1013, @gaborcsardi).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,8 @@
 
 * `use_appveyor()` template now creates `failure.zip` artifact instead of
   polluting the logs with `R CMD check` output (#1017, @krlmlr, @HenrikBengtsson).
+* Fix a bug in the installers (`install_github`, etc.), when upgrades are
+  not requested (#1013, @gaborcsardi).
 
 * 'Check failed:' now includes the package name for when Ncpus>1 so you
    know which package has failed and can start looking at the output without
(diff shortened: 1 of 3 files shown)


hadley:
I think that logic holds. @jimhester can you please also check?

jimhester:
That looks correct to me, current behavior seems broken.

@gaborcsardi it would be great to have a test for this as well to prevent a regression in the future. Seems a little tricky to test though, you could supply a `parse_deps()` object to `update.package_deps()` with a mock `install_packages` function.

gaborcsardi:
Yeah, I am adding a test case in the `remotes` package, exactly the same way:
https://github.com/MangoTheCat/remotes/blob/800ac4486fba4e3e183559aac6c02ff818d007e8/tests/testthat/test-install-github.R#L263-L295
I can add it here, too.

Need some more time to get it right, though. Might be easier to create an object by hand, though.

gaborcsardi:
Btw. I found this while trying to get 100% test coverage for `remotes`. This was the single line of code not covered. :) 

`covr` FTW!

gaborcsardi:
I'll refactor the whole `compare_versions` code, because I just can't keep the `-2, ..., 2` codes in my mind.... they also seem to be broken (another way), but I am sure until I refactored it...... 

The refactoring can be in another PR, or here, as you wish.

hadley:
Here is fine

gaborcsardi:
I have added test cases. 

I have also added comments to `compare_versions` and related functions, and renamed `compare_versions` arguments, for readability. (Did not find more bugs.)

Rebased.

jimhester:
@gaborcsardi maybe we could just set some constants for the various numeric codes, which should make this more self-documenting. https://github.com/gaborcsardi/devtools/pull/1 has an implementation, although I did not use the constants in the tests.

Also I am not sure my choice of `UNINSTALLED` was the best, maybe `NOTINSTALLED` would be better?

hadley:
@jimhester I like that approach!

gaborcsardi:
Merged Jim's PR, and rebased.

PR #1015, opened 2016-01-12
make 'create', 'setup' more flexible (closes #966)

### LINKED ISSUE #966: make 'devtools::create()', 'devtools::setup()' more flexible
I find myself doing something like this often:

1. Create a new project in RStudio (but forget to create it as an R package),
2. Wish that I could call `devtools::create()` or `devtools::setup()` to turn it into an R package.

Two things to 'wish' for:

1. `devtools::create()` shouldn't fail for empty directories (or directories only a .Rproj file),
2. `devtools::setup()` shouldn't fail if a `.Rproj` file already exists.

kevinushey:
- 'devtools::create()' works when supplied an empty directory,
- 'devtools::setup()' no longer 'stop()'s when a .Rproj file
  already exists

### DIFF
--- a/R/infrastructure.R
+++ b/R/infrastructure.R
@@ -19,7 +19,8 @@ use_testthat <- function(pkg = ".") {
 
   check_suggested("testthat")
   if (uses_testthat(pkg)) {
-    stop("Package already has testing infrastructure", call. = FALSE)
+    message("* testthat is already initialized")
+    return(invisible(TRUE))
   }
 
   # Create tests/testthat and install file for R CMD CHECK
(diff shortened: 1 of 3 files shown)


R/infrastructure.R
```diff
@@ -74,7 +74,7 @@ use_rstudio <- function(pkg = ".") {
 
   path <- file.path(pkg$path, paste0(pkg$package, ".Rproj"))
   if (file.exists(path)) {
-    stop(pkg$package, ".Rproj already exists", call. = FALSE)
```
hadley: I think this should probably still print a `message()` - see the new behaviour for `use_git()` and `use_github()`

kevinushey: Done!

hadley:
LGTM - could you please also add a bullet to NEWS?

hadley:
Thanks!

PR #1017, opened 2016-01-13
AppVeyor template also creates Rcheck.zip as artifact

krlmlr:
instead of polluting the logs

Follow-up to #945.


hadley:
Thanks!

PR #1018, opened 2016-01-13
Set a GITHUB_PAT to limit Appveyor failures due to GitHub API limiting.

jimhester:
Because this is encrypted it is not active in PR, so I don't know how
useful this will actually be...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,6 +25,8 @@ on_failure:
 environment:
   WARNINGS_ARE_ERRORS: 1
   _R_CHECK_FORCE_SUGGESTS_: 0
+  GITHUB_PAT:
+    secure: 1WIIHQ7ce5IPmeRC2f/L0ze87IJi6jTFIWgX68mSFH5jym+QxZK2HLHQ85ImdkIU
 
 artifacts:
   - path: '*.Rcheck\**\*.log'


jimhester:
I actually think this might be better done just by manually adding an unencrypted GITHUB_PAT in the Appveyor settings. https://ci.appveyor.com/project/hadley/devtools/settings/environment. I don't have access to the appveyor settings, they don't inherit GitHubs settings apparently. See http://www.appveyor.com/docs/team-setup

If you just make a PAT (https://github.com/settings/tokens) without any scopes it should work fine for this purpose though.

jimhester:
I have addressed this in a different way (limiting the number of requests devtools makes https://github.com/hadley/devtools/commit/7e6a29023f1c2fce83d20754590f0c81a1ff85b8 https://github.com/hadley/devtools/commit/3251999b8fabcde110761e96284f4c8f0ff2e4f7. Still might be useful to set a PAT within appveyor though...

gaborcsardi:
I agree it is useful to set a PAT, my packages are regularly over the rate limit, too. I guess we are all competing with each other.....

jimhester:
Yeah the Appveyor folks must have only a few external IPs so you hit it pretty quickly, travis either has an exception from GitHub or uses many more external IPs.

PR #1019, opened 2016-01-13
More informative error message when a install_ function cannot be found.

jimhester:
Previously this gave the same error as other Malformed remotes, this should point users to the proper problem.

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -174,7 +174,7 @@ dev_remote_type <- function(remotes = "") {
         mode = "function",
         inherits = FALSE),
       error = function(e) {
-        stop("Malformed remote specification '", x, "', 'devtools:::", fun_nme, "' does not exist", call. = FALSE)
+        stop("Unknown remote type", call. = FALSE)
       })
     list(repository = repo, type = type, fun = fun)
   }
--- a/R/deps.R
+++ b/R/deps.R
@@ -167,13 +167,14 @@ dev_remote_type <- function(remotes = "") {
     } else {
       stop("Malformed remote specification '", x, "'", call. = FALSE)
     }
+    fun_nme <- paste0("install_", tolower(type))
     tryCatch(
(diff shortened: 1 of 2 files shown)


R/deps.R
```diff
@@ -167,13 +167,14 @@ dev_remote_type <- function(remotes = "") {
...
+    fun_nme <- paste0("install_", tolower(type))
     tryCatch(
-      fun <- get(x = paste0("install_", tolower(type)),
+      fun <- get(x = fun_nme,
         envir = asNamespace("devtools"),
         mode = "function",
         inherits = FALSE),
       error = function(e) {
-        stop("Malformed remote specification '", x, "'", call. = FALSE)
+        stop("Malformed remote specification '", x, "', 'devtools:::", fun_nme, "' does not exist", call. = FALSE)
```
hadley: How about just `stop("Unknown remote type")`?

jimhester: Works for me, done in https://github.com/jimhester/devtools/commit/7b378ee0e6a2afcbc21172e61a790f613f335b15

R/deps.R
```diff
@@ -167,13 +167,14 @@ dev_remote_type <- function(remotes = "") {
...
+    fun_nme <- paste0("install_", tolower(type))
     tryCatch(
-      fun <- get(x = paste0("install_", tolower(type)),
+      fun <- get(x = fun_nme,
         envir = asNamespace("devtools"),
         mode = "function",
         inherits = FALSE),
       error = function(e) {
-        stop("Malformed remote specification '", x, "'", call. = FALSE)
+        stop("Unknown remote type", call. = FALSE)
```
hadley: And maybe put the type into just to be helpful?

jimhester: Done! https://github.com/hadley/devtools/commit/f4151b9da8e68051eadc4e9cc452648c520585e6

hadley:
Thanks!

PR #1021, opened 2016-01-13
Lookup code dynamically rather than copy directly

jimhester:
Previously this code was copied from `base::loadNamespace`, this instead extracts the necessary code directly from the `loadNamespace` definition.

The previous code had drifted some from what was in base R, this will prevent this in the future.

This is in effort to address build failures on windows (https://ci.appveyor.com/project/hadley/devtools/build/1.0.172#L310) although that effort is still in progress.


jimhester:
I guess one bad consequence of this approach is it generates a `WARNING` from R CMD check because the code from `loadNamespace` has calls to internal `methods` functions. https://ci.appveyor.com/project/hadley/devtools/build/1.0.174#L263. Could get rid of them but it would require more meta-programming :frowning:.

I am still not sure why we are getting failures on windows, still working on it.

wch:
I like this general approach -- there's a bunch of stuff that devtools borrowed from the R sources that could be retrieved dynamically like this.

To avoid the R CMD check warning, you could get the code either at run time or load time.

hadley:
Also note, but never talk about, `%:::%` :wink:

jimhester:
In this case we are setting the enclosing environment of the function to methods anyway, so I could just strip the `methods:::` calls from the code (https://github.com/jimhester/devtools/commit/9c3eb0b7495f39c8f82aed709aca19188069de98).

hadley:
I'm not sure if this has ended up any simpler/better than the previous code. What do you think?

jimhester:
I think the approach has merit, when we need to copy code it seems better if it is done programmatically rather than by hand.

I think we can wait to merge this PR though. I want to use this same idea for other code in devtools so I will add to this PR as I go.

hadley:
@wch what other functions need to be in this PR?

wch:
Here's what I found. I think the categories below are correct, but I could be wrong about some of them:

These functions are taken verbatim from R:

* `add_classes_to_exports`
* `addNamespaceDynLibs`

These functions are almost verbatim, but with very small changes:

* `makeNamespace` - there is one change, to call `register_namespace()` instead of `.Internal(registerNamespace(name, env))`, but that might not be necessary if the code is extracted dynamically from R.
* `assignNativeRoutines`
* `process_imports`

These functions contain sections that are taken from R. For some, it might make sense to spin those parts out into separate function, but not for all:

* `setup_ns_exports`
* `load_dll`
* `library.dynam2`

wch:
I think the code extraction should be done in `.onLoad`, so that R CMD check doesn't complain about stuff like `:::` and internal function calls. I *think* doing it at load time will avoid those problems but I don't know for sure -- I don't know at exactly what point R CMD check scans the parse tree for those problems.

Doing extraction at load time will also ensure that the extracted code matches the user's version of R. With the current code in this PR, the extraction happens in the bundled package -> binary package phase, so a binary package on CRAN might extract code from (say) R 3.2.4, but the user may try run it on 3.2.0.

jimhester:
I am now using `delayedAssign()` to do the assignment, which ensures the functions will only be extracted from the currently running version of R. This also removes any issues with R CMD check.

Current progress

- [x] add_classes_to_exports
- [x] addNamespaceDynLibs
- [ ] makeNamespace
- [ ] assignNativeRoutines
- [ ] process_imports
- [ ] setup_ns_exports
- [ ] load_dll
- [ ] library.dynam2

wch:
Some care is needed when using `delayedAssign` at the top level of package code. In the past we had used it in Shiny, but it caused problems because it didn't behave the way we had expected. According to what I wrote in some emails, the promises will be evaluated on package load, not when the symbol is first accessed, for reasons that aren't entirely clear to me.

Being evaluated on package load is safe, but I could have been wrong before about the exact time when that happens. It would be a good idea to check that they don't get evaluated during the package source -> package bundle phase, or during the package bundle -> binary package phase. (Maybe add a `cat("foo", file=stderr())` to the expression and see when it gets printed.)

See rstudio/shiny@d7eb9b2, and if you want a bit more information, I can forward some of the old emails relating to the use of `delayedAssign` at the top level of a package.

jimhester:
@wch Thanks for the background winston, I was actually just running into some strange issues with `delayedAssign()` and `assignNativeRoutines` so am going to try to do it in .onLoad() instead.

wch:
I guess there's also the issue that bindings in a package namespace are locked so you can't modify them with `.onLoad`. You'd probably have to store them in an environment.

hadley:
I suspect `delayedAssign()` at the top-level of a package gets caught in the package lazy-load mechanism that's used for data sets

jimhester:
Actually the package namespace is not locked until after `.onLoad()` runs, so you can actually assign to the package namespace. We are currently doing this already for the `withr::` functions (https://github.com/hadley/devtools/blob/master/R/load-dll.r)

wch:
Oh cool, that simplifies things a bit.

jimhester:
Ok this is probably ready to review now, `setup_ns_exports` and `library.dynam2` are both different enough from the base R code I don't think there is much we can extract and all the others have been done.

Any thoughts on making this code more clear would be appreciated, doing meta-programming is always a little tricky to understand at first glance.

R/imports-env.r
```diff
@@ -65,33 +65,30 @@ load_imports <- function(pkg = ".") {
...
+      vI <- .split_description(.read_description(file.path(pkg$path, "DESCRIPTION")))$Imports
+      nsInfo <- devtools::parse_ns_file(pkg)
+      ns <- devtools::ns_env(pkg)
+      lib.loc <- NULL
+      tryCatch(error = warning, {
+        .(for1)
+        .(for2)
+        .(for3)
+      })
+    }, list(
```
hadley: That's a little different to before - previously the `tryCatch()` covered a much smaller scope. That's important so that `load_all()` can continue to (mostly) work if the NAMESPACE isn't quite right

R/imports-env.r
```diff
@@ -65,33 +65,30 @@ load_imports <- function(pkg = ".") {
...
+        for1 = extract_lang(body(loadNamespace),
+          function(x) length(x) > 2 && identical(x[1:3], quote(for(i in nsInfo$imports) NULL)[1:3]))[[1]],
```
hadley: This seems like a common pattern - extract into a function?

```R
comp_lang <- function(x, y) {
  if (length(x) < length(y)) return(FALSE)

  identical(x[1:length(y)], y)
}
```

Or maybe that's just how `extract_lang` should work if the 2nd arg is a call?

And maybe `extract_lang` should get an `i` argument?

hadley: I'm not sure what the `[[1]]` does

jimhester: The `extract_lang()` function always returns a list of functions it found, so just returning the first item. `extract_lang()` should probably do this automatically...

R/imports-env.r
jimhester: The only increased scope is the for loop indexing though, which is just a list, so if `nsInfo$imports` doesn't exist it will just return `NULL`, which won't throw any conditions.

I agree in general it is nice to keep the scope small, in this case I don't think it can change the behavior and it will be quite a bit more work to modify each for loop appropriately.

hadley: The problem is if one of the imports is badly formed - before it would still import everything after that

jimhester: Ah I see now, because the tryCatch is around each iteration in the loop rather than the whole thing.

Ok I will get that working as before then.

jimhester:
I modified the calls to use the new `comp_lang()` function which I think cleaned things up a bit. Also moved the tryCatch blocks to inside the for loops.

hadley:
I think this is fine, provided @wch agrees.  

I also wonder if we should capture in one place if any of the replacement fails, and throw a single warning on attach - "Devtools is incompatible with the current version of R. `load_all()` may function incorrectly." Otherwise we'll never find out.

If we do more of this it might be worthwhile to write `ast_patch()` which does some sort of templatised search-and-replace on ASTs.  Maybe we need a whole package that provides xml2 like operations on ASTs?

wch:
It looks good to me. I agree that a compatibility check and warning is a good idea.

jimhester:
I added the suggested warning and some simple tests to prevent regressions in the future.

jimhester:
I added the warning as well as some simple tests.

hadley:
LGTM

wch:
👍

PR #1023, opened 2016-01-14
check for existence of remote tracking branch in git_sync_status(); fixes #1000

### LINKED ISSUE #1000: Error in git2r::lookup with devtools::release()
I'm getting this error when running devtools::release()

    Error in git2r::lookup(r, git2r::branch_target(upstream)) : 
      error in evaluating the argument 'sha' in selecting a method for function 'lookup': Error in (function (classes, fdef, mtable)  : 
      unable to find an inherited method for function ‘branch_target’ for signature ‘"NULL"’

This is using the current CRAN versions of devtools and git2r, and R versions 3.2.3 on OSX 10.11.1. 

I've tried using the github versions of devtools and git2r, but I get the same error. 

dr_devtools() says that I look healthy. Any ideas what's going on?

jennybc:


hadley:
Thanks!

PR #1024, opened 2016-01-14
Install sha even on local install

rmflight:
I noticed that `install_local` will add the SHA to a DESCRIPTION file, but calling `install` does not. Given that I (and I would imagine lots of others) frequently **install** a package that is being personally developed using `install` instead of `install_local`, and the SHA information points to a unique version of the package that is recorded in `session_info`, then it seems a natural thing to do.

This pull request adds the new method `remote_metadata.package` so that the SHA can be obtained from a normal package, and then adds a parameter `add_sha` to the `install` function, so users can decide if they want the SHA added or not. I also added `add_sha=FALSE` to some of the other tests that use `install` so the DESCRIPTION files don't get modified. I also added tests that the SHA is now being recorded in `session_info`.

Travis is having difficulty with any `R` based builds, but this passes all tests on my local machine.


jimhester:
One thing that makes this problematic is it assumes the version you are installing is actually committed. If you have not actually committed the changes the SHA will still be the same as last committed version even though the installed code will be different.

Maybe we could check if the `git status` shows no modifications or uncommitted files before adding the SHA?

rmflight:
Fair enough. I'll add that in, and think about the tests to check it.

jimhester:
You can use something like the following
```r
r <- git2r::repository(".", discover = TRUE)
s <- git2r::status(s)
#> working directory clean
str(s)
#>  $ staged   : Named list()
#>  $ unstaged : Named list()
#>  $ untracked: Named list()
#>  - attr(*, "class")= chr "git_status"

all(vapply(git2r::status(r), length, integer(1)) == 0)
#> [1] TRUE

# Putting this in a function the default argument for add_sha could be `add_sha = is_clean(pkg)`
is_clean <- function(path) {
  r <- git2r::repository(".", discover = TRUE)
  all(vapply(git2r::status(r), length, integer(1)) == 0L))
}
```

I am not sure `is_clean` is a great name for the function, probably too generic...

rmflight:
This is something that should be done on any **local** install, correct? So it would apply to both `install` and `install_local`, because this property could occur in both cases. So it would go in the logic for `remote_metadata` for `local` and `package`, and perhaps give a **warning**, but continue to install?

https://github.com/rmflight/devtools/blob/install_sha/R/install-local.r#L44

I like the idea of maybe having a `clean_tree` function that is the default argument to `add_sha`. I'll try a couple of things and see what works, especially for both `install` and `install_local`

rmflight:
what about `is_wd_clean` for the function? where `wd` is `working directory`, this fits with the message from git on a clean repo of: **working directory clean**

rmflight:
Poking around the `git.R` file, there is the function `git_uncommitted` (https://github.com/rmflight/devtools/blob/install_sha/R/git.R#L11), but it is currently unexported. Should I use this function and export it, or create an alias function with a nicer name that is exported? I like `is_wd_clean` or `git_wd_clean` as functions that match up to the status messages that git supplies with `git status`

jimhester:
You can't use `git_uncommitted` directly as it throws an error if you are not in a git repository, so you will have to write a wrapping function that calls `uses_git()` first.

```r
git_wd_clean <- function(path = ".") {
  uses_git(path) && !git_uncommitted(path)
}
```

You shouldn't actually have to export this function IIRC

rmflight:
OK, now I remember why I didn't do that as a default, and why it is hard. Modifying the DESCRIPTION with the SHA results in a dirty working directory. And so on and so forth. I can do it if I can add an argument to `git_uncommited` to skip certain files, in this case DESCRIPTION

rmflight:
Nope, never mind, just realized I got my logic completely backwards. WD should be clean if all changes are committed, so we do the check, and **then** make the modification. So this shouldn't be that big an issue, but I just spent past hour thinking it was. Will have to undo and then try again. Will try later, going out atm.

rmflight:
Added the modifications we talked about. All tests pass. If you think the general approach is good, then I would need to add documentation on the `devtools.git.wd.clean` **option**, and I maybe should add a couple of more tests that things do continue working properly when called from `install`, but overall I think it works well.

jimhester:
I personally think you should keep things simple and make the `git_wd_clean()` implementation just

```r
git_wd_clean <- function(path = ".") {
  uses_git(path) && !git_uncommitted(path)
}
```

We really want to err on the side of caution with this and only add a SHA reference if we are sure everything in the working directory is committed. Particularly once https://github.com/hadley/devtools/pull/903 is merged and doesn't reinstall packages if the SHA is the same as the previous install. If a user really wants to ignore a dirty working directory they could always just use `add_sha = TRUE`.

Installing without a SHA is the current behavior and works fine, so I don't see much downside on being conservative and only adding one if we are sure it is valid.

@hadley any thoughts?

rmflight:
Yeah, but it gets more complicated if we want it to revert to `local` from `local (/@sha)` when there is a dirty working directory. The current method modifies the DESCRIPTION file in place, and then leaves it there. I have made another branch (rmflight/devtools/install_sha2) that I think is more along the lines of what you proposed in the first place, @jimhester.

Do we want the behavior of going from `local (/@sha)` to `local` if the working directory is dirty? If so, that may take a bit more work. If so, I'd also take feedback on how to do that.

My initial thought is to copy DESCRIPTION, modify the one in place, install, then put original DESCRIPTION back. This should keep `git` from complaining too much as well.

hadley:
Quick thoughts:

* @jimhester's implementation of `git_wd_clean()` is adequate, and doesn't need to be exported

* `git_wd_clean()` would be better called `git_committed()` or similar

* You shouldn't need to modify the description file. Why can't you use the same approach `install_github() ` uses to add extra metadata?

rmflight:
OK, I was basing my implementation on `install_local`, which does modify
the DESCRPTION file, but I will look again at the implementations.

Will take a look at all this again tonight.

rmflight:
OK, have confirmation that the actual DESCRIPTION file is modified.

Based on the `master` branch, I modified `install_remote` with a `readLines()` and a `cat` of the DESCRIPTION file before and after `add_metadata`.

Then I did a call to `install_github("rmflight/categoryCompare2")`

**before `add_metadata`**

```
Package: categoryCompare2
Version: 0.99.158
Title: Meta-analysis of high-throughput experiments using feature
        annotations
Author: Robert M. Flight <rflight79@gmail.com>
Maintainer: Robert M. Flight <rflight79@gmail.com>
URL: https://github.com/rmflight/categoryCompare2
BugReports: https://github.com/rmflight/categoryCompare2/issues
License: GPL-2
Depends: R (>= 3.1.0), Biobase, BiocGenerics
Suggests: knitr, methods, GO.db, estrogen, org.Hs.eg.db, hgu95av2.db,
        limma, affy, genefilter, testthat
Imports: AnnotationDbi, hwriter, GSEABase, Category, GOstats, annotate,
        colorspace, graph, RCytoscape (>= 1.5.11)
LazyLoad: yes
Description: Facilitates comparison of significant annotations
        (categories) generated on one or more feature lists.
        Interactive exploration is facilitated through the use of
        RCytoscape (heavily suggested).
SystemRequirements: Cytoscape (>= 2.8.0) (if used for visualization of
        results, heavily suggested), CytoscapeRPC plugin (>= 1.8)
biocViews: Annotation, GO, MultipleComparison, Pathways, GeneExpression
VignetteBuilder: knitr
```

**after `add_metadata`**

```
Package: categoryCompare2
Version: 0.99.158
Title: Meta-analysis of high-throughput experiments using feature
        annotations
Author: Robert M. Flight <rflight79@gmail.com>
Maintainer: Robert M. Flight <rflight79@gmail.com>
URL: https://github.com/rmflight/categoryCompare2
BugReports: https://github.com/rmflight/categoryCompare2/issues
License: GPL-2
Depends: R (>= 3.1.0), Biobase, BiocGenerics
Suggests: knitr, methods, GO.db, estrogen, org.Hs.eg.db, hgu95av2.db,
        limma, affy, genefilter, testthat
Imports: AnnotationDbi, hwriter, GSEABase, Category, GOstats, annotate,
        colorspace, graph, RCytoscape (>= 1.5.11)
LazyLoad: yes
Description: Facilitates comparison of significant annotations
        (categories) generated on one or more feature lists.
        Interactive exploration is facilitated through the use of
        RCytoscape (heavily suggested).
SystemRequirements: Cytoscape (>= 2.8.0) (if used for visualization of
        results, heavily suggested), CytoscapeRPC plugin (>= 1.8)
biocViews: Annotation, GO, MultipleComparison, Pathways, GeneExpression
VignetteBuilder: knitr
RemoteType: github
RemoteHost: api.github.com
RemoteRepo: categoryCompare2
RemoteUsername: rmflight
RemoteRef: master
RemoteSha: 0478ba3af9914f0df7598b47c7ccc9d9c3c91de8
GithubRepo: categoryCompare2
GithubUsername: rmflight
GithubRef: master
GithubSHA1: 0478ba3af9914f0df7598b47c7ccc9d9c3c91de8
```

This works OK on *github* remotes because who cares about modifying the downloaded source that will get redownloaded. 

`install_local` also takes the directory, and copies the whole thing off to a temp directory, and then does `install_remote` on it, from what I can tell https://github.com/hadley/devtools/blob/master/R/install-local.r#L19

rmflight:
Yep, confirmed that `install_local` first copies everything off to a temp directory, and then does the same process as `install_remote`. I'll think about this some more and think of which function can be modified without breaking anything.

PR #1027, opened 2016-01-15
Install sha (second attempt)

rmflight:
@jimhester, based on our earlier conversations, here is a second attempt at adding the SHA1 to the package install. It uses `git_committed` as we discussed as the default function of `add_sha`, and then if that passes, will then use `install_local` to copy the package to a temp directory and do the install from there, which automatically adds the SHA1.

Note that the call to `install_local` probably needs to have all the other `install` arguments added to it to truly work as intended, but this currently passes all the tests, including the ones I added. 

If this passes muster, I will add the rest of the install arguments to the `install_local` call.

The only ugly part of this I don't like is the `return()` statement that keeps a second round of `install` from happening and installing the one in the current directory, because that causes a `NULL` return value. I don't know how to avoid that ATM.


jimhester:
Here's an alternative plan that might make things simpler. How about instead of having `add_metadata()` add the metadata to the `DESCRIPTION` _before_ it gets installed (https://github.com/hadley/devtools/blob/master/R/install-remote.R#L20) we add a `metadata=` argument to `install` and have the remotes pass the metadata to it and add it _after_ installation (e.g. at https://github.com/hadley/devtools/blob/master/R/install.r#L106).

That removes the whole chicken and the egg issue and we can just write a function for `metadata=` that by default checks if the git working directory is clean and adds the SHA metadata if so.

jimhester:
See proof on concept https://github.com/jimhester/devtools/commit/aa15cd0f3e9e8c8eadfc8e9c962f4bd958ec060f which works with limited testing on my machine.

jimhester:
Also note this is using `remote_metadata.local_remote` directly, but we will probably want to define a new function that uses your `git_committed()` function instead of just `uses_git()` in the [if statement](https://github.com/hadley/devtools/blob/master/R/install-local.r#L46).

rmflight:
OK, I like this idea, but have some questions:

1. I didn't think what I ended up doing in this pull request was **that** complicated. Sure it is a 2 pass call to `install`, but it keeps it doing what it does best.

2. This is completely different than the approach that all of the other `install_` use to add metadata, and I worry about debugging that in the future

3. I thought it was bad practice for functions to be able to modify system files once they are copied into the system. From that perspective I would worry about this not passing a CRAN review.

I'm willing to implement and properly test your suggestion, but just wanted to bring these up.

jimhester:
The idea would be to change the other `install_` functions to pass their `remote_metadata` calls to `install`, everything would use the same path, so nothing would need to be special cased.

3. Maybe relevant, but as long as it doesn't generate a NOTE from R CMD check it should be ok.

rmflight:
OK, I like that idea even better.

Final point, add to this pull request, or start over and submit a new one?

jimhester:
Sorry I was not more clear in my proof on concept what I had in mind.

I would keep this pull request. Actually I would have kept the previous one open so that the discussion was all together, you can always 
```shell
# reset state to previous state
git reset upstream/master
# make new commits
...
git push --force
```
To remove the old commits if needed.

rmflight:
So currently, the `remote_metadata` is an S3 function, with methods for `github_remote`, `local_remote`, etc. So I would see adding a new method for `package`, and then have something like:

```
install <- functioin(...., metadata = git_committed){

if (is.function(metadata)) {
  metadata <- git_committed(pkg$path)
} 

R CMD stuff

if (metadata) {
  new_metadata <- remote_metadata(pkg)
  add_metadata(base::system.file(package = pkg$package), new_metadata)
}
```

Does that seem reasonable? I'll start actually coding something up.

jimhester:
I would really follow my proof on concept. Just write a new method for `package`

```r
remote_metadata.package <- function(x, bundle = NULL, source = NULL) {
  list(
    RemoteType = "package",
    RemoteUrl = x$path,
    RemoteSha = if (git_committed(x$path)) git_sha1(path = x$path)
  )
}
```
Make the default argument for install `metadata = remote_metadata(as.package(pkg))` and add it after the R CMD check.
```r
if (length(metadata)) {
  add_metadata(base::system.file(package = pkg$package), metadata)
}
```
Then just change https://github.com/hadley/devtools/blob/master/R/install-remote.R#L20-L25 to pass the metadata to `install` rather than adding it beforehand.
```diff
diff --git a/R/install-remote.R b/R/install-remote.R
index b1192b4..a22232a 100644
--- a/R/install-remote.R
+++ b/R/install-remote.R
@@ -17,12 +17,13 @@ install_remote <- function(remote, ..., quiet = FALSE) {
   source <- source_pkg(bundle, subdir = remote$subdir)
   on.exit(unlink(source, recursive = TRUE), add = TRUE)

-  add_metadata(source, remote_metadata(remote, bundle, source))
-
   # Because we've modified DESCRIPTION, its original MD5 value is wrong
   clear_description_md5(source)

-  install(source, ..., quiet = quiet)
+  install(source,
+    metadata = remote_metadata(remote, bundle, source),
+      ..., quiet = quiet)
+
 }
```

You actually don't have to change any of the `install_` functions directly.

jimhester:
https://github.com/jimhester/devtools/commit/212d272b9726529b034ee6329da856f0c68d16ea has a working implementation passing the regular devtools tests (https://travis-ci.org/rmflight/devtools/builds/102617084#L2277-L2279). It just needs documentation and some additional tests written if possible.

rmflight:
OK, trying it out, and there is an issue. `packageDescription` (which `session_info` uses to get package information` does not read from DESCRIPTION directly, but rather reads `package.rds`, which looks like it is generated at INSTALL. This gets read by `readRDS`

rmflight:
I've got an implementation, but the test indicates that it doesn't work at modifying the information that shows up in `session_info`, which is what I think has to happen for this to be useful.

rmflight:
As I noted, https://github.com/hadley/devtools/commit/70b9b25890ca1e89d902d6bec83a8591e4d75c56#diff-17d36a2a2b0eac6b55b22dae5787dc48R167, this test fails because `packageDescription` uses `readRDS`. So we either need to change the `add_metadata` to modify both the DESCRIPTION and the `package.rds` file, or change `session_info` to use `read.dcf` in the package location.

Do you or @hadley know offhand if there is a function for writing RDS files outside of what INSTALL does?

rmflight:
just found it, looks like `saveRDS`

rmflight:
I also think I might need to create some tests just for `metadata` checking

jimhester:
Mmm yes that is an issue, see https://github.com/jimhester/devtools/commit/bd0cff1e906f647b19a03005267fe70f821b4b58 for code which updates it as well as the `DESCRIPTION`.

rmflight:
OK, this works, and passes all of the previous `devtools` tests, and passes the new ones I wrote to check that the SHA1 is being written in.

R/install-local.r
```diff
@@ -46,3 +46,13 @@ remote_metadata.local_remote <- function(x, bundle = NULL, source = NULL) {
     RemoteSha = if (uses_git(x$path)) git_sha1(path = x$path)
   )
 }
+
+#' @export
+remote_metadata.package <- function(x, bundle = NULL, source = NULL) {
```
jimhester: If we change this to the following it will give us consistent results with the legacy behavior if the package is not using git or has an unclean working directory.

```r
compact(list(
    RemoteType = "local",
    RemoteUrl = x$path,
    RemoteSubdir = x$subdir,
    RemoteSha = if (git_committed(x$path)) git_sha1(path = x$path)
  ))
```

jimhester: Actually I am wrong, just looked at the code for `pkg_source` https://github.com/hadley/devtools/blob/81dd313e1d143a3bdaaa2e1f09c887801db66694/R/session-info.r#L156-L159

jimhester:
Ok this is looking good, I think we should modify https://github.com/hadley/devtools/blob/81dd313e1d143a3bdaaa2e1f09c887801db66694/R/session-info.r#L156-L159 to work with this situation better. Right now the code assumes there is always a `RemoteUsername` and `RemoteRepo` if there is a `RemoteType`, which in this case (and when using install_local`) that is not true.

rmflight:
In for a penny, in for a pound. May as well make it consistent somewhere else.

hadley:
It would be super useful if you could also fill in the github info if the repo has been pushed to github. (That would make my life easier when working with shinyapps and development versions of my packages)

rmflight:
That sounds really useful and cool @hadley, but can we maybe make an issue and assign it to me, and I can work on that particular feature some other time?

I want to get the reporting in `session_info` cleaned up, and then have this as a stand alone pull request and be done with it at this point.

rmflight:
ok, rmflight@bb9e337 has a modified `session_info` that will generate:

from github:
categoryCompare2    * 0.99.158   2016-01-15 Github (rmflight/categoryCompare2@0478ba3)

from local git with clean tree and `metadata` not NULL:
testMetadataInstall * 0.0.0.9000 2016-01-15 package (@373e621)  

and then local git with a dirty working directory:
testMetadataInstall * 0.0.0.9000 2016-01-15 package               

does that work for you @jimhester ?

jimhester:
This should fill in the github user/repo information if the repository has a github remote set up. It doesn't necessarily mean the given revision is actually pushed to that remote however...

```diff
diff --git a/R/install-local.r b/R/install-local.r
index 6877f5d..b7d843e 100644
--- a/R/install-local.r
+++ b/R/install-local.r
@@ -49,10 +49,18 @@ remote_metadata.local_remote <- function(x, bundle = NULL, source = NULL) {

 #' @export
 remote_metadata.package <- function(x, bundle = NULL, source = NULL) {
-  list(
-    RemoteType = "package",
-    RemoteUrl = x$path,
-    RemoteSubdir = x$subdir,
-    RemoteSha = if (git_committed(x$path)) git_sha1(path = x$path)
+  res <- list(
+    RemoteType = "local",
+    RemoteUrl = x$path
   )
+
+  if (git_committed(x$path)) {
+    res$RemoteSha <-  git_sha1(path = x$path)
+  }
+  if (uses_github(x$path)) {
+    info <- github_info(x$path)
+    res$RemoteUsername <- info$username
+    res$RemoteRepo <- info$repo
+  }
+  res
 }
diff --git a/R/session-info.r b/R/session-info.r
index 9bdba3d..571970a 100644
--- a/R/session-info.r
+++ b/R/session-info.r
@@ -158,7 +158,7 @@ pkg_source <- function(desc) {
     user_repo <- character(0)

     # in current definitions these should always be together
-    if (!is.null(desc$RemoteUsername) && (!is.null(desc$RemoteRepo))) {
+    if (!is.null(desc$RemoteUsername) && !is.null(desc$RemoteRepo)) {
       user_repo <- paste0(desc$RemoteUsername, "/", desc$RemoteRepo)
     }
```

jimhester:
Also I am not sure having a `RemoteType = 'package'` is useful, which is why I switched it back to 'local' in the diff.

rmflight:
OK, fair enough on the `package` vs `local`. 

If changing this, I feel like `remote_metadata.local_remote` and `remote_metadata.package` should be doing the exact same thing as far as annotating the metadata.

jimhester:
If the two functions are literally the same I think we should just make them aliases.
```r
remote_metadata.package <- remote_metadata.local_remote
```
Also could you rebase this on the current master and possibly squash some of the more trivial commits. https://help.github.com/articles/about-git-rebase/ has some help on using Git rebase if you need a reference.

rmflight:
OK. Originally went with two functions so it would be easier to change in the future if decided they should be different, but you are right, currently they are the same, so should just alias one from the other.

I will give rebasing an attempt, this is the first time I've tried to use it.

jimhester:
Just remember you can always abort the rebase with `git rebase --abort` if needed. If you are worried about messing something up you can checkout the current state into a new branch, then switch back. The backup won't be altered by the rebase.

```
git checkout install_sha
git checkout -b backup
git checkout install_sha
```

rmflight:
Thanks, that helps.

rmflight:
OK, successfully rebased onto current devtools/master

R/git.R
```diff
@@ -14,6 +14,10 @@ git_uncommitted <- function(path = ".") {
   any(st != 0)
 }
 
+git_committed <- function(path = ".") {
```
hadley: Given that this function is not just the negation of `git_committed()` I think it's confusing.

R/install-local.r
```diff
@@ -39,10 +39,21 @@ remote_download.local_remote <- function(x, quiet = FALSE) {
...
-  list(
+  res <- list(
     RemoteType = "local",
-    RemoteUrl = x$path,
-    RemoteSubdir = x$subdir,
-    RemoteSha = if (uses_git(x$path)) git_sha1(path = x$path)
+    RemoteUrl = x$path
   )
+
+  if (git_committed(x$path)) {
```
hadley: Just do `uses_git(x$path) && !git_uncommitted(x$path)` here

R/install.r
```diff
@@ -61,6 +62,8 @@ install <- function(pkg = ".", reload = TRUE, quick = FALSE, local = TRUE,
                     keep_source = getOption("keep.source.pkgs"),
                     threads = getOption("Ncpus", 1),
                     force_deps = FALSE,
+                    add_sha = git_committed,
```
hadley: Is this argument used anywhere?

R/install-remote.R
```diff
@@ -17,12 +17,12 @@ install_remote <- function(remote, ..., quiet = FALSE) {
   source <- source_pkg(bundle, subdir = remote$subdir)
   on.exit(unlink(source, recursive = TRUE), add = TRUE)
 
-  add_metadata(source, remote_metadata(remote, bundle, source))
+  metadata <- remote_metadata(remote, bundle, source)
 
   # Because we've modified DESCRIPTION, its original MD5 value is wrong
   clear_description_md5(source)
```
hadley: I think this now needs to move

R/install.r
```diff
@@ -47,6 +47,7 @@
 #'   It defaults to the option \code{"Ncpus"} or \code{1} if unset.
 #' @param force_deps whether to force installation of dependencies even if their
 #'   SHA1 reference hasn't changed from the currently installed version.
+#' @param metadata metadata to add to \code{DESCRIPTION}.
```
hadley: Need more details here. What sort of object is it? How does the addition happen?

R/install.r
```diff
@@ -107,6 +110,10 @@ install <- function(pkg = ".", reload = TRUE, quick = FALSE, local = TRUE,
   R(paste("CMD INSTALL ", shQuote(built_path), " ", opts, sep = ""),
     quiet = quiet)
 
+  if (length(metadata)) {
```
hadley: I prefer to be explicit: `length(metadata) > 0`

R/install-remote.R
```diff
@@ -32,11 +32,23 @@ install_remotes <- function(remotes, ...) {
...
+    desc <- read_dcf(path)
+
+    desc <- modifyList(desc, meta)
+
+    write_dcf(path, desc)
+  }
 
-  desc <- modifyList(desc, meta)
+  path <- file.path(pkg_path, "Meta", "package.rds")
+  if (file.exists(path)) {
```
hadley: This needs to be an if else:

```R
if (file.exists(source_desc)) { 

} else if (file.exists(binary_desc)) { 

} else {
  stop("No DESCRIPTION found.", call. = FALSE(
}
```

R/install-remote.R
rmflight: so remove the `clear_description_md5`, as it is no longer needed because the metadata is not being added **before** install?

R/install-local.r
rmflight: done in 0688b7e

R/install.r
rmflight: Sorry, missed this while making merges in rebasing ontop of current master

R/install.r
rmflight: so something like:

```
#' @param metadata \code{list} of metadata (normally generated by \code{remote_metadata}) to 
#'    add to the DESCRIPTION file after installation.
```

R/install-remote.R
jimhester: I actually did this without an else on purpose because I wanted to add the metadata to _both_ the binary description as well as the text description in the same way. Some functions only look at the text DESCRIPTION, others will use the binary if it is available, so I wanted the same results regardless.

rmflight: Yes, some functions will look to one or the other, so both files (if present) need to be modified. It might be a good idea to add a warning that neither file exists and metadata couldn't be added?

```
if (file.exists(source_desc)) {}

if (file.exists(binary_desc)) {}

if (!file.exists(source_desc) && !file.exists(binary_desc)) {
    warning("No DESCRIPTION found, metadata not added!", call. = FALSE)
}
```

jimhester: I think it should be an error if neither can be found, but agree it is a good idea to check that case.

rmflight: no problem.

rmflight: done in 30f9d406

R/install-remote.R
rmflight: removed call to `clear_description_md5` in 6467f7c, running tests suggests it doesn't break anything. I wonder if the function is needed at all in the code anymore? I think this was the only place it was being called.

rmflight:
Given that this makes the `metadata` a first class thing that the user can see, should the various `remote_metadata` methods be documented as far as what **metadata** will be added to the DESCRIPTION?

R/install-remote.R
```diff
@@ -31,12 +28,28 @@ install_remotes <- function(remotes, ...) {
...
-  desc <- read_dcf(path)
+  source_desc <- file.path(pkg_path, "DESCRIPTION")
+  binary_desc <- file.path(pkg_path, "Meta", "package.rds")
+  if (file.exists(source_desc)) {
+    desc <- read_dcf(source_desc)
+
+    desc <- modifyList(desc, meta)
+
+    write_dcf(source_desc, desc)
+  }
```
hadley: Still missing the else clauses here

hadley: Oh, I see you added that below. But that makes it unclear that there are three options - I'd prefer the if else if else structure I suggested.

R/install.r
```diff
@@ -47,6 +47,8 @@
 #'   It defaults to the option \code{"Ncpus"} or \code{1} if unset.
 #' @param force_deps whether to force installation of dependencies even if their
 #'   SHA1 reference hasn't changed from the currently installed version.
+#' @param metadata \code{list} of metadata (normally generated by
+#'   \code{remote_metadata}) to add to the DESCRIPTION file after installation.
```
hadley: I'd say: `Named list of metadata entries to be added to the \code{DESCRIPTION} prior to installation.`

R/session-info.r
```diff
@@ -153,10 +153,27 @@ pkg_source <- function(desc) {
                   desc$GithubRepo, "@",
                   substr(desc$GithubSHA1, 1, 7), ")")
   } else if (!is.null(desc$RemoteType)) {
-    str <- paste0(desc$RemoteType, " (",
-      desc$RemoteUsername, "/",
-      desc$RemoteRepo, "@",
-      substr(desc$RemoteSha, 1, 7), ")")
+    remote_type <- desc$RemoteType
+
+    user_repo <- character(0)
```
hadley: This definition (and `sha` and `user_sha` below) would be better off in an else clause.

R/session-info.r
```diff
@@ -153,10 +153,27 @@ pkg_source <- function(desc) {
...
+      user_repo <- paste0(desc$RemoteUsername, "/", desc$RemoteRepo)
+    }
+
+    sha <- character(0)
+
+    if (!is.null(desc$RemoteSha)) {
+      sha <- paste0("@", substr(desc$RemoteSha, 1, 7))
+    }
+
+    user_sha <- character(0)
```
hadley: This logic is all a bit confusing, and I think it could do with some refactoring.

tests/testthat/test-dll.r
```diff
@@ -12,7 +12,7 @@ test_that("unload() unloads DLLs from packages loaded with library()", {
   on.exit(.libPaths(old_libpaths), add = TRUE)
 
   # Install package
-  install("testDllLoad", quiet = TRUE, args = "--no-multiarch")
+  install("testDllLoad", quiet = TRUE, add_sha = FALSE, args = "--no-multiarch")
```
hadley: I think all these `add_sha = FALSE` need to go away now?

R/install-remote.R
hadley: Oh, @jimhester pointed out that both might exist, so this is the right thing to do.  A comment explaining that would be helpful.

rmflight: Unless I am mistaken, that if else structure you suggested will NOT modify BOTH files if they are both present. This modifies either one if they are present, which is what we want to happen, because different functions will read DESCRIPTION, others read the `rsd` file.

You proposed:
```
if_else_function <- function(e1, e2){
  if (e1) {
    print("e1")
  } else if (e2) {
    print("e2")
  } else {
    print("no e1 or e2")
  }
}
```

which gives:

```
> if_else_function(TRUE, FALSE)
[1] "e1"
> if_else_function(TRUE, TRUE)
[1] "e1"
> if_else_function(FALSE, TRUE)
[1] "e2"
```

I (and @jimhester) think we want

```
if_else_function2 <- function(e1, e2){
  if (e1) {
    print("e1")
  } 
  if (e2) {
    print("e2")
  }
  
  if (!e1 && !e2) {
    print("no e1 or e2")
  }
}
```

which gives

```
> if_else_function2(TRUE, FALSE)
[1] "e1"
> if_else_function2(TRUE, TRUE)
[1] "e1"
[1] "e2"
> if_else_function2(FALSE, TRUE)
[1] "e2"
> if_else_function2(FALSE, FALSE)
[1] "no e1 or e2"
```

R/install.r
rmflight: just to be clear, we are now modifying the DESCRIPTION **after** installation.

tests/testthat/test-dll.r
rmflight: yep, they sure do.

rmflight: gone as of adcb043

R/install.r
jimhester: I am not sure the prior / during / after installation distinction is useful knowledge for the user and will likely just make things confusing. Just `Named list of metadata entries to be added to the \code{DESCRIPTION}.` seems sufficient to me.

R/session-info.r
rmflight: I admit the logic is confusing, it took me a good while to work through it myself. 

is it the setting of variables to `character(0)` and then setting them is the problem? I did that to avoid a bunch of `if ... else` constructs that in my mind clutter up the code, but I can do it that way instead.

R/install-remote.R
rmflight: apologies, you commented again while I was creating this example.

R/session-info.r
rmflight: the logic is bad because we want to delineate between:

```
remoteType (username/repo@commit)
remoteType (username)
remoteType (@commit)
remoteType
```

rmflight: Hopefully a little clearer in cbb090e

R/install-remote.R
rmflight: explanation added in 7ec9b258

hadley:
I don't think the logic is quite right. When I run `install()` with this version (on devtools itself), the installed DESCRIPTION is not modified, and the DESCRIPTION in the current directory is.

hadley:
The problem is that `base::find.package(pkg$package)` doesn't give the installed path for a package loaded with devtools, but instead gives the path to the source package.  This will only be a problem if you've run `load_all()` before `install()`, but that seems like a fairly common scenario to me (and is certainly something I'd do)

rmflight:
Is there an easy way to tell that it was done by load_all so as to avoid
this? Or maybe we can look for the package in the directory specified by
.libPaths()?

jimhester:
@rmflight Best thing to do is just use `inst()` here instead of `find.package()`. It will always give the path to the installed copy of the package.

See https://github.com/jimhester/devtools/commit/fc359c214d81073375e0c0ec0cf85b827f28edb7

rmflight:
Used inst() in 1c3fccd, and added a test of `load_all(); install()` to double check that the correct instance was indeed modified.

hadley:
Ok - this now just needs a bullet in NEWS.md and we're good to merge.

@jjallaire @kevinushey FYI: this PR will add git/github information when installing from a local package. I vaguely remember there's some interaction with packrat, so you need to know about this.

rmflight:
@jimhester: using `withr::with_temp_libpaths()` in the test suite 2f54a5e6

@hadley: bullet added to News.md 11a5f6ba

rmflight:
OK, I'm curious, how do I resolve a conflict that I can't see?

hadley:
Don't worry about it, I just merged by hand.

rmflight:
Thank you! And thank you and @jimhester for all of your support and coaching through this process. Just so you know, this was my first pull request that involved any amount of real code, and you both made it a pretty cool experience.

PR #1030, opened 2016-01-15
package_find_repos works properly with multiple repos provided

### LINKED ISSUE #943: install_version fails because of CRANextra
`devtools::install_version` apparently tries to download already installed packages from the "CRANextra" repo, which can cause the installation to fail for no good reason.

My `getOption('repos')` is

                                CRAN                            CRANextra 
         "https://cran.rstudio.com/" "http://www.stats.ox.ac.uk/pub/RWin" 

and when I try `devtools::install_version('data.table', '1.9.4')`, the very end of the 
installation fails with

    ...
    ** testing if installed package can be loaded
    *** arch - i386
    *** arch - x64
    * DONE (data.table)
    Downloading package from url: http://www.stats.ox.ac.uk/pub/RWin/src/contrib/Archive/data.table/data.table_1.9.4.tar.gz
    Error in download(bundle, x$url, x$config) : 
      client error: (404) Not Found

__even though the package was installed succesfully__ (`library(data.table)` succeeds).

If I remove "CRANextra" from repos, with, e.g., 

    devtools::install_version('data.table', '1.9.4', repos='https://cran.rstudio.com/')
    
the installation succeeds.

jimhester:
Fixes #943 

As far as I can tell only CRAN mirrors have a `Meta/archive.rds` file, the following all 404 (they all have valid `src/contrib/PACKAGES` though).

- http://bioconductor.org/packages/3.2/bioc/src/contrib/Meta/archive.rds
- http://packages.ropensci.org/src/contrib/Meta/archive.rds
- http://www.stats.ox.ac.uk/pub/RWin/src/contrib/Meta/archive.rds

The test fails with the current master, but succeeds with this PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # devtools 1.9.1.9000
 
+* `install_version()` now more robust when handling multiple repos (#943, #1030,
+  @jimhester).
+
 * Be more verbose about which package is installed for revdep check
   (#926, @krlmlr).
 
--- a/R/install-version.r
+++ b/R/install-version.r
@@ -41,7 +41,7 @@ install_version <- function(package, version = NULL, repos = getOption("repos"),
     }
   }
 
-  url <- paste(repos, "/src/contrib/Archive/", package.path, sep = "")
+  url <- paste(info$repo[1L], "/src/contrib/Archive/", package.path, sep = "")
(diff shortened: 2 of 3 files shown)


jimhester:
Ok this should be ready to merge unless there is something you think needs changing.

PR #1041, opened 2016-01-21
Print message if SHA1 has not changed.

### LINKED ISSUE #1040: install_github needs to be more informative
If not reinstalling because sha matches

jimhester:
Message currently works as follows
``` r
install_github("jimhester/covr")
#> Skipping install for github remote, the SHA1 has not changed since last install.
#>   Use `force = TRUE` to force installation
```

Let me know if you want to tweak the copy or if it should only be printed interactively.

Fixes #1040

### DIFF
--- a/R/install-remote.R
+++ b/R/install-remote.R
@@ -67,8 +67,15 @@ different_sha <- function(remote = NULL,
     local_sha <- local_sha(remote_package_name(remote))
   }
 
-  different <- remote_sha != local_sha
-  isTRUE(different) || is.na(different)
+  different <- isTRUE(remote_sha != local_sha)
+  different <- different || is.na(different)
+  if (!different) {
+     message(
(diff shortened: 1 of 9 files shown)


R/install-remote.R
```diff
@@ -67,8 +67,15 @@ different_sha <- function(remote = NULL,
...
   }
 
-  different <- remote_sha != local_sha
-  isTRUE(different) || is.na(different)
+  different <- isTRUE(remote_sha != local_sha)
+  different <- different || is.na(different)
+  if (!different) {
+     message(
+       "Skipping install for ", sub("_remote", "", class(remote)[1L]), " remote,",
+       " the SHA1 has not changed since last install.\n",
```
hadley: Maybe include (some of) the sha in the output? "...the SHA1 (axcvsadf9234) ..."

PR #1051, opened 2016-01-26
Find the package root for dr_github()

### LINKED ISSUE #1050: `dr_github()` fails if run outside of the package base directory.
``` r
setwd("devtools/R")
dr_github()
#> Warning in read.dcf(path): cannot open compressed file './DESCRIPTION',
#> probable reason 'No such file or directory'
#> Error in read.dcf(path): cannot open the connection
```

I ran into this when checking memoise reverse dependencies. Now that #1002 runs `\donttest{}` blocks https://github.com/hadley/devtools/blob/master/R/doctor.R#L71-L73 was run and caused an error.

We should probably switch those to `\dontrun{}`.

jimhester:
Fixes #1050

### DIFF
--- a/R/doctor.R
+++ b/R/doctor.R
@@ -75,6 +75,9 @@ dr_github <- function(path = ".") {
   if (!uses_git(path)) {
     return(doctor("github", "Current path is not a git repository"))
   }
+  if (!uses_github(path)) {
+    return(doctor("github", "Current path is not a GitHub repository"))
+  }
 
   msg <- character()
   r <- git2r::repository(path, discover = TRUE)


hadley:
Am I missing some reason that you're not doing this?

```R
  if (!uses_github(path)) {
    return(doctor("github", "Current path is not a github repository"))
  }
```

jimhester:
Yes because I like making things more complicated than they need to be :frowning:. https://github.com/hadley/devtools/commit/7ef2c95e1b2a9c2abbbd1ff26f62f54f1677b65a does as you suggest!

hadley:
Haha, thanks!

PR #1052, opened 2016-01-27
importFrom withr functions used in load_all

jimhester:
This allows use of load_all() with withr as well, using `withr::` was
causing the withr namespace to be implicitly loaded, which was the cause
of the namespace locking errors.

Fixes jimhester/withr#34

I had to remove the deprecated with_collate, with_envvar functions from devtools in this PR as they conflicted with the ones we are importing from withr. I think the tradeoff is worth it however.


hadley:
I'd rather remove all of the with functions, or none of them, but removing in the next version (which is only a couple of months away seems too soon)

hadley:
Can you import functions with an alias?

jimhester:
I don't think you can import with an alias using `importFrom` in the `NAMESPACE`, but we could do it explicitly at install-time `withr_with_dir <- withr::with_dir` which I think will allow us to keep the deprecated functions and fix this issue.

gaborcsardi:
Of course install-time stuff is always dangerous, because you might end-up with an out-of-date version in devtools. E.g. if a bug is fixed in withr (assuming it has bugs, sorry :), then you need to re-install devtools to fix it. 

IMO it is better to "import" them in `.onLoad()`.

I know that you know all this, just a reminder. :)

jimhester:
Thanks Gabor, you are correct doing it on `.onLoad()` is preferred. https://github.com/jimhester/devtools/commit/4e7e30e3faad543aecfb2014f23dfdaf3c4ff631 should preserve the deprecation behavior and allow `load_all("withr")` to work as well.

jimhester:
My previous commit was inadvertently assigning into the global namespace. https://github.com/hadley/devtools/commit/014f3074d27dcdf695196da637b631ef7e77a5e8 properly assigns the aliased functions into the package namespace.

hadley:
LGTM

PR #1054, opened 2016-01-28
extend use_package

stevenpollack:
- created a 'version' parameter that allows for version comparisons to
  be incorporated into DESCRIPTION. E.g., `version='min'` would introduce
  "package (>= packageVersion)" into the selected type of dependency...
- bumped version number to 1.11.0.0

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: devtools
 Title: Tools to Make Developing R Packages Easier
-Version: 1.10.0.9000
+Version: 1.11.0.0
 Authors@R: c(
     person("Hadley", "Wickham", , "hadley@rstudio.com", role = c("aut", "cre")),
     person("Winston", "Chang", role = "aut"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: devtools
 Title: Tools to Make Developing R Packages Easier
-Version: 1.11.0.0
+Version: 1.10.0.9001
 Authors@R: c(
     person("Hadley", "Wickham", , "hadley@rstudio.com", role = c("aut", "cre")),
     person("Winston", "Chang", role = "aut"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: devtools
 Title: Tools to Make Developing R Packages Easier
-Version: 1.10.0.9001
+Version: 1.10.0.9000
 Authors@R: c(
     person("Hadley", "Wickham", , "hadley@rstudio.com", role = c("aut", "cre")),
     person("Winston", "Chang", role = "aut"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,16 @@
+# devtools 1.10.0.9001
+
+## New features
+
+* `use_package()` has gotten its signature extended to include arguments `version`,
+  and `compare` which allow users to insert more fine-grained dependency 
+  specifications in their DESCRIPTION files (e.g. `utils (>= 3.2.0)`).
+  
+## Bug fixes and minor improvements
+
+* unit tests for `use_package()` have been implemented (these include checks for
+  idempotency).
+
 # devtools 1.10.0.9000
 
 ## New features
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,13 +1,6 @@
-# devtools 1.10.0.9001
-
-## New features
-
 * `use_package()` has gotten its signature extended to include arguments `version`,
   and `compare` which allow users to insert more fine-grained dependency 
   specifications in their DESCRIPTION files (e.g. `utils (>= 3.2.0)`).
-  
-## Bug fixes and minor improvements
-
 * unit tests for `use_package()` have been implemented (these include checks for
   idempotency).
 
--- a/R/infrastructure.R
+++ b/R/infrastructure.R
@@ -301,46 +301,54 @@ use_package_doc <- function(pkg = ".") {
 #'   "Depends", "Suggests", "Enhances", or "LinkingTo" (or unique abbreviation)
 #' @param pkg package description, can be path or package name. See
 #'   \code{\link{as.package}} for more information.
-#' @param version Type of comparison for the version of \code{package}: must be
-#'  one of "minimum", "exact", or "none".
+#' @param version Value indicating whether \code{package}'s version should be included
+#'   in \code{pkg}'s DESCRIPTION. Either \code{NULL}, \code{TRUE}, or a valid version
+#'   string. If \code{version} is \code{TRUE}, \code{package}'s current version
+#'   will be used. See examples.
+#' @param compare The comparator used for \code{package}'s version. All valid
+#'   CRAN comparators (i.e. \code{==}, \code{>=}, \code{<=}, \code{>}, and
+#'   \code{<}) are accepted, but not considered unless \code{version} is
+#'   non-\code{NULL}.
 #' @family infrastructure
 #' @export
 #' @examples
 #' \dontrun{
 #' use_package("ggplot2")
 #' use_package("dplyr", "suggests")
-#' use_package("stringr", "suggests", version="min")
-#'
+#' use_package("devtools", version = TRUE) # "devtools (>= 1.10.0.9000)"
+#' use_package("devtools", version = TRUE, compare = "==") # "devtools (== 1.10.0.9000)"
+#' use_package("devtools", version = "1.10") # "devtools (>= 1.10)"
+#' use_package("devtools", version = "1.10", compare = ">") # "devtools (> 1.10)"
 #' }
-use_package <- function(package, type = "Imports", pkg = ".", version = "none")  {
+use_package <- function(package, type = "Imports", pkg = ".", version = NULL,
+                        compare =  c(">=", ">", "==", "<=", "<")) {
+
   stopifnot(is.character(package), length(package) == 1)
   stopifnot(is.character(type), length(type) == 1)
 
   if (!is_installed(package)) {
     stop(package, " must be installed before you can take a dependency on it",
-      call. = FALSE)
+         call. = FALSE)
   }
 
-  versions <- c("none", 'minimum', 'exact')
-  names(versions) <- versions
-
-  version <- versions[[match.arg(tolower(version), names(versions))]]
-
-  dependency <- switch(version,
-         minimum = paste0(" (>= ", packageVersion(package), ")"),
-         exact = paste0(" (== ", packageVersion(package), ")"),
-         NULL)
+  if (!is.null(version)) {
+    # add a version dependency
+    if (isTRUE(version)) {
+      version <- packageVersion(package)
+    }
+    compare <- match.arg(compare)
+    package_txt <- paste0(package, " (", compare, " ", version, ")")
+  } else {
+    package_txt <- package
+  }
 
   types <- c("Imports", "Depends", "Suggests", "Enhances", "LinkingTo")
   names(types) <- tolower(types)
 
   type <- types[[match.arg(tolower(type), names(types))]]
 
-  # update modification to DESCRIPTION to include version
-  packageAndDependency <- paste0(package, dependency)
-
-  message("Adding ", packageAndDependency, " to ", type)
-  add_desc_package(pkg, type, packageAndDependency)
+  message("Adding ", package_txt, " to ", type)
+  add_desc_package(pkg, type, package_txt)
 
   msg <- switch(type,
     Imports = paste0("Refer to functions with ", package, "::fun()"),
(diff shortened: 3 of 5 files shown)


jimhester:
Specifying exact version dependencies `pkg (== 1.1.1)` is not really useful in practice because `install.packages()` always installs the latest package from the repository. This means that your package will no longer be installable if any of the dependencies with exact specifications release a new version. Because of this I think `use_package()` could just have a `version=` parameter that would add `(>= version)` if set or do nothing if `NULL`.

Please either do not bump the devtools version or bump the minor version number only, e.g. to `1.10.0.9001`.

stevenpollack:
I agree that `==` is not useful in practice, but it doesn't hurt to provide the option, so I figured "why not?"...

I was wondering what the versioning semantics were...

I couldn't find any unit tests for this function; are there none?

stevenpollack:
Also, I'm not in love with the parameter name `version`... Can we come up with something better?

stevenpollack:
@jimhester can I get an update on the status of this PR?

jimhester:
IMHO the interface for this should be as follows.
``` r
use_package <- function(package, type = "Imports", pkg = ".", 
  version = NULL, compare =  c(">=", ">", "==", "<=", "<")) {

  if (!is.null(version)) { # add a version dependency
    if (isTRUE(version)) {
      version <- packageVersion(package)
    }
    compare <- match.arg(compare)
    package_txt <- paste0(package, " (", compare, " ", version, ")")
  } else {
    package_txt <- package
  }
  package_txt

  # ... rest of function
}
use_package("devtools")
#> [1] "devtools"
use_package("devtools", version = TRUE)
#> [1] "devtools (>= 1.10.0.9000)"
use_package("devtools", version = TRUE, compare = "==")
#> [1] "devtools (== 1.10.0.9000)"
use_package("devtools", version = "1.10")
#> [1] "devtools (>= 1.10)"
use_package("devtools", version = "1.10", compare = ">")
#> [1] "devtools (> 1.10)"
```

This allows the default behavior to remain the same as it is currently. If you want to have a version dependency you can pass `version = TRUE` to use the installed version, or specify the explicit version you want. If you want to use a different comparison function you can pass that to `compare`.

stevenpollack:
I really like this! Will implement shortly.

jimhester:
Please add a note about this change to the NEWS.md file.

It also would be nice to have some tests for this function, there are not currently any written. I would make a simple test package like used in the tests for `use_data()` https://github.com/hadley/devtools/blob/master/tests/testthat/test-infrastructure.r. The package you can add could be a base package like `utils`, which would ensure it is always available.

stevenpollack:
@jimhester (first draft) of unit tests have been created -- I'm not sure if the choice of doing everything in an `apply` over an `expand.grid` of all test cases is "best practice", but I figure I'd throw it out there, and hear if you had any better ideas.

stevenpollack:
@jimhester or @hadley -- thoughts?

hadley:
We're both involved in other projects right now but will review it before the next release.

stevenpollack:
Right on, thanks.

NEWS.md
```diff
@@ -1,3 +1,16 @@
+# devtools 1.10.0.9001
+
+## New features
```
hadley: Please don't add headings yourself. I add these during the final release process.

R/infrastructure.R
```diff
@@ -291,6 +291,27 @@ use_package_doc <- function(pkg = ".") {
   writeLines(out, file.path(pkg$path, path))
 }
 
+# help function for use_package
+# takes the same arguments as use_package
+build_package_txt <- function(package, version = NULL,
+                              compare =  c(">=", ">", "==", "<=", "<")) {
```
hadley: I think you should just drop the compare argument as the only one that's useful is `>=`.

R/infrastructure.R
```diff
@@ -301,30 +322,46 @@ use_package_doc <- function(pkg = ".") {
 #'   "Depends", "Suggests", "Enhances", or "LinkingTo" (or unique abbreviation)
 #' @param pkg package description, can be path or package name. See
 #'   \code{\link{as.package}} for more information.
+#' @param version Value indicating whether \code{package}'s version should be included
+#'   in \code{pkg}'s DESCRIPTION. Either \code{NULL}, \code{TRUE}, or a valid version
+#'   string. If \code{version} is \code{TRUE}, \code{package}'s current version
+#'   will be used (see examples). If \code{version} is a string that is not a
```
hadley: I think you can eliminate the sentence describing what happens with invalid version specs

R/infrastructure.R
```diff
@@ -301,30 +322,46 @@ use_package_doc <- function(pkg = ".") {
...
+use_package <- function(package, type = "Imports", pkg = ".", version = NULL,
+                        compare =  c(">=", ">", "==", "<=", "<")) {
+
   stopifnot(is.character(package), length(package) == 1)
   stopifnot(is.character(type), length(type) == 1)
 
   if (!is_installed(package)) {
     stop(package, " must be installed before you can take a dependency on it",
-      call. = FALSE)
+         call. = FALSE)
```
hadley: Looks like you added spurious indentation here

R/infrastructure.R
```diff
@@ -301,30 +322,46 @@ use_package_doc <- function(pkg = ".") {
...
   }
 
+  package_txt <- build_package_txt(package, version, compare)
```
hadley: Should be defined closer to where it's used

R/infrastructure.R
stevenpollack: Like I mentioned in the PR to @jimhester : it costs very little to keep it in. However, I can envision a case where someone is developing a package and wants to use `>=` and `<=` . Or, if they're writing production code that'll turn into a microservice in a docker container, `==` might also make sense.

NEWS.md
stevenpollack: Okie dokie. Is there a "contributing to devtools" page somewhere that could tell people about these little details? E.g., how to add to `NEW.md`, or change the version number?

R/infrastructure.R
stevenpollack: I think that was from RStudio's `Reformat Code` function -- is there a linter I should/could be using for working with this project?

R/infrastructure.R
stevenpollack: this is defined as close as I can make it (one line above `use_package()`'s roxumentation), while still making it accessible to the [unit tests](https://github.com/hadley/devtools/pull/1054/files#diff-67b82d9b2f120c72b1173ce45fcdb8e7R120)

R/infrastructure.R
hadley: I read that, but I disagree. I don't there's any point in exposing an interface that people are unlikely to use. But if you can find a couple of packages in the wild that use `<` or `==`, I might change my mind.

R/infrastructure.R
jimhester: ? why can't you move it to L362?

stevenpollack: because adding it to L362 would make it unaccessible to L120 in the unit test.

R/infrastructure.R
stevenpollack: I can tell you that where I used to work, there was serious consideration for `==` in our private package development since we didn't want to have to update our code when dependencies changed. In the end, we said "what the hell" and went with `>=`. That being said, changes to `stringr` (from 0.6.3 to 1.0.0) ended up "forcing" some time to package maintenance.  

I don't agree with that we should drop `compare` because it's "unlikely" that someone will use the other options. I thought the spirit of `devtools` is to help people build packages (with minimal opinions, beyond best practice). Restricting the user to either `package` or `package (>= packageVersion)` in the `DESCRIPTION` file seems mighty opinionated.

R/infrastructure.R
jimhester: I think Hadley was talking about the `package_txt` variable definition, not the `build_package_txt` function definition.

stevenpollack: Woops, totally misunderstood 😅. Sorry about that.

R/infrastructure.R
stevenpollack: Not to barrage you with updates, however, I ran a script that checked CRAN's available packages for all dependencies that called on comparators (≤, <, >, or ==). There were [10 packages](https://gist.github.com/stevenpollack/0b97208a3e097dbfc5fc#file-packages_with_odd_comparators-tsv), all of which published after 26.1.2015...

The most surprising fact is that popular use of `==`, with 3 counts. Whereas, `≤` and `<` have a total of 8...

stevenpollack:
So... it's been nearly 2 weeks... Any status update?

hadley:
I'll get to it when I get to it.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: devtools
 Title: Tools to Make Developing R Packages Easier
-Version: 1.10.0.9000
+Version: 1.10.0.9001
```
hadley: No need to bump the version

R/infrastructure.R
hadley: Please just stick with `>=` - I'm not interested in supporting the other cases.

tests/testthat/test-infrastructure.r
```diff
@@ -59,3 +59,68 @@ test_that("use_data", {
     expect_equal(local_test_data_item_to_save, 42L)
   })
 })
+
+test_that("use_package", {
```
hadley: This test feels like overkill to me - the function is not that complicated so I don't think it needs an extensive testing harness. I'd rather have a few smaller tests that each tested on specific failure point.

stevenpollack: With the exception of a change to [L75](https://github.com/hadley/devtools/pull/1054/files#diff-67b82d9b2f120c72b1173ce45fcdb8e7R75), I don't really see how this is "overkill" so much as it's just thorough. Every potential use of the function is tested

1. for proper functionality
2. for idempotency 

and finally a check to make sure the docs aren't lying and an error is thrown on an invalid version number is performed.

If this is, indeed, overkill, which tests would you keep?

hadley: I'd do one test for proper functionality, and one test for idempotency.

The problem with the current test is that it is so long that I can't easily understand the goal in a glance. That's a problem because I have to maintain this code, and if `use_package()` needs to change in the future, I need to understand what the tests actually do so that I don't break existing behaviour.

tests/testthat/test-infrastructure.r
```diff
@@ -59,3 +59,96 @@ test_that("use_data", {
     expect_equal(local_test_data_item_to_save, 42L)
   })
 })
+
+test_that("add_desc_package only modifies 'field' section of DESCRIPTION", {
```
stevenpollack: @hadley not sure if you're going to like this, but I broke up the tests and made their descriptions a bit more verbose... Let me know...

tests/testthat/test-infrastructure.r
```diff
@@ -59,3 +59,96 @@ test_that("use_data", {
...
+  # test error is thrown with invalid `version`;
+  # use the tests/testthat/testData/ as test package.
+  expect_error(use_package("utils", "Imports", "testData", "invalid"))
+})
+
+# Set up a grid of the potential inputs to use_package and individual test
+# them for correctness and behaviour. These are only partial tests, since the
+# checks for idempotency, error-throwing, and making sure *only* the indicated
+# fields are affected, are all handled above.
+test_that("use_package can modify Imports, Suggests and Depends", {
```
hadley: I think you can drop this entire test - 30 lines of code seems overkill to test a single line in the function. You're effectively just testing that `match.arg()` does what it's supposed to.

tests/testthat/test-infrastructure.r
hadley: These are much better!

tests/testthat/test-infrastructure.r
```diff
@@ -59,3 +59,96 @@ test_that("use_data", {
...
+  old_desc <- read_dcf(path_to_test_desc)
+
+  # be sure to return the description to original state on exit
+  on.exit(write_dcf(path_to_test_desc, old_desc))
+
+  # add testPackage to the Imports: of the testData package;
+  # do it twice and check that the description file is the same as in
+  # the second time.
+  import_text <- "testPackage (>= 1.0.0)"
+  descriptions <- lapply(c(1,2), function(call_repetition) {
```
hadley: Can you eliminate the `lapply()`? I think copying and pasting is fine if you only have two copies of basically the same code.

tests/testthat/test-infrastructure.r
stevenpollack: Dropped.

tests/testthat/test-infrastructure.r
stevenpollack: 🙇

hadley:
I'm going to close this PR in recognition that it's been open for over year with no movement.

stevenpollack:
To be honest, I had done my part and answered all follow up's from package
maintainers... So, if I may write candidly, this should've been merged in
months ago... Why it has yet to be, is beyond me.

PR #1060, opened 2016-02-01
update hardcoded http://cran.rstudio.com to https

bbolker:


jimhester:
I think we need to bump the R version in DESCRIPTION as well to at least 3.2, possibly 3.2.2. Versions prior to that did not properly support https repositories. See https://support.rstudio.com/hc/en-us/articles/206827897 for more details. 

Maybe devtools should set the proper download method for https for you?

hadley:
Maybe 

```R
cran_url <- function() {
  if (capabilities("libcurl")) {
    "https://cloud.r-project.org"
  } else {
    "http://cloud.r-project.org"
  }
}
```

?

jimhester:
`capabilities("libcurl")` should work for R >= 3.2 on OS X and linux, but maybe not on windows?

Maybe the easiest thing would be to try establishing a https connection and if it fails fallback to http?

```r
cran_url <- memoise::memoise(function() {
  tryCatch({
    stopifnot(identical(httr::status_code(httr::HEAD("https://cloud.r-project.org")), 200L))
    "https://cloud.r-project.org"
    }, error = function(e) "http://cloud.r-project.org")
})
```

jimhester:
The BiocInstaller does a similar thing (https://github.com/Bioconductor/BiocInstaller/blob/master/R/zzz.R#L43-L69)

hadley:
I would really rather keep it simple and just condition of libcurl. Maybe something like:

```R
protocol <- function() {
  if (getRversion() >= "3.3.0" && capabilities("libcurl")) {
    "https"
  } else {
    "http"
  }
}
```

hadley:
@jjallaire pointed out that there are other situations in which an https mirror will still succeed so should test directly with `download.file()`.

Maybe something like:

```R
mirror_ok <- function(url) {
  tryCatch(
    {
      download.file(paste0(url, "/Rlogo.svg"), destfile = tempfile(), quiet = TRUE)
      TRUE
    }, 
    error = function(...) FALSE
  )
}


cran_mirror <- function() {
  if (mirror_ok("https://cran.rstudio.com")) {
    "https://cran.rstudio.com"
  } else if (mirror_ok("http://cran.rstudio.com")) {
    "http://cran.rstudio.com"
  } else {
    stop("Failed to connect to RStudio CRAN mirror", call. = FALSE)
  }
}
```

hadley:
Would still need to be memoised (if it didn't fail)

jjallaire:
I just realized that one problem with this is that an attacker can arrange
for the test to fail thus forcing you back to http. I think you need some
variation of this code to test whether the current download method is
secure:

https://github.com/rstudio/rstudio/blob/master/src/cpp/session/modules/SessionPackages.R#L1045-L1076

hadley:
I'm going to close this and start a new PR

PR #1062, opened 2016-02-02
Use the last archived version if version == NULL

### LINKED ISSUE #1033: install_version fails for archived packages
I am assuming that it is supposed to install the latest available version in this case. 

```r
> install_version("igraph0")
Downloading package from url: http://cran.rstudio.com/src/contrib/Archive/c("http://cran.rstudio.com", "http://cran.rstudio.com", "http://cran.rstudio.com", "http://cran.rstudio.com", "http://cran.rstudio.com", "http://cran.rstudio.com", "http://cran.rstudio.com", "http://cran.rstudio.com")
Error in download(bundle, x$url, x$config) :
  client error: (404) Not Found
```

This is hopefully enough to reproduce it.

jimhester:
If the package is not available we should install only the last archived
version.

Fixes #1033

### DIFF
--- a/R/install-version.r
+++ b/R/install-version.r
@@ -46,28 +46,29 @@ install_version <- function(package, version = NULL, repos = getOption("repos"),
 }
 
 package_find_repo <- function(package, repos) {
+  archive_info <- function(repo) {
+    if (length(repos) > 1)
+      message("Trying ", repo)
+
+    archive <-
+      tryCatch({
+        con <- gzcon(url(sprintf("%s/src/contrib/Meta/archive.rds", repo), "rb"))
+        on.exit(close(con))
+        readRDS(con)
+      },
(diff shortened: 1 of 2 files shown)


R/install-version.r
```diff
@@ -46,25 +46,35 @@ install_version <- function(package, version = NULL, repos = getOption("repos"),
 }
 
 package_find_repo <- function(package, repos) {
-  for (repo in repos) {
-    if (length(repos) > 1)
-      message("Trying ", repo)
+  res <- do.call(rbind.data.frame,
```
hadley: Could we drop some nesting here?

jimhester: 0ec4145 defines a helper function so we don't have so much nesting.

hadley:
Thanks!

Issue #1037, opened 2016-01-20
Implement update_packages


hadley:
Following proposal in #975 

This should be implemented as a new generic `update_remote()` function for remote objects, and will probably also need some code to create the remote object from a DESCRIPTION. This should make it possible to share a lot of code with the existing installation functions.

PR #975, opened 2015-11-09
Augmented update_packages to update those installed from remote repos

bearloga:
- Moved update_packages() from deps.R to update.R
- Make pkgs argument optional in update_packages to allow updating all
installed packages (installed from CRAN or otherwise)
- Added utility functions: get_local_info() and get_remote_sha1() which
are used in update_remotes()

<strike>P.S. It looks like **git2r** is getting a `remote_ls` (https://github.com/ropensci/git2r/pull/172) but that hasn't propagated yet, so I'm using `git ls-remote` at OS level for now.</strike> **Update**: using `git2r::remote_ls()` now

PR #1067, opened 2016-02-03
Update remotes

jimhester:
Implementation of `update_remotes` as defined in #1037, #975.

I also improved the SVN support to include the SVN revision, so this will now work with all remotes in devtools.

I still need to add tests for this, but this works with a simple test on all remote types on my machine.


R/update-remote.R
```diff
@@ -0,0 +1,24 @@
...
+    if (!yesno("Are you sure you want to update all installed packages with remotes?")) {
+      pkgs <- installed.packages()[, "Package"]
+    } else {
+      return(invisible())
+    }
+  }
+
+  remotes <- compact(lapply(pkgs, package2remote))
+
+  if (!isTRUE(force)) {
```
hadley: Is there are reason to do this here, rather than relying on the individual install methods?

jimhester:
https://github.com/jimhester/devtools/commit/ed000dd17f5561464915b28d2d5d3c2a3e0c3caa does as you suggest moving the SHA checking into `install_remote()` so it works the same across all remote types.

I also moved the remote updating into `update.package_deps`, so it will automatically work with the existing `update_packages()`.

R/install-remote.R
```diff
@@ -108,3 +116,46 @@ remote_download <- function(x, quiet = FALSE) UseMethod("remote_download")
 remote_metadata <- function(x, bundle = NULL, source = NULL) UseMethod("remote_metadata")
 remote_package_name <- function(remote, ...) UseMethod("remote_package_name")
 remote_sha <- function(remote, ...) UseMethod("remote_sha")
+
+package2remote <- function(x, ...) {
+  x <- packageDescription(x)
+  if (is.null(x$RemoteType)) {
+    return(NULL)
```
hadley: What do you think about having a CRAN package type here? That might make it easy for `update_packages()` to work with everything

jimhester: This has been done at https://github.com/jimhester/devtools/commit/8bdb10bb92972cc26d02a0d1ef6c711c24bbec, let me know your thoughts...

R/install-CRAN.r
```diff
@@ -0,0 +1,58 @@
...
+#' @param pkgs Character vector of packages to install.
+#' @inheritParams package_deps
+#' @export
+#' @family package installation
+#' @examples
+#' \dontrun{
+#' install_CRAN("ggplot2")
+#' install_CRAN(c("httpuv", "shiny")
+#' }
+install_CRAN <- function(pkgs, repos = getOption("repos"), type = getOption("pkgType"), ..., quiet = FALSE) {
```
hadley: I'd prefer all lowercase to be consistent

R/install-CRAN.r
```diff
@@ -0,0 +1,58 @@
...
+    name = pkg,
+    repos = repos,
+    pkg_type = type)
+}
+
+
+#' @export
+remote_download.CRAN_remote <- function(x, quiet = FALSE) {
+  dest_dir <- tempdir()
+  download.packages(x$name, destdir = dest_dir, repos = x$repos, type = x$pkg_type)[1, 2]
```
hadley: I forget how this works, but are we already computing the dependencies that are missing/out-of-date?

jimhester: Dependencies are all done in `install()` itself which calls `install_deps()` so we don't have to worry about it here.

R/install-CRAN.r
jimhester: Done at https://github.com/jimhester/devtools/commit/16dbe00e4bc70bcd3eb0d2e09e2c221c39860715

krlmlr:
While you're at it: Is there a way to *not* install/upgrade remotes, just to install a single package? I tried

```
devtools::install("path", dependencies = FALSE, upgrade_dependencies = FALSE, metadata = NULL)
```

to no avail. (Now I know that the "metadata" arg does something different, never mind.) To me, it looks like install_dev_remotes() is called unconditionally. For now, I'll be using `RCMD("INSTALL")` as workaround.

jimhester:
@krlmlr You are correct, the remotes should respect `update_dependencies` as well.

jimhester:
https://github.com/hadley/devtools/pull/1067/commits/c35630ca941b043b8560c63801ab7a34ac13507d has a simple implementation to handle #1109

PR #1068, opened 2016-02-04
Fix revdep_check() when git2r is not loaded

krlmlr:
When installing package during revdep check, use new libpath in addition to existing libpath. Reason: if git2r is not loaded yet, an error is raised if it not in the temporary library.

### DIFF
--- a/R/revdep.R
+++ b/R/revdep.R
@@ -118,10 +118,11 @@ revdep_check <- function(pkg = ".", recursive = FALSE, ignore = NULL,
     dir.create(srcpath)
 
   message("Installing ", pkg$package, " ", pkg$version, " from ", pkg$path)
-  withr::with_libpaths(libpath, install(pkg, reload = FALSE, quiet = TRUE))
+  withr::with_libpaths(libpath, action = "prefix",
+                       install(pkg, reload = FALSE, quiet = TRUE))
   on.exit(remove.packages(pkg$package, libpath), add = TRUE)
 
-  res <- withr::with_envvar(c(
+  withr::with_envvar(c(
     NOT_CRAN = "false",
     RGL_USE_NULL = "true"
   ), {
--- a/R/revdep.R
+++ b/R/revdep.R
@@ -119,7 +119,7 @@ revdep_check <- function(pkg = ".", recursive = FALSE, ignore = NULL,
 
   message("Installing ", pkg$package, " ", pkg$version, " from ", pkg$path)
   withr::with_libpaths(libpath, action = "prefix",
-                       install(pkg, reload = FALSE, quiet = TRUE))
+                       install(pkg, reload = FALSE, quiet = TRUE, dependencies = TRUE))
   on.exit(remove.packages(pkg$package, libpath), add = TRUE)
 
   withr::with_envvar(c(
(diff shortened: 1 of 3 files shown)


jimhester:
This LGTM. I ran into this as well. I think this is a regression caused by the switch to withr. `withr::with_libpaths` has a default action of 'replace', while I believe the devtools version prepended.

krlmlr:
There are two more instances of `with_libpaths()`:

```
R/check-cran.r:51:  libpaths_orig <- withr::with_libpaths(libpath, {
R/build.r:72:  withr::with_libpaths(c(temp_lib, .libPaths()), R(cmd, path, quiet = quiet))
```

jimhester:
I was wrong about it being a regression of `set_libpaths`, it did replace completely previously.

That first instance was deliberately changed to from prepending to replace at
https://github.com/hadley/devtools/commit/516e59fbeebf438adb9c44bb239bf7186bff62ab. The second looks like we can replace with `action = 'prefix'`.

krlmlr:
Merged with master, and replaced instance of with_libpaths() by with_temp_libpaths().

R/revdep.R
```diff
@@ -118,10 +118,11 @@ revdep_check <- function(pkg = ".", recursive = FALSE, ignore = NULL,
     dir.create(srcpath)
 
   message("Installing ", pkg$package, " ", pkg$version, " from ", pkg$path)
-  withr::with_libpaths(libpath, install(pkg, reload = FALSE, quiet = TRUE))
+  withr::with_libpaths(libpath, action = "prefix",
+                       install(pkg, reload = FALSE, quiet = TRUE))
```
hadley: I think this maybe needs `dependencies = TRUE`

R/revdep.R
```diff
@@ -135,19 +136,8 @@ revdep_check <- function(pkg = ".", recursive = FALSE, ignore = NULL,
 
     list(check_dir = check_dir, libpath = libpath, pkg = pkg, deps = pkgs)
   })
-
-  rule("Saving check results to `revdep/check.rds`")
```
hadley: I think you accidentally deleted this?

krlmlr:
Restored lines killed when resolving merge conflicts, added "dependencies = TRUE".

hadley:
BTW I think if you always add the NEWs bullet at the top, it's less likely to cause merge conflicts because it gets anchored to the unchanging heading. (That's my theory anyway)

krlmlr:
Nope: I have put together a [test script](https://github.com/krlmlr/merge-news/blob/master/test.sh) that simulates adding a bullet point to the top of the NEWS file, a merge conflict is raised when trying to merge the second branch. Using .gitattributes helps, but only locally, not on GitHub.

In my code, I record the NEWS entries in the bodies of the merge commit messages. Then, just before release, I use

```
git log --first-parent <tag>.. --format=format:"%b" | sed "/^$/d"
```

where `<tag>` is the last released tag. The log command lists precisely the merge commits, the sed removes empty lines.

jimhester:
setting `NEWS.md merge=union` in the projects `.gitattributes` works locally, but GitHub ignores that setting when doing its merge. That is unfortunate :frowning:. It is probably still worth setting it for devtools as it would make merging easier. https://github.com/isaacs/github/issues/487 is the un-official issue for GitHub support for union merges.

PR #1073, opened 2016-02-05
Additions to check_failures to support use on travis

jimhester:
I don't know if this is the best API. I would like to be able to run something on travis like

```shell
Rscript -e "devtools::check_failures(\"$RCHECK_PATH\", print = TRUE, fail = \"warning\")"
```

And have devtools print all the errors, warnings and notes _and_ then fail appropriately (so the travis build fails).

With this proposal I can just pass `fail = "error"` if `warnings_are_errors = false` or `fail = "warning"` if it is true.

But this is a little weird, so suggestions appreciated!

### DIFF
--- a/R/revdep-summarise.R
+++ b/R/revdep-summarise.R
@@ -148,13 +148,16 @@ check_logs <- function(path) {
 #'   call to \code{\link{check}}
 #' @param error,warning,note logical, indicates if errors, warnings and/or
 #'   notes should be returned
+#' @param print print the result before returning.
+#' @param fail Signal an error if the given class of error is (or worse) occurred.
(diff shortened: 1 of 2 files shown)


hadley:
It feels like this might be better as multiple functions:

* One function returns a named list with notes, warnings, and errors
* Another function filters and prints
* Another function throws the error and has defaults so can be called easily from the command line

I think we should probably also print this out at the end of `check()` and use the same function to throw the error.

hadley:
Actually let me take a stab at this one.

hadley:
This isn't quite perfect yet, but should be a better place to build off of.

Issue #1029, opened 2016-01-15
release() creates slightly different source pkg than build()


jread-usgs:
An odd thing, but release() on https://github.com/USGS-R/geoknife writes `datagroup-methods.Rd` differently than `build()`. The difference results in a warning from the pkg created with `release()` that doesn't exist in the on created with `build()`:
```r
* checking for missing documentation entries ... WARNING
Undocumented S4 methods:
 generic '[' and siglist 'datagroup'

**Comment by hadley** [comment]
Hmmm, I'm trying to replicate this locally, but the tests segfault. Also you have a merge conflict in your `.gitgnore` which you should probably fix. Also the vignettes are really really slow, which makes R CMD check painful.

**Comment by jread-usgs** [comment]
Here is what I see from https://github.com/jread-usgs/geoknife/tree/devtools_release (I removed the vignette and most tests from this branch):

```r
devtools::document(roclets=c('rd'))
```
```
Updating geoknife documentation
Loading geoknife
Creating a new generic function for ‘title’ in package ‘geoknife’
Creating a new generic function for ‘start’ in package ‘geoknife’
Creating a new generic function for ‘url’ in package ‘geoknife’
```
```r
devtools::release()
```
```
Warning: Git not synched with remote.
Updating geoknife documentation
Loading geoknife
Creating a new generic function for ‘title’ in package ‘geoknife’
Creating a new generic function for ‘start’ in package ‘geoknife’
Creating a new generic function for ‘url’ in package ‘geoknife’
Writing datagroup-methods.Rd
Setting env vars ------------------------------------------------------------
```
etc

Note that the `datagroup-methods.Rd` is modified in the `release()` but not `document()`. The `release()` version of the Rd causes the following doc warning: 
```
* checking for missing documentation entries ... WARNING
Undocumented S4 methods:
  generic '[' and siglist 'datagroup'
```

am on the following:
```
* using R version 3.2.3 (2015-12-10)
* using platform: x86_64-apple-darwin13.4.0 (64-bit)
* using session charset: UTF-8
* using option ‘--as-cran’
```

hadley:
`release()` just calls `check(pkg, cran = TRUE, check_version = TRUE, manual = TRUE)`, so I'm not sure what's going on here.

PR #1077, opened 2016-02-06
Add args argument to release() and submit()

krlmlr:
so that packages with heavy PDF vignettes can use `args = "--compact-vignettes=both"` .

Related: #1029, CC @jread-usgs

Follow-up to #981.


krlmlr:
Just released tikzDevice with

```
devtools::release(args="--compact-vignettes=both")
```

using this branch. Twice. :-)

AppVeyor fails because of GitHub's rate limiting.

hadley:
Thanks!

PR #1085, opened 2016-02-11
Print more detailed error message if given JSON response

jimhester:
This takes code from `github_reponse()` to parse the GitHub response information on an error so downloading can give a more informative error message.

If the response does not contain an errors or message component the previous behavior is used.

I could not use `httr::stop_for_status()` directly at https://github.com/hadley/devtools/compare/master...jimhester:download_errors?expand=1#diff-475284c83f91b5f64ec0e4afdbd4c955R135 because the call stack would be uninformative with `sys.call(-1)`.

The following generates the following output after spamming it a bunch of times to produce an error.
```
download("test", "https://github.com/jimhester/withr/archive/1.0.1.tar.gz")
No encoding supplied: defaulting to UTF-8.
Error in download("test", "https://github.com/jimhester/withr/archive/1.0.1.tar.gz") :
  Too Many Requests (RFC 6585) (HTTP 429).
```

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -11,16 +11,22 @@ github_response <- function(req) {
   parsed <- jsonlite::fromJSON(text, simplifyVector = FALSE)
 
   if (httr::status_code(req) >= 400) {
-    errors <- vapply(parsed$errors, `[[`, "message", FUN.VALUE = character(1))
-
-    stop(
-      parsed$message, " (", httr::status_code(req), ")\n",
-      paste("* ", errors, collapse = "\n"),
-      call. = FALSE
-    )
+    stop(github_error(req))
   }
 
   parsed
+}
+
+github_error <- function(req) {
+  text <- httr::content(req, as = "text")
+  parsed <- jsonlite::fromJSON(text, simplifyVector = FALSE)
+  errors <- vapply(parsed$errors, `[[`, "message", FUN.VALUE = character(1))
+
+  structure(list(
+      call = sys.call(-1),
+      message = paste0(parsed$message, " (", httr::status_code(req), ")\n",
+        paste("* ", errors, collapse = "\n")),
+      class = c("condition", "error", "github_error")))
 }
 
 github_GET <- function(path, ..., pat = github_pat()) {
--- a/R/github.R
+++ b/R/github.R
@@ -22,11 +22,12 @@ github_error <- function(req) {
   parsed <- jsonlite::fromJSON(text, simplifyVector = FALSE)
   errors <- vapply(parsed$errors, `[[`, "message", FUN.VALUE = character(1))
 
-  structure(list(
+  structure(
+    list(
       call = sys.call(-1),
       message = paste0(parsed$message, " (", httr::status_code(req), ")\n",
-        paste("* ", errors, collapse = "\n")),
-      class = c("condition", "error", "github_error")))
+        paste("* ", errors, collapse = "\n"))
+    ), class = c("condition", "error", "github_error"))
 }
 
 github_GET <- function(path, ..., pat = github_pat()) {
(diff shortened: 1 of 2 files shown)


R/utils.r
```diff
@@ -124,7 +124,19 @@ first_upper <- function(x) {
 
 download <- function(path, url, ...) {
   request <- httr::GET(url, ...)
-  httr::stop_for_status(request)
+
+  if (httr::status_code(request) >= 400) {
+    # Parse the response if it is from GitHub, use stop_for_status otherwise.
+    tryCatch(github_response(request),
```
hadley: Couldn't we put the `stop_for_status()` in the github response? I think that would make this cleaner.

R/utils.r
```diff
@@ -124,7 +124,19 @@ first_upper <- function(x) {
...
+      error = function(e) {
+        if (inherits(e, "github_response")) {
```
hadley: If you did go this route (which I don't think we should), it would be better to write:

```R
tryCatch(
   ...,
   github_response = function(e) stop(e),
   error = function(e) ...
)
```

jimhester: Ah true, I forgot you could do that based on the class of the error.

R/utils.r
jimhester: I guess the issue was what happens when the response is not from GitHub but somewhere else and is not JSON but another format? Then 
```
text <- httr::content(req, as = "text")
parsed <- jsonlite::fromJSON(text, simplifyVector = FALSE)
```
Is going to throw an error, but in that case we don't care about the parsing error, we just want to return the stop_for_status of the response.

jimhester: Maybe a simpler plan would be to write a `download_github()` function and call that instead of `download()` from install_github.

R/utils.r
```diff
@@ -123,7 +123,7 @@ first_upper <- function(x) {
 }
 
 download <- function(path, url, ...) {
-  request <- httr::GET(url, ...)
+  request <- GET(url, ...)
```
hadley: Doesn't that still need `httr`?

jimhester: Yep sorry, copied this code from an old blame before we were using `httr::` and I missed fixing this line.

R/install-github.r
```diff
@@ -287,3 +287,14 @@ remote_sha.github_remote <- function(remote, url = "https://github.com", ...) {
     unname(res[found[1]])
   }, error = function(e) NA)
 }
+
+download_github <- function(path, url, ...) {
+  request <- GET(url, ...)
```
hadley: And this line? ;)

jimhester: Gahhhhh!

Ok I reviewed this whole thing again, 413300a should actually be correct.

R/github.R
```diff
@@ -11,18 +11,24 @@ github_response <- function(req) {
...
   parsed
 }
 
+github_error <- function(req) {
+  text <- httr::content(req, as = "text")
+  parsed <- jsonlite::fromJSON(text, simplifyVector = FALSE)
+  errors <- vapply(parsed$errors, `[[`, "message", FUN.VALUE = character(1))
+
+  structure(list(
+      call = sys.call(-1),
```
hadley: Can you please tweak indenting here? The nesting of `list()` inside `structure()` isn't obvious.

jimhester: Done https://github.com/hadley/devtools/pull/1085/commits/7863c754753e8843d4963f889cac33b465f69b97

R/github.R
```diff
@@ -11,18 +11,25 @@ github_response <- function(req) {
...
 
   if (httr::status_code(req) >= 400) {
-    errors <- vapply(parsed$errors, `[[`, "message", FUN.VALUE = character(1))
-
-    stop(
-      parsed$message, " (", httr::status_code(req), ")\n",
-      paste("* ", errors, collapse = "\n"),
-      call. = FALSE
-    )
+    stop(oithub_error(req))
```
hadley: g?

jimhester: Ugg, fixed in https://github.com/hadley/devtools/pull/1085/commits/d9a7d67c525b0495689832e22e597766d6e11481

PR #1091, opened 2016-02-15
New arg 'manual = FALSE' to check()

### LINKED ISSUE #1087: check() doesn't use --no-manual anymore by default
It also breaks if pdflatex isn't installed. I think this is caused by 3188bbfdea, but I'm not sure how to handle it:

- [ ] Add a new arg "manual = FALSE" to check()?
- [ ] Redo the pdflatex check in check_built()?
- [ ] ?

krlmlr:
Fixes #1087.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -55,9 +55,9 @@
   `R CMD check` finds an `ERROR`), returning an object that summarises
   the check failures.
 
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!

PR #1097, opened 2016-02-19
use relative path rather than full path in .Rbuildignore

kevinushey:


jimhester:
Thanks!

@hadley I am just merging this as it is clearly wrong and a minor change.

hadley:
I thought I'd fixed this but I forgot to push :/

PR #1098, opened 2016-02-22
Use relative path in .Rbuildignore

krlmlr:
...instead of absolute path; fixes a regression that affects use_build_ignore() and all infrastructure functions that call it.


jimhester:
Dupe of #1097

krlmlr:
Missed it.

PR #1103, opened 2016-02-28
load_code loads files from R/unix and R/windows, closes #1102

### LINKED ISSUE #1102: load_all and R/unix and R/windows directories
Seems like `load_all()`, more precisely `load_code()` does not load `.R` files in the `R/unix` and `R/windows` subdirectories. As "Writing R Extensions" says in Section [Package Subdirectories](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Package-subdirectories):

> The R and man subdirectories may contain OS-specific subdirectories named unix or windows.

gaborcsardi:
Closes #1102

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # devtools 1.10.0.9000
 
+* `load_all()` correctly handles `unix` and `windows` subdirectories
+  within `R` (@gaborcsardi, #1102)
+
 * `revdep_check()` outputs a status message giving elapsed and estimated 
   remaining time every 10 packages.
  
--- a/R/load-code.r
+++ b/R/load-code.r
@@ -40,8 +40,10 @@ parse_collate <- function(string) {
 find_code <- function(pkg = ".") {
   path_r <- file.path(pkg$path, "R")
 
-  code_paths <- dir(path_r, "\\.[Rrq]$", full.names = TRUE)
-  r_files <- withr_with_collate("C", sort(code_paths))
+  r_files <- withr_with_collate(
+    "C",
+    tools::list_files_with_type(path_r, "code", full.names = TRUE)
+  )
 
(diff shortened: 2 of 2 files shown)


jimhester:
roxygen2 has this same deficiency, I will send a PR there as well.

jimhester:
Do we need to sort the filenames here, or change the collation order around the call to `list_files_with_type()`?

`list.files()` returns them sorted, but I guess if you are using a local with a different collation order than C it wouldn't be sorted properly?

jimhester:
Setting the locale/collate goes all the way back to the initial devtools checkin https://github.com/gaborcsardi/devtools/commit/fbae60ced0afee0e7c0f8dc3b5b1bb48d303f3dd#diff-170bec08bbfc9e0be8b468b0fb147ec7R21, I am not sure if it is actually required or not in practice though...

gaborcsardi:
I have no idea about the locale setting. I am trying to test, but the docs says:
```
     Sys.setlocale("LC_COLLATE", "C")   # turn off locale-specific sorting,
                                        # usually, but not on all platforms
```
so maybe this is hard to test.

```r
> library(tools)
> devtools:::withr_with_collate("C", list_files_with_type(".", type = "code"))
[1] "./gabor.R" "./gábor.R" "./gbor.R"
> Sys.getlocale()
[1] "en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8"
> list_files_with_type(".", type = "code")
[1] "./gabor.R" "./gábor.R" "./gbor.R"
```

I guess we can just leave the locale, to be on the safe side, I'll update the PR.

gaborcsardi:
How about now? :)

hadley:
Thanks!

PR #1104, opened 2016-03-02
Support compiling using new toolchain

jimhester:
If the R version is compiled with the new toolchain set the BINPREF
environment varaible so the proper toolchain is used for 32 and 64 bit
builds.

### DIFF
--- a/R/rtools.r
+++ b/R/rtools.r
@@ -275,10 +275,10 @@ version_info <- list(
   "3.3" = list(
     version_min = "3.2.0",
     version_max = "3.3.99",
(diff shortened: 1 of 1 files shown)


R/rtools.r
```diff
@@ -267,7 +275,11 @@ version_info <- list(
   "3.3" = list(
     version_min = "3.2.0",
     version_max = "3.3.99",
```
hadley: We should probably also bump this to support 3.4

hadley: Bump

hadley:
Thanks!

PR #1117, opened 2016-03-18
Use github_document format in README.Rmd

jennybc:
I made a related PR to the R Packages book then realized this needed to happen to make it true!

https://github.com/hadley/r-pkgs/pull/345


hadley:
Thanks!

PR #1121, opened 2016-03-21
Add a default GITHUB_PAT if running on a CI service

jimhester:
This will prevent most 403 errors due to rate limiting.

This PAT is from my lintr-bot account (which has access to no repositories) and has no permissions, so all it can do is read publicly available information. So it should be safe to use.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -75,7 +75,7 @@ github_pat <- function(quiet = FALSE) {
                   "0a7f1ed",
                   "c485e443")
     if (!quiet) {
-      message("Using bundled GitHub PAT")
+      message("Using bundled GitHub PAT. Please add your own PAT to the env var `GITHUB_PAT`")
     }
     return(pat)
   }
--- a/R/github.R
+++ b/R/github.R
@@ -61,7 +61,7 @@ github_tag <- function(username, repo, ref = "master") {
 #' @keywords internal
 #' @export
 github_pat <- function(quiet = FALSE) {
-  pat <- Sys.getenv('GITHUB_PAT')
+  pat <- Sys.getenv("GITHUB_PAT")
   if (nzchar(pat)) {
     if (!quiet) {
       message("Using GitHub PAT from envvar GITHUB_PAT")
--- a/R/github.R
+++ b/R/github.R
@@ -62,21 +62,24 @@ github_tag <- function(username, repo, ref = "master") {
 #' @export
 github_pat <- function(quiet = FALSE) {
   pat <- Sys.getenv('GITHUB_PAT')
+  if (nzchar(pat)) {
+    if (!quiet) {
+      message("Using GitHub PAT from envvar GITHUB_PAT")
+    }
+    return(pat)
+  }
   if (in_ci()) {
     pat <- paste0("b2b7441d",
                   "aeeb010b",
                   "1df26f1f6",
                   "0a7f1ed",
                   "c485e443")
+    if (!quiet) {
+      message("Using bundled GitHub PAT")
+    }
+    return(pat)
   }
-  if (identical(pat, "")) return(NULL)
-
-
-  if (!quiet) {
-    message("Using github PAT from envvar GITHUB_PAT")
-  }
-
-  pat
+  return(NULL)
 }
 
 in_ci <- function() {
--- a/R/github.R
+++ b/R/github.R
@@ -64,9 +64,21 @@ github_pat <- function(quiet = FALSE) {
   pat <- Sys.getenv('GITHUB_PAT')
   if (identical(pat, "")) return(NULL)
 
+  if (in_ci()) {
+    pat <- paste0("b2b7441d",
+                  "aeeb010b",
+                  "1df26f1f6",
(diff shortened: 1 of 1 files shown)


R/github.R
```diff
@@ -64,9 +64,21 @@ github_pat <- function(quiet = FALSE) {
   pat <- Sys.getenv('GITHUB_PAT')
   if (identical(pat, "")) return(NULL)
```
krlmlr: Would the hard-coded PAT below be ever used, if $GITHUB_PAT is empty?

jimhester: You are correct, fixed with 199b0c2eab7e67af06586d1c2eda8fe34add2f36. Thank you for catching that I clearly was not thinking it through correctly.

R/github.R
```diff
@@ -62,11 +62,23 @@ github_tag <- function(username, repo, ref = "master") {
 #' @export
 github_pat <- function(quiet = FALSE) {
   pat <- Sys.getenv('GITHUB_PAT')
+  if (in_ci()) {
+    pat <- paste0("b2b7441d",
```
hadley: I don't think the branch is quite right here. It should be after the check for empty env var, and should give a different message.

jimhester: Yeah I guess you are right this function should

1. If GITHUB_PAT is set, use that (say `Using github PAT from envvar GITHUB_PAT`)
2. If GITHUB_PAT is not set and in a CI, use the default PAT (with appropriate message?)
3. Otherwise return NULL

krlmlr: Could the lintr-bot become a bottleneck? 5000 reqs/hour is not much.

hadley: Yes please!

hadley: @krlmlr it could, but it's a better bottleneck than the current unauthenticated approach, and then message could tell people what they need to do to use their own token

jimhester: 5000 is a lot better than 60 though!

jimhester:
https://github.com/hadley/devtools/pull/1121/commits/422f4d3d534eea776973d6c8e0f8aac41fceb3da should use the correct logic. Let me know if you want to change the message text.

R/github.R
```diff
@@ -62,21 +62,24 @@ github_tag <- function(username, repo, ref = "master") {
...
+    return(pat)
+  }
   if (in_ci()) {
     pat <- paste0("b2b7441d",
                   "aeeb010b",
                   "1df26f1f6",
                   "0a7f1ed",
                   "c485e443")
+    if (!quiet) {
+      message("Using bundled GitHub PAT")
```
hadley: Maybe: "Using bundled GitHub PAT. Please add your own PAT to the env var GITHUB_PAT"?

R/github.R
krlmlr: Not sure: 5000 per token vs. 60 per IP-address. Or is it 5000 per token per IP?

Dreaming out loud, we could implement a web service that provides ephemeral read-only GitHub tokens upon request. Each token is linked to a realm, new realms require a timeout (e.g., 1 minute, gradually decreasing) to prevent abuse. CC @gaborcsardi

gaborcsardi: Hmmmm. I am sure if GH would be happy about this. It is just a way going around (well, loosen) their rate limits. Also, I don't think we could implement realms without requiring authentication.

I think a token is fine for now, but maybe I would create a `devtools-robot` user for it.

Btw. what is GH's approach to publishing these tokens? Is that OK?

krlmlr: A "realm" is just a string that identifies clients. New clients have to wait, existing trusted clients don't. But this doesn't really add any security, just a way to track and perhaps block users.

We sample from a list of tokens it if this turns out to become a bottleneck.

I can't find a reference about publishing tokens other than "guard them carefully" ... "like a password".

Why exactly are we using the API instead of direct download e.g. from https://github.com/hadley/devtools/archive/master.zip?

gaborcsardi: > A "realm" is just a string that identifies clients. New clients have to wait, existing trusted clients don't. But this doesn't really add any security, just a way to track and perhaps block users.

Yes, that's why you would need authentication. If it is done properly.

> I can't find a reference about publishing tokens other than "guard them carefully" ... "like a password".

:) Well, imo that means that we should not publish tokens. Even if they can only read public data.

> Why exactly are we using the API instead of direct download

In theory the download link might change. I doubt that it will often, so probably it would make sense to try the current scheme first, and then fall back to the API query. I think this is way better than setting up a token service.

R/github.R
jimhester: Done https://github.com/hadley/devtools/pull/1121/commits/5b01dadfd5dcf8f0b7372d3b659336e37e6f010e

R/github.R
jimhester: Oh you should definitely not publish tokens. In fact GitHub will automatically revoke them if they are plaintext in any commit (which is why I had to split it up).

I think the API is used because GitHub enterprise instances won't necessarily have a archive/master.zip link but most Travis builds are going to be using public GitHub so it is probably worth using the normal link in the common case.

PR #1131, opened 2016-03-25
allow "http://" type Github hosts, not only "https://"

ampu3ro:
My company has an enterprise account with root endpoint `http://github.bdap.enernoc.com/api/v3` for the API. I would imagine some others have hosts like this as well. In fact, Github's [manual](https://developer.github.com/enterprise/2.5/v3/) refers to endpoints as `http(s)://`. 

The minor tweak in the PR defaults to `https:` if not specified in the `host` argument, otherwise allowing something like
```S
install_github("me/myrepo",host="http://github.bdap.enernoc.com/api/v3")
```
to work. Hope it helps!


ampu3ro:
PTAL. Sorry about the unnecessary commenting :/ first time

jimhester:
This PR breaks the tests on linux and windows. I would suggest running the tests locally with `devtools::test()` and figuring out why this change breaks them, then update your code so it no longer does so.

ampu3ro:
Can't say it was easy (rtools is super finicky!), but I think I got all the tests to pass. Not sure what's failing in the AppVeyor tests though...

ampu3ro:
much better solution. thanks

PR #1135, opened 2016-03-27
release() warns if NEWS.Rd exists

### LINKED ISSUE #1133: release() should warn if both inst/NEWS.Rd and NEWS.md exist
Just happened to me; the former takes precedence, which is probably not what you want.

krlmlr:
Fixes #1133.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,12 @@
 * `create_description()` now sets `Encoding: UTF-8`. This helps non-English
   package authors (#1123).
 
+* `release()` warns if both `inst/NEWS.Rd` and `NEWS.md` exist (@krlmlr, #1135).
+
+* `release()` doesn't throw error if Git head is detached (@krlmlr, #1136).
+
+* `use_cran_badge()` now uses HTTPS URL (@krlmlr, #1124).
+
 * `uses_github()` now confirms that you've picked a good title and description
   (#1092).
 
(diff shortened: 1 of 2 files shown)


R/check-devtools.r
```diff
@@ -98,6 +98,13 @@ check_news_md <- function(pkg) {
     "NEWS.md is not ignored",
     "NEWS.md now supported by CRAN and doesn't need to be ignored."
   )
+
+  news_rd_path <- file.path(pkg$path, "inst/NEWS.Rd")
+  check_status(
+    !file.exists(news_rd_path),
+    "NEWS.Rd removed",
```
hadley: "NEWS.Rd does not exist" would be a better message, I think.

Also this is going to be a false positive for people who use `inst/NEWS.Rd` only

krlmlr:
Changed error message, added pending NEWS. This check is invoked only if NEWS.md exists.

NEWS.md
```diff
@@ -1,5 +1,11 @@
 # devtools 1.10.0.9000
 
+* `release()` warns if both `inst/NEWS.Rd` and `NEWS.md` exist (@krlmlr, #1135).
+
+* `release()` doesn't throw error if Git head is detached (@krlmlr, #1136).
```
hadley: Where did these come from? Can I ignore?

krlmlr: These are good I think -- already merged (#1124, #1136) but didn't have NEWS.

hadley:
Thanks!

Issue #1128, opened 2016-03-24
Function to check that head is up to date from remote


hadley:
From @jimhester, should be able to do with `fetch_heads(repo)`. Replace existing function that checks remote is up to date with master.

jennybc:
For a github remote specifically, this is now a simple API call: https://developer.github.com/changes/2016-02-24-commit-reference-sha-api/.

krlmlr:
To me, this looks like git_sync_status() with the fetch() call re-enabled. Am I missing anything? What's the advantage of asking fetch_heads()?

hadley:
@krlmlr the goal is to add a warning to `release()` if there are commits on the master remote that you don't have locally. What's the best way to figure that out?

krlmlr:
I think git_sync_status() is a good template. ahead_behind() does the checking, local branch is in sync if ahead = behind = 0, if behind > 0 the remote has new commits, if ahead > 0 local has new commits. The fetch() call synchronizes with the remote.

For the release() check, you'd probably:

1. try to fetch(), if this fails (offline?) a warning is issued
2. if fetch() succeeds, check ahead_behind()
3. if ahead *or* behind, warn

hadley:
Ah the problem seems to be that git2r doesn't support the git protocol for updating: "Error in 'git2r_remote_fetch': Unsupported URL protocol".  That makes it not very useful for me.

krlmlr:
git: works for me on Linux, but not https: . If errors in fetch() raised a warning only, this would still be useful.

krlmlr:
Is there a reason why the [Git-specific checks](https://github.com/hadley/devtools/blob/5eb095b314238f002118b24aafc1d0b42c378bb5/R/release.r#L57-L67) are not part of the devtools checks in release_checks()? Anyway perhaps it's better to refactor to a new `git_checks()`.

hadley:
Because they're checks that your git state is good, not for problems with the package

krlmlr:
I'll work on git_checks() then.

krlmlr:
Right: The git@github.com remote protocol (for r/w access) seems unsupported, too.

krlmlr:
@hadley: I got `git@github.com` remotes working with git2r by installing ssh2 libraries on my system. Could you check the full output of

```
install.packages("git2r")
```

to see if git2r is configured with SSH support?

krlmlr:
[...]

PR #1136, opened 2016-03-27
release() with detached Git head doesn't throw error

krlmlr:
@stewid: Can this be done better? I have found is_commit(), but not is_branch().

### DIFF
--- a/R/git.R
+++ b/R/git.R
@@ -29,7 +29,7 @@ git_sync_status <- function(path = ".") {
     return(invisible(FALSE))
   }
 
-  c1 <- git2r::commits(r)[[1]]
+  c1 <- git2r::lookup(r, git2r::branch_target(r_head))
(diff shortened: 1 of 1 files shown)


stewid:
@krlmlr  Looks fine :+1: . I have added the `is_branch` method to git2r https://github.com/ropensci/git2r/commit/a5e65c3d52337f910dbdfbb22c7b54cf418473b0

I suggest that you add `n = 1` in `c1 <- git2r::commits(r, n = 1)[[1]]`
https://github.com/krlmlr/devtools/blob/feature/git-detached/R/git.R#L32 to avoid listing all commits

krlmlr:
Thanks. The code is now using `branch_target()`, looks like this is the most accurate solution (equivalent to `commits(r, n = 1, time = FALSE, topological = FALSE)`).

hadley:
Thanks!  Do you have any thoughts on #1128?

PR #1138, opened 2016-04-02
Remove NOTE about improper registration

jimhester:
The comment mostly explains the reasoning and why this works. Basically checkFF calls `try(eval(e, pkg_env))` where `e` is the first argument to `.C()` and `pkg_env` is the package namespace. `dll$foo` was previously not found in the package namespace which is why the NOTE was happening. This is really a false positive so we should be able to turn it off with `dontCheck()`, but CRAN disables that when you set `--as-cran`. So the only way to remove this NOTE is to define an object in the package namespace, which is what this does.

### DIFF
--- a/R/has-devel.r
+++ b/R/has-devel.r
@@ -1,3 +1,15 @@
+# The checking code looks for the objects in the package namespace, so defining
+# dll here removes the following NOTE
+# Registration problem:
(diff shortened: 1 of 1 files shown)


hadley:
Nice!

PR #1140, opened 2016-04-04
Fully qualify attach calls to remove NOTE

jimhester:
The code looking for attach looks only for `attach` as the first element in a call. Using `base::attach()` invalidates this so no NOTE is returned.

With #1138 devtools is now NOTE free, which should make future submissions a little smoother for us and CRAN.


hadley:
Crafty!

Issue #1101, opened 2016-02-27
add host argument to use_github()?


ijlyttle:
Hi,

The use-case for this is for people who use enterprise GitHub.

I have looked at the code, and I think this can be done, but it would involve (I think) also adding a host argument to:

- `github_GET()`
- `github_POST()`
- `use_github_links()`

The change to `use_github_links()` could be a bit hairy as it might have to map `api.github.com` to `https://github.com` and `github.foo.com/api/v3` to `https://github.foo.com`. 

I am happy to do a PR on this, if you think it is worthwhile to go forward. If so, a couple quick questions:

1. Are there other functions I would need to touch?

2. I also see a problem in testing the code, as it would require a public enterprise version of github to test against. Would it be enough to do private, informal testing, and also (of course) ensure that any changes do not break the existing behavior?

hadley:
I'd be happy to review a PR of this. I don't think you need to worry too much about testing - as long as it works for public github, and the url is pluggable, it should work for any internal github.

ijlyttle:
Will do - thanks!

ijlyttle:
done in #1215

PR #1142, opened 2016-04-08
Adds host argument to `use_github()`

ijlyttle:
Addresses #1101

This is to let you use this function with an enterprise installation of GitHub.

The goal is to have a `host` argument to `use_github()` work in the same way as the host argument to `install_github()`

Here are my claims:

- all the tests: `test(filter = "github")` remain successful
- `use_github()` remains working for `host = "api.github.com"` 
- `use_github()` works for `host = "github.hostname.com/api/v3"`

I tested the last two manually - a repository was created at github (public and private), was pushed, and the right thing appears in DESCRIPTION.

Here's my caveat: in a few places, I went through some contortions to get the right mappings of url's and paths. My goal here was to be clear about the changes I was making - to favor ugly code on the surface to making changes to deeper implementations. In other words, to proscribe what I wanted to happen, but not to try to imagine (yet) the optimal way of making it happen.

I remain at your service if you think that anything here should be done differently.


hadley:
Ooops, closed the wrong issue

R/git.R
```diff
@@ -112,10 +112,10 @@ github_remote_parse <- function(x) {
 
   if (grepl("^https", x)) {
     # https://github.com/hadley/devtools.git
-    re <- "github.com/(.*?)/(.*)\\.git"
+    re <- "https://[^/]*github[^/]*/(.*?)/(.*)\\.git"
```
hadley: I think you can just omit the protocol, simply changing the bit after `github`

R/git.R
```diff
@@ -112,10 +112,10 @@ github_remote_parse <- function(x) {
...
   } else if (grepl("^git", x)) {
     # git@github.com:hadley/devtools.git
-    re <- "github.com:(.*?)/(.*)\\.git"
+    re <- "git@[^:]*github[^:]*:(.*?)/(.*)\\.git"
```
hadley: Same here

R/github.R
```diff
@@ -23,16 +23,51 @@ github_response <- function(req) {
   parsed
 }
 
-github_GET <- function(path, ..., pat = github_pat()) {
+github_GET <- function(path, ..., pat = github_pat(),
+                       host = "api.github.com") {
```
hadley: You'll need to update this for our recent tweaks which include the protocol in the host string

R/github.R
```diff
@@ -23,16 +23,51 @@ github_response <- function(req) {
...
+
+  url <- paste0("https://", host, "/")
+
+  # if the host contains a path, i.e. "github.hostname.com/api/v3",
```
hadley: Why would it contain a path?

R/infrastructure-git.R
```diff
@@ -59,6 +59,8 @@ use_git <- function(message = "Initial commit", pkg = ".") {
 #' @param auth_token Provide a personal access token (PAT) from
 #'   \url{https://github.com/settings/tokens}. Defaults to the \code{GITHUB_PAT}
 #'   environment variable.
+#' @param host GitHub API host to use. Override with your GitHub enterprise
+#'   hostname, for example, "github.hostname.com/api/v3".
```
hadley: Why do you give the full path here? I think it should just be the protocol and host name

hadley:
I think this looks good overall, but the main change is that you should use the host consistently everywhere - that means it should always look like `http://api.github.com`, including the protocol, but not including the path. I think that will simplify the code

R/git.R
ijlyttle: done in f0b9cc9

R/git.R
ijlyttle: done in f0b9cc9

R/github.R
ijlyttle: I do this because the endpoint prefix for enterprise Github is different from github.com.

enterprise: `http(s)://hostname/api/v3/` [ref](https://developer.github.com/v3/enterprise/)
github.com: `https://api.github.com` [ref](https://developer.github.com/v3/)

Of course, this does not mean that stuffing everything into `host` is the right way to do it :)

Just to spitball on a possible paths forward, I could use two args (with defaults): `host = "https://api.github.com"` and `path_prefix = NULL`. To use enterprise github, one could specify 
`host = "https://github.hostname.com"` and `path_prefix = "api/v3"`.

Please let me know what you think and I'll change things accordingly.

R/infrastructure-git.R
ijlyttle: Please see previous response

R/github.R
hadley: Why not just always include the path in the host name? Nothing would change for public github, and you'd just need to document that internal github needs the path added.

ijlyttle: I will try that out - thanks!

ijlyttle:
Will investigate build failures and try again...

ijlyttle:
I reworked things a bit, to put all the url logic into `use_github()`, along with some comments that hopefully lay out the motivation. 

In short, the host argument to `use_github()` must contain a protocol and a hostname, and may include a path to describe the endpoint-root - per your last suggestion (which I can't seem to find on this page anymore).

My claims:

- by manual test, `use_github()` remains working for host = "https://api.github.com"
- by manual test, `use_github()` works for host = "https://github.hostname.com/api/v3"
- all the tests: test(filter = "github") remain successful, except for a series of warnings that appear when messages are expected (output below). This seems unrelated to this PR; I can file a separate issue if you like.

Happy to amend as needed.

```
> test(filter = "github")
Loading devtools
Testing devtools
git usage and GitHub connections: .W.....W..........W.W........
GitHub: ........................

Warnings -------------------------------------------------------------------------------------
1. git (non-)usage is detected, diagnosed, and can be added (@test-github-connections.R#30) - DR_GITHUB FOUND PROBLEMS

2. GitHub non-usage is handled (@test-github-connections.R#47) - DR_GITHUB FOUND PROBLEMS

3. github info and links can be queried and manipulated (@test-github-connections.R#94) - DR_GITHUB FOUND PROBLEMS

4. github info and links can be queried and manipulated (@test-github-connections.R#95) - DR_GITHUB FOUND PROBLEMS

DONE =========================================================================================
```

jennybc:
Those tests look familiar to me ... @ijlyttle I sent you a PR that clears these warnings, if you wish to just fix as part of this.

ijlyttle:
Hi @jennybc - Thanks! I'm glad it was a straightforward fix.

I'll wait for the signal from @hadley to see how he would like to weave the threads together.

hadley:
Can you please merge/rebase? Something is currently wrong with this PR as it's showing changes to >100 files.

It would be great if you want to incorporate @jennybc's changes into this PR

ijlyttle:
Will do - I must have got something backwards, sorry for the mess!

ijlyttle:
I have been twiddling around with the GitHub site - things should appear much cleaner if I just make a new pull request - accordingly:

- I close this PR, with the expectation that a new one will appear very soon.
- The new PR will have incorporated the PR from @jennybc

PR #1147, opened 2016-04-14
CONTRIBUTING.md typo

seankross:
hand -> handy

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,7 +91,7 @@ A pull request is a process, and unless you're a practiced contributor it's unli
     otherwise we don't get any notification that your pull request is
     ready for review.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -91,7 +91,7 @@ A pull request is a process, and unless you're a practiced contributor it's unli
     otherwise we don't get any notification that your pull request is
     ready for review.
 
-Don't worry if your pull request isn't perfect. It's a learning process and we'll be hand to help you out. 
+Don't worry if your pull request isn't perfect. It's a learning process and we'll be handy to help you out.
```
jennybc: or even "happy"?

rmflight: From the context, I think it was probably meant to be `on hand`, but I agree with @jennybc that `happy` or some other positive emotion. `happy` is probably the simplest.

PR #1149, opened 2016-04-14
added return(invisible()) in release

seankross:
Selecting a "no" option at the first few steps returns `NULL` which is printed to the console, unlike in later steps where `release()` returns invisibly. This PR fixes the issue.


hadley:
Thanks!

PR #1151, opened 2016-04-15
More informative message when forcing unload

wch:
A lot of people see this message when installing a package from Github and think there's a bug in the package being loaded and/or don't know what to do about it. This makes the message more informative and tells the user what to do if it starts acting up: restart R.

```
unloadNamespace("shiny") not successful. Forcing unload.
```


hadley:
Thanks!

PR #1155, opened 2016-04-18
Bugfix for gcc-493

jimhester:
I ended up special casing the situation on win-builder as it seems like a reasonable layout to support. The old toolchain installs to `Rtools/gcc-4.6.3` so installing the new on to `Rtools/gcc-4.9.3` does make some sense even if it is not the default installation location.

This also includes some cleanup, - https://github.com/hadley/devtools/commit/1b0fc663b4fe19019f318b8be18bfbfa691acde3 suppresses the output of `create()` used in the infrastucture tests so it won't clutter the output.
- 1023cb9 uses a mock response for the install_version tests as they were brittle on appveyor also turns on caching and uses http CRAN URLs for testing on old release.

I am mainly opening this as a PR to make sure it is working before merging.


jimhester:
https://github.com/hadley/devtools/pull/1155/commits/678dc1e225c9df2be8e8201738a1a7dfa4feb07e should be ok to merge.

It got a clean check from win-builder (http://win-builder.r-project.org/ZrVV35IPB27C/00check.log)

R/rtools.r
```diff
@@ -131,21 +133,53 @@ setup_rtools <- function(cache = TRUE, debug = FALSE) {
   TRUE
 }
 
-scan_path_for_rtools <- function(debug = FALSE) {
+scan_path_for_rtools <- function(
+  debug = FALSE,
```
hadley: Can you indent these lines please?

hadley:
LGTM. Can you add a bullet to NEWS? Let me know once you merge, and I'll release

R/rtools.r
jimhester: How much indentation? Just one additional level to line up with `function`?

hadley: I like to have arguments line up with the `(` in `function(`

jimhester: Great that is what I did :smile:!

jimhester:
Ok now merged, can be released whenever you are ready.

PR #1156, opened 2016-04-19
Modify logic for BINPREF detection/assignment

jimhester:
Uwe requested that we do not look for the default paths used on winbuilder as he may change them in the future. We can't test BINPREF directly, but we _can_ query `R CMD config CC` and see if that points to a valid compiler, if so there is no need for devtools to do anything with BINPREF.

This also fixes test files that were missing because of my global `.gitignore` and adds a `system_output()` function to run a command a capture the output to a character vector.

I am still waiting for a clean build on win-builder, will ping here when I get the results.


jimhester:
http://win-builder.r-project.org/AR9c4VAKB8y6/00check.log report is clean.

Issue #1159, opened 2016-04-21
clean_dll should not delete .def files


jeroenooms:
Def files are often part of source code, for example in the mongolite package. They should not be deleted: https://github.com/hadley/devtools/blob/b6a23be508043fd0ffc24316034b5e45bfc1a941/R/compile-dll.r#L54

PR #1161, opened 2016-04-27
Only clean def files if they are 'packageName.def'

jimhester:
The clean patterns are originally from R CMD build (https://github.com/wch/r-source/blob/4ba41332b405f29569dad311f36500391fe6f41f/src/library/tools/R/build.R#L406-L433), which @hadley referenced in the original PR (https://github.com/hadley/devtools/pull/137#discussion_r1364386)

The previous behavior was too loose, only def files with the package name should be cleaned.

The same is true for `.a` and `.dll` as well, however @wch thought there may be cases we do need to clean all of them.

Fixes https://github.com/hadley/devtools/issues/1159
cc @jeroenooms

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,6 +7,8 @@
 
 * `install_*` functions and `update_packages()` refactored to allow updating of
   packages installed using any of the install methods. (@jimhester, #1067)
+* `clean_dll()` Only removes package_name.def files and now operates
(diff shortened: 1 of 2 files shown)


wch:
Now that I've thought about it some more, I think it's probably better to remove only the `pkgname.a` and `pkgname.dll` files instead of `*.a` and `*.dll`.

Issue #1157, opened 2016-04-19
Bring install_bitbucket interface and backend up to parity with install_github


imanuelcostigan:
GitHub interface

``` R
install_github(repo, username = NULL, ref = "master", subdir = NULL,
  auth_token = github_pat(quiet), host = "api.github.com", force = FALSE,
  quiet = FALSE, ...)
```

Bitbucket interface:

``` R
install_bitbucket(repo, username, ref = "master", subdir = NULL,
  auth_user = NULL, password = NULL, force = FALSE, quiet = FALSE, ...)
```

Particular areas of focus:
- be able to use API token through `auth_token`.
- specify host using `host`. Note: need to check whether Bitbucket and Bitbucket Server REST APIs are consistent

I'll submit a pull request for this

imanuelcostigan:
Notes:

* Current code base can install private and public repos.
* Cannot download zip/tarball of repo through API (v1 or v2). Need to guess naming convention from Downloads section of repo website.

Tasks:

* [ ] Work out Bitbucket vs Bitbucket Server API differences 
* [ ] Check how to specify port number for Bitbucket Server 
* [ ] Confirm that Bitbucket **doesn't** have Github-like PAT functionality which replaces need for username/password pair. Can we read these values in by default from an option / env var? Using option might be better as this could be coupled to `host` specification in an option (see below).
* [ ]  Work out whether `host` parameter approach used for `install_github` makes sense. It's a bit difficult to remember the host name for an enterprise instance of Github or Bitbucket. Should this be stored in an R option / environment variable?

imanuelcostigan:
## OAuth2.0

User setup:

1. Create an OAuth consumer key from `https://bitbucket.org/account/user/{USERNAME}/api`
2. Create two environment variables: `BITBUCKET_CONSUMER_KEY` and `BITBUCKET_CONSUMER_SECRET` that store resulting key/secret pair. See `vignette("api-packages", "http")` appendix for steps to set up across all systems.
3. Call `bitbucket_pat()` which does the OAuth handshake and retrieves an access token and caches to `~/` (does this fall afoul of CRAN rules?). See https://github.com/imanuelcostigan/devtools/commit/d29084e2e36228123d87a72be5864bbec9678092

imanuelcostigan:
## Bitbucket Server

Will create separate set of functions supporting `install_bitbucket_server()` as bitbucket.org and Bitbucket Server are different products with different APIs.

imanuelcostigan:
[...]

PR #1166, opened 2016-04-30
Improve install_bitbucket interface and backend 

imanuelcostigan:
Key user visible changes:

* Bitbucket authenticates over the OAuth protocol rather than using a username 
  and password combination. Access token thus obtained is cached and refreshed
  automatically.
* Bitbucket `refs` are now resolved in `install_bitbucket()` in a manner 
  more consistent with `install_github()`.

Other changes of note:

* Bitbucket `remote_package_name()` and `remote_sha()` methods no longer call
  github_remote class methods, but are now specifically defined for Bitbucket.
* New Bitbucket repo name parser. This is basically a copy of the github parser, but does not support `version` references as Bitbucket does not have this functionality
* Added Bitbucket test cases.

Keen to hear feedback on one particular aspect of the first (and most significant) change: namely, I store the OAuth negotiated access token in `~/` so that R sessions started in **any** directory can make use of this, rather than just sessions opened in the directory in which the token is initially created. I do print a message to the console with words to this effect, but it does appear to fall afoul of [CRAN submission policies](https://cran.r-project.org/web/packages/policies.html): 

> Packages should not write in the users’ home filespace, nor anywhere else on the file system apart from the R session’s temporary directory (or during installation in the location pointed to by TMPDIR: and such usage should be cleaned up). Installing into the system’s R installation (e.g., scripts to its bin directory) is not allowed.


imanuelcostigan:
See #1157

imanuelcostigan:
I am not sure what is causing the following error on `check()`:

```
E  checking tests
   Running the tests in ‘tests/test-that.R’ failed.
   Last 13 lines of output:
        })
     4: FUN(X[[i]], ...)
     5: stop("Function ", name, " not found in environment ", environmentName(env), ".", 
            call. = FALSE)
     
     Updating collate directive in  /private/var/folders/9f/kz67r0rn3qb9tx_rmgbps94r0000gq/T/RtmpDcLHv0/testCollateOrder/DESCRIPTION 
     
     testthat results ================================================================
     OK: 364 SKIPPED: 7 FAILED: 1
     1. Error: Bitbucket remote build works (@test-bitbucket.R#33) 
     
     Error: testthat unit tests failed
     Execution halted
```

When I run `test(filter = "bit")`, tests all run correctly.

hadley:
@jimhester can you review please?

R/install-bitbucket.r
```diff
@@ -56,20 +80,10 @@ remote_download.bitbucket_remote <- function(x, quiet = FALSE) {
...
-  if (!is.null(x$password)) {
-    auth <- httr::authenticate(
-      user = x$auth_user %||% x$username,
-      password = x$password,
-      type = "basic")
-  } else {
-    auth <- NULL
-  }
-
+  auth <- x$auth_token %||% NULL
```
jimhester: There is no point in having `%||% NULL`, the right hand side will only be used if the left hand side is already null, so you should just pass x$auth_token directly to download.

R/bitbucket.R
```diff
@@ -0,0 +1,138 @@
...
+    "pullrequests", x)
+  response <- bitbucket_GET(path, ..., host = params$host,
+    api_version = api_version)
+
+  params$username <- response$author$username
+  params$ref <- response$source$branch$name
+  params
+}
+
+bitbucket_host <- function (host = NULL) {
```
jimhester: This seems an unnecessary complication, why can't you just have `https://api.bitbucket.org` be the default host?

R/bitbucket.R
```diff
@@ -0,0 +1,138 @@
...
+#'
+#' Use as \code{ref} parameter to \code{\link{install_bitbucket}}.
+#' Allows installing a specific pull request.
+#'
+#' @param pull The pull request to install
+#' @seealso \code{\link{install_bitbucket}}
+#' @export
+bitbucket_pull <- function(pull) structure(pull, class = "bitbucket_pull")
+
+bitbucket_resolve_ref <- function(x, params) UseMethod("bitbucket_resolve_ref")
```
jimhester: Rather than creating a new generic maybe we should change the existing `github_resolve_ref` generic to `resolve_ref` and have methods for github and bitbucket pull and releases appropriately.

R/bitbucket.R
```diff
@@ -0,0 +1,138 @@
...
+  repo_rx <- "([^/@#]+)"
+  subdir_rx <- "(?:/([^@#]*[^@#/]))?"
+  ref_rx <- "(?:@([^*].*))"
+  pull_rx <- "(?:#([0-9]+))"
+  ref_or_pull_rx <- sprintf("(?:%s|%s)?", ref_rx, pull_rx)
+  bitbucket_rx <- sprintf("^(?:%s%s%s%s|(.*))$",
+    username_rx, repo_rx, subdir_rx, ref_or_pull_rx)
+
+  param_names <- c("username", "repo", "subdir", "ref", "pull", "invalid")
+  replace <- stats::setNames(sprintf("\\%d", seq_along(param_names)), param_names)
```
jimhester: I think rather than duplicating this logic it would be better to have a function which takes a path, regex and names and returns the result that can be used both here and in `parse_git_repo()`.

```r

parse_repo <- function(path, rx, nms) {
   replace <- stats::setNames(sprintf("\\%d", seq_along(nms)), nms)
  params <- lapply(replace, function(r) gsub(rx, r, path, perl = TRUE))
  if (params$invalid != "")
    stop(sprintf("Invalid repo: %s", path))
  params <- params[sapply(params, nchar) > 0]

  params
}

jimhester:
I would prefer to keep as much implementation consistent with the GitHub methods as possible. If there are methods which are unchanged from the github implementation they should simply be aliases. Bitbucket is not currently widely used compared to GitHub by the main devtools developers so if the code is simply copied they will likely not be kept up to date.

As noted storing the access token in the users' home directory is against CRAN policies, I think another location will have to be considered. There also needs to be a way for users to customize where this token is placed, either a R option or environment variable (or both).

Rather than merging master into your branch could you rebase your branch on master, it will remove the unnecessary merge commits (you will have to `git push --force-with-lease` after doing this).

imanuelcostigan:
Hot off the press! [Bitbucket just released app passwords (aka PAT)](https://blog.bitbucket.org/2016/06/06/app-passwords-bitbucket-cloud/?utm_source=twitter&utm_medium=social&utm_campaign=bitbucket_app_passwords). This is for those with and without 2FA turned on (which the blog post isn't clear about), but the [docs](https://confluence.atlassian.com/bitbucket/app-passwords-828781300.html) are.

I'll look into this in more depth on the weekend. But it's looking likely that I will close this PR and submit a new one. I'll refer back to the feedback provided above in any new PR.

jimhester:
@imanuelcostigan sounds good, I agree a PAT solution is probably easier to implement and use. Thanks for working on this!

hadley:
@imanuelcostigan Do you think you'll be able to do this in the next week? If yes, we can get it into the next version of devtools; if no, it'll have to wait

R/bitbucket.R
imanuelcostigan: Agreed

R/bitbucket.R
imanuelcostigan: that makes sense. Should I include this change in my PR?

jimhester: Yeah please do

imanuelcostigan:
Replaced by #1220

Issue #1125, opened 2016-03-23
should load_all() source test helpers?


jennybc:
Moving over here as directed. Related to https://github.com/hadley/testthat/issues/415.

From off-github discussion:

> I think load_all() would basically call source_test_helpers() on your
> behalf and put them in an environment that's attached to the search
> path.  I think with that + test_path() you'd be able to load_all() and
> then run individual lines of tests with no difficulty

PR #1169, opened 2016-05-02
Option to turn off helper loading

pitakakariki:
I'd prefer not to load my testthat helpers with `load_all`.

I set a bunch of options like default number of simulations and progress bars which need to be different between interactive use and testing.

nb: Relevant issue is https://github.com/hadley/devtools/issues/1125

### DIFF
--- a/R/load.r
+++ b/R/load.r
@@ -66,6 +66,7 @@
 #' @param export_all If \code{TRUE} (the default), export all objects.
 #'   If \code{FALSE}, export only the objects that are listed as exports
 #'   in the NAMESPACE file.
+#' @param helpers load \pkg{testthat} test helpers. 
 #' @param quiet if \code{TRUE} suppresses output from this function.
 #' @inheritParams as.package
 #' @keywords programming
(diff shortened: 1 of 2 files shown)


hadley:
I'd rather not add an argument for this, but instead provide some other way to scope your helpers so they work differently between formal and informal testing.

R/load.r
```diff
@@ -86,7 +87,7 @@
 #' }
 #' @export
 load_all <- function(pkg = ".", reset = TRUE, recompile = FALSE,
-  export_all = TRUE, quiet = FALSE, create = NA) {
+  export_all = TRUE, helpers=TRUE, quiet = FALSE, create = NA) {
```
hadley: Can you please put spaces around `=`?

krlmlr:
#1256 (now merged) relies on the helpers being loaded already by load_all(). We'll need to use `helpers = TRUE` there, too, or use `helpers = FALSE` and revert #1256.

krlmlr:
On that note, I think it would be great if `helpers = TRUE` added `testthat` to the search path, so that the testthat functions are available. @hadley: What do you think?

R/load.r
pitakakariki: Spaces are there now.

hadley:
@pitakakariki any interest in merging and updating? We'll be working on devtools for the next few weeks.

pitakakariki:
@hadley will do. Should I wait till devtools is passing on Appveyor or is it okay to do it now?

pitakakariki:
I've sent a pull request to pkgload here https://github.com/r-lib/pkgload/pull/41

PR #1172, opened 2016-05-05
Add ability to choose a file in `source_gist()`

ateucher:
This adds an argument `filename` to the `source_gist()` function so that a user can specify a particular file in a gist that has more than one (currently it will source the first .r file in a gist).

e.g.:
```r
source_gist("gist.github.com/ateucher/0689a1460a0ac136d3b69c371de3a57e", filename = "hi.r")
source_gist("gist.github.com/ateucher/0689a1460a0ac136d3b69c371de3a57e", filename = "hello.r")
```


ateucher:
Awesome, thanks!

PR #1173, opened 2016-05-05
Added citation creator

coatless:
As devtools has shortcuts to build many different R package files, it would be nice to have a way to automatically create a CITATION file using the DESCRIPTION information.


hadley:
Hmmmm, I was initially quite positive about this, but after looking at it closely the interface doesn't seem quite right to me yet. The problem is that this basically generates a citation that's the same as the default you get if you don't have a CITATION file. I think it would be more useful to provide an interface that helped you generate the citation for a published paper.  I'd be happy to re-review if that's what you implement.

I include some code below that I wrote to simply the process of extracting author names. I don't think it's that useful now (since you don't just want to include all the package authors), but I'm putting it here since I did spend a few minutes writing it.

```R
package_authors <- function(pkg = ".", has_role = NULL) {
  pkg <- as.package(pkg)

  authors <- eval(parse(text = pkg$`authors@r`))

  if (!is.null(has_role)) {
    authors <- Filter(function(x) any(has_role %in% x$role), authors)
  }

  authors
}

author_name <- function(x) {
  if (is.null(x$family)) {
    x$given
  } else {
    paste0(x$family, ", ", x$given)
  }
}
```

coatless:
@hadley 

Can-do. 

Is it possible to re-open this PR or should I submit a new one?

hadley:
I think it'll be sufficiently different that it's better to open a new one

PR #1174, opened 2016-05-06
Always unload package DLLs before installation

jimhester:
This also finds the namespace registry once onLoad() rather than
dynamically, so the devtools DLL can be unloaded in load_all() without
needing ns_registory() when creating a new namespace.

As far as I am aware the namespace registry environment is only created once per session, so it should remain stable after it is read in .onLoad().

The main impetus for this PR was a comment by @richfitz on slack
> Having been through a bunch of pain with windows users with devtools this month, I would advise going a CRAN/drat/local cran route.  Recent devtools is enthusiastic about upgrading packages which leaves R sessions in unrecoverable positions as curl and devtools get upgraded unsuccessfully due to file locking.

Users could still have issues with prior versions of devtools, but after this newer devtools versions (and other packages) should now upgrade more cleanly on Windows.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # devtools 1.11.1.9000
 
+* Always unload package DLLs before installation (including devtools) so new
+  DLLs can be copied properly on Windows. (@jimhester)
+
 * `install_*` functions and `update_packages()` refactored to allow updating of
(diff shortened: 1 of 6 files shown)


richfitz:
Thanks Jim! More than happy to test things out here on our windows users.

PR #1177, opened 2016-05-09
Add use_catch and use_inst arguments to use_testthat() and use_rcpp()

Ironholds:
Two patches, really.

Patch the first adds use_catch = FALSE as an argument to use_testthat(); it's awesome to be able to take advantage of testthat's "catch" integration but kind of funky that devtools doesn't have a way of initialising it yet - testthat does, but, well, this is what devtools is _for_. Users will now be able to use `use_testthat()` to trivially set up not only R-side but C++-side unit testing.

Patch the second (less generally useful, perhaps, but I spend a disturbing amount of my time hunting for examples of this and cussing) addresses the phenomenon of Rcpp packages with `/inst/include/somefile.h`, intended to be reused by other packages with `[[Rcpp::depends()]]`, and an associated `Makevars` line to expose that code within the package and thus make it testable and directly accessible. With the new `use_inst` argument to `use_rcpp` - set to FALSE by default for backwards-compatibility purposes - users don't need to hunt _or_ cuss, they can just set it to TRUE and watch as initialising Rcpp also initialises a directory for exportable code, and a Makevars flag for accessing it.

I'm hoping that these patches will be found to be useful, well-designed and viable, because I did some serious damage to a fresh manicure typing them out.

### DIFF
--- a/R/infrastructure.R
+++ b/R/infrastructure.R
@@ -40,7 +40,7 @@ use_testthat <- function(pkg = ".", use_catch = FALSE) {
   )
 
   if (use_catch) {
-    message("* Adding testthat to LinkingTo")
+    message("Adding framework for Catch unit tests")
     add_desc_package(pkg, "LinkingTo", "testthat")
     use_template("test-example.cpp", "src/test-example.cpp")
     use_template("test-runner", "src/test-runner.cpp")
@@ -116,10 +116,19 @@ use_vignette <- function(name, pkg = ".") {
 }
 
 #' @section \code{use_rcpp}:
-#' Creates \code{src/} and adds needed packages to \code{DESCRIPTION}.
+#' Creates \code{src/} and adds needed packages to \code{DESCRIPTION}. Optionally,
+#' if \code{use_inst} is TRUE, creates an \code{/inst/include} directory (and
+#' links to it in the \code{Makevars} file, so you can access the code stored there
+#' from within your package) for Rcpp code you expect to be directly depended-on by
+#' other R packages.
+#'
+#' @param whether to also initialise \code{/inst/include} (a common directory for
+#' hosting C++ code you expect to be directly-depended on by other R packages)
+#' and associated flags.
+#'
 #' @export
 #' @rdname infrastructure
-use_rcpp <- function(pkg = ".") {
+use_rcpp <- function(pkg = ".", use_inst = FALSE) {
   pkg <- as.package(pkg)
   check_suggested("Rcpp")
 
@@ -132,6 +141,11 @@ use_rcpp <- function(pkg = ".") {
   message("* Ignoring generated binary files.")
   ignore_path <- file.path(pkg$path, "src", ".gitignore")
   union_write(ignore_path, c("*.o", "*.so", "*.dll"))
+
+  if (use_inst) {
+    use_directory("inst/include/", pkg = pkg)
+    use_template("inst_include_template", "/src/Makevars")
+  }
 
   message(
     "Next, include the following roxygen tags somewhere in your package:\n\n",
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,12 @@
 * `install_*` functions and `update_packages()` refactored to allow updating of
   packages installed using any of the install methods. (@jimhester, #1067)
 
+* `use_testthat()` now (optionally) initialises the "catch" framework for
+  unit-testing package C++ code.
+
+* `use_rcpp()` optionally adds an `/inst/include` directory for users writing
+  Rcpp code they expect other packages to directly depend on.
+  
 # devtools 1.11.1
 
 * Bug fix in `search_path_for_rtools()` using the gcc-4.9.3 toolchain when
--- a/inst/templates/inst_include_template
+++ b/inst/templates/inst_include_template
@@ -1,0 +1,1 @@
+PKG_CXXFLAGS=-I../inst/include
--- a/man/infrastructure.Rd
+++ b/man/infrastructure.Rd
@@ -27,7 +27,7 @@ use_rstudio(pkg = ".")
 
 use_vignette(name, pkg = ".")
 
-use_rcpp(pkg = ".")
+use_rcpp(pkg = ".", use_inst = FALSE)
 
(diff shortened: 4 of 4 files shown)


Ironholds:
Oh, also I fixed a couple of warnings in the test suite caused by calling deprecated testthat functions.

hadley:
LGTM. @kevinushey can you take a quick look too please?

jimhester:
This is missing the parameter name at https://github.com/hadley/devtools/pull/1177/files#diff-5edd38b303cb4b300beb72d1161397edR125 which is causing travis to fail https://travis-ci.org/hadley/devtools/jobs/128773783#L3741-L3745.

R/infrastructure.R
```diff
@@ -33,6 +39,12 @@ use_testthat <- function(pkg = ".") {
     pkg = pkg
   )
 
+  if (use_catch) {
```
kevinushey: This won't get run if `testthat` infrastructure (minus Catch) has already been initialized (e.g. you can't run `devtools::use_testthat(catch = TRUE)` to add Catch to a package already using `testthat`).

Could that be alleviated? E.g. move this to a separate function, `use_catch()`), and call it earlier in the `use_testthat()` call?

R/infrastructure.R
```diff
@@ -33,6 +39,12 @@ use_testthat <- function(pkg = ".") {
...
+    message("Adding framework for Catch unit tests")
+    add_desc_package(pkg, "LinkingTo", "testthat")
+    use_template("test-example.cpp", "src/test-example.cpp")
```
kevinushey: It looks like this fails if the `src/` directory does not exist already.

R/infrastructure.R
```diff
@@ -33,6 +39,12 @@ use_testthat <- function(pkg = ".") {
...
+    use_template("test-runner", "src/test-runner.cpp")
```
kevinushey: Should this be `use_template("test-runner.cpp", "src/test-runner.cpp")?

R/infrastructure.R
```diff
@@ -33,6 +39,12 @@ use_testthat <- function(pkg = ".") {
...
+  }
```
kevinushey: I think you also need the `R` template file: https://github.com/hadley/testthat/blob/master/R/test-compiled-code.R#L183-L187

R/infrastructure.R
```diff
@@ -121,6 +142,11 @@ use_rcpp <- function(pkg = ".") {
   ignore_path <- file.path(pkg$path, "src", ".gitignore")
   union_write(ignore_path, c("*.o", "*.so", "*.dll"))
 
+  if (use_inst) {
+    use_directory("inst/include/", pkg = pkg)
+    use_template("inst_include_template", "/src/Makevars")
```
kevinushey: FWIW the 'overwrite?' dialog for `use_template` is kind of awkward here (you either need to accept that your original Makevars is overwritten, or you get an error when you cancel / reject the dialog)

Ironholds:
Thanks for feedback and thoughts all! The impression I'm getting is:

1. The actual idea isn't a bad one;
2. The implementation needs work.

Is this accurate? If so I'll tweak things in line with Jim and Kevin's excellent feedback this weekend (I'm spending my free time today doing Foundation-related shi- uh. Stuff.)

kevinushey:
Yes indeed -- thanks for being the one to get started on this work!

hadley:
@Ironholds do you think you'll be able to do this in the next week? If yes, we can get it into the next version of devtools; if no, it'll have to wait

Ironholds:
Absolutely! This coming weekend is most likely (last weekend was Survey
Release Hell)

R/infrastructure.R
Ironholds: Good, hah, catch!

R/infrastructure.R
Ironholds: Makes sense; do you think use_catch() should be exported as well, or?

R/infrastructure.R
Ironholds: Is that a patch issue or a "we should look into that" issue or?

R/infrastructure.R
kevinushey: I think that makes sense -- export `use_catch()` on the `devtools` side, use that function within the `use_testthat()` call when `catch = TRUE`, and (after this is merged) we could deprecate `use_catch()` within `testthat`. Does that make sense, @hadley?

R/infrastructure.R
kevinushey: I think this behavior would be a bit friendlier:

1. Only apply the template if `src/Makevars` doesn't exist already, or
2. Amend the `Makevars` file to add `PKG_CPPFLAGS = ../inst/include` if it already exists.

I would prefer having behavior 1) in this PR so that we don't overwrite a users potentially hand-crafted / fully populated Makevars file; 2) is more of a 'nice-to-have'.

kevinushey: Ah, and now that I'm looking at it -- should we also generate `src/Makevars.win`?

R/infrastructure.R
hadley: Maybe it would be easier to just have this as `use_catch()` rather than an argument? `use_catch()` would call `use_testthat()` if the testing infrastructure hadn't already been set up (in the same way that `use_test()` does)

Ironholds: Yeah, that's sort of what I was thinking - presumably it'd call use_rcpp() too since there's the breaking edge case where catch is called but /src/ doesn't exist? Unless I'm horribly overthinking it.

R/infrastructure.R
Ironholds: Good call. I'll try for 2, fall back to 1.

R/infrastructure.R
kevinushey: I don't think we need `use_rcpp()` for Catch (it doesn't depend on Rcpp); I think `use_catch()` should be able to handle this without also adding Rcpp infrastructure. Does that make sense?

hadley:
Sorry for never merging this :(  And this code has migrated to the usethis package. Would you mind resubmitting over there?

PR #1183, opened 2016-05-10
Installing a GitHub remote when package is installed from CRAN (and various other fixes)

krlmlr:
To reproduce error fixed by this PR:

- Install dplyr 0.4.3 from CRAN
- Install hadley/dtplyr@ec0aaea8cb452 from GitHub

I'm not sure why parse_one_remote() tried to read information from the *installed* package. This PR fixes my use case, but might break others.

Now package2remote() is unused, but it might get a different purpose later, so I left it in.


krlmlr:
This grew bigger than anticipated, let me know if you like me to split it.

R/deps.R
```diff
@@ -167,17 +167,7 @@ parse_one_remote <- function(x) {
       envir = asNamespace("devtools"), mode = "function", inherits = FALSE),
     error = function(e) stop("Unknown remote type: ", type, call. = FALSE))
 
-  simple_remote <- fun(repo)
```
hadley: Why does this code go away?

krlmlr: package2remote(remote_package_name(...)) looks at the *installed* package, which may have a different kind of remote than the requested remote.

jimhester: Yeah removing this looks right, I just ran into this as well when debugging Kirills' dtplyr issue.

jimhester:
As of 39222f8cbbc82cdc719a48746bc6b912a45a58a5 `install_github("hadley/dtplyr@ec0aaea")` now works on my machine regardless of the state of dplyr, installed from CRAN, dev version installed from GitHub or locally and not installed dtplyr is installed properly regardless.

PR #1184, opened 2016-05-10
Install Suggested package after package installation

jimhester:
This includes packages specified in Remotes:. If build_vignettes is TRUE
they are still installed before the package.

This fixes hadley/dplyr#1809

### DIFF
--- a/R/install.r
+++ b/R/install.r
@@ -96,9 +96,15 @@ install <-
 
   # If building vignettes, make sure we have all suggested packages too.
   if (build_vignettes && missing(dependencies)) {
-    dependencies <- TRUE
+    dependencies <- standardise_dep(TRUE)
+  } else {
+    dependencies <- standardise_dep(dependencies)
   }
-  install_deps(pkg, dependencies = dependencies, upgrade = upgrade_dependencies,
+
+  initial_deps <- dependencies[dependencies != "Suggests"]
+  final_deps <- dependencies[dependencies == "Suggests"]
+
+  install_deps(pkg, dependencies = initial_deps, upgrade = upgrade_dependencies,
     threads = threads, force_deps = force_deps, quiet = quiet, ...)
 
   # Build the package. Only build locally if it doesn't have vignettes
@@ -123,6 +129,9 @@ install <-
   built_path <- normalizePath(built_path, winslash = "/")
   R(paste("CMD INSTALL ", shQuote(built_path), " ", opts, sep = ""),
     quiet = quiet)
+
+  install_deps(pkg, dependencies = final_deps, upgrade = upgrade_dependencies,
+    threads = threads, force_deps = force_deps, quiet = quiet, ...)
 
   if (length(metadata) > 0) {
     add_metadata(inst(pkg$package), metadata)
--- a/R/install.r
+++ b/R/install.r
@@ -104,6 +104,11 @@ install <-
   initial_deps <- dependencies[dependencies != "Suggests"]
   final_deps <- dependencies[dependencies == "Suggests"]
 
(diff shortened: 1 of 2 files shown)


jimhester:
@krlmlr The main issue with this currently is it queries the remotes twice, but getting rate limited seems to be less of an issue with the embedded PAT now.

jimhester:
@hadley if the small hack used in https://github.com/hadley/devtools/pull/1184/commits/7f54aaf4c9f2b31d498d096d98004d4a906b8fb5 (use of the `installed` environment to store state) is OK I can merge this, I verified it correctly handles the dplyr -> dtplyr issue in both directions.

R/install.r
```diff
@@ -96,9 +96,20 @@ install <-
...
   if (build_vignettes && missing(dependencies)) {
-    dependencies <- TRUE
+    dependencies <- standardise_dep(TRUE)
+  } else {
+    dependencies <- standardise_dep(dependencies)
   }
-  install_deps(pkg, dependencies = dependencies, upgrade = upgrade_dependencies,
+
+  initial_deps <- dependencies[dependencies != "Suggests"]
+  final_deps <- dependencies[dependencies == "Suggests"]
```
krlmlr: What about "Enhances"?

krlmlr:
Does this also work for install_deps(".") for dplyr?

R/install.r
jimhester: I think Enhances should be fine to installed in the initial installation. Note `Enhances:` are not installed unless explicitly set by default (see ?install.packages).
> logical indicating whether to also install uninstalled
packages which these packages depend on/link
to/import/suggest (and so on recursively).  Not used if
‘repos = NULL’.  Can also be a character vector, a subset of
‘c("Depends", "Imports", "LinkingTo", "Suggests",
  "Enhances")’
          
> Only supported if ‘lib’ is of length one (or missing), so it is
unambiguous where to install the dependent packages.  If this is not the case
it is ignored, with a warning.

>  The default, ‘NA’, means ‘c("Depends", "Imports",
  "LinkingTo")’.

> ‘TRUE’ means to use ‘c("Depends", "Imports", "LinkingTo",
  "Suggests")’ for ‘pkgs’ and ‘c("Depends", "Imports",
  "LinkingTo")’ for added dependencies: this installs all the
packages needed to run ‘pkgs’, their examples, tests and
vignettes (if the package author specified them correctly).

> In all of these, ‘"LinkingTo"’ is omitted for binary
packages.

jimhester:
> Does this also work for install_deps(".", dependencies = TRUE) for dplyr?

Yes, that seems to work fine on my machine

R/install.r
krlmlr: What happens if there's a dependency chain like A -> (enhances) B -> (imports) A and I try to install(A) or install_deps(A)?

krlmlr:
I can confirm that dtplyr -> dplyr works (https://travis-ci.org/krlmlr/dplyr/builds/129929004), but not dplyr -> dtplyr -> dplyr if I do install_deps(dependencies = TRUE) (https://travis-ci.org/krlmlr/dplyr/builds/129933568). I think the "release" build always runs with an empty cache because it's failing.

jimhester:
Looks like you forgot to put dtplyr back in `Remotes:` on that test https://github.com/krlmlr/dplyr/blob/4b2911fa657df1b0e4cf03ad2ca4b81c95df489b/DESCRIPTION

krlmlr:
Thanks. Of course I forgot, now it works. My mistake.

That's just awesome! Are you planning a release anytime soon?

PR #1190, opened 2016-05-23
Disable JIT during/at compiler package loading/unloading.

kalibera:
This patch is needed to make devtools pass its own tests with the JIT enabled. The patch disables JIT compilation temporarily during loading the compiler namespace (because otherwise the JIT compilation during the loading process interferes, it loads the compiler namespace faster than devtools and locks it before devtools can insert shims into it). The patch also disables JIT (permanently) before unloading the compiler namespace - there seems to be no other way than keeping it disabled, because the call to enable it would immediately load it, which would make the unloading a waste of time. It may not be so important after all to be able to load/unload the compiler package (it should perhaps be loaded at all times anyway), but the devtools unit tests depend on this, and hence the patch.

### DIFF
--- a/R/load.r
+++ b/R/load.r
@@ -92,6 +92,15 @@ load_all <- function(pkg = ".", reset = TRUE, recompile = FALSE,
 
   if (!quiet) message("Loading ", pkg$package)
 
+  if (pkg$package == "compiler") {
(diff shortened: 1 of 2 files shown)


hadley:
Thanks!

PR #1192, opened 2016-05-25
Install binary packages from zip files directly on windows

jimhester:


R/install-remote.R
```diff
@@ -27,6 +27,11 @@ install_remote <- function(remote, ..., force = FALSE, quiet = FALSE) {
     return(invisible(FALSE))
   }
 
+  if (is_windows && inherits(remote, "cran_remote")) {
```
hadley: I think it would be a little safer to condition on cran binary packages

hadley:
Also needs a bullet in NEWS

PR #1194, opened 2016-05-26
Add install_bioc to install bioconductor packages from their SVN.

jimhester:
This is not quite done, but wanted to get it out there.

The scope grew slightly during the process, this fixes some issues I found with the new package dependency code as well.


jimhester:
This should be ready to be reviewed now, seems to work well.

hadley:
It looks fine to me, but you're probably more familiar with this code than I am, so take it with a grain of salt. Feel free to merge.

PR #1196, opened 2016-05-31
use GitHub tarball API when downloading packages

kevinushey:
This PR modifies `install_github` to use the `tarball` endpoint rather than `zipball`.

`devtools` uses its own utility `decompress` function behind the scenes, which farms out to an appropriate decompression utility based on the file extension, so everything should just work here (and appears to in some manual testing)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # devtools 1.11.1.9000
 
+* `install_github()` now downloads tarballs rather than zipballs. (@kevinushey)
+
 * Suggested packages, including those specified as `Remotes:` are now installed
   after package installation. This allows you to use circular `Remotes:`
   dependencies for two related packages as long as one of the dependencies is a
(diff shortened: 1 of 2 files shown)


jimhester:
Is there anything in particular motivating the change? looks like for devtools the zip file is 604Kb, while the tar is 456K, so it seems worthwhile just for the the space/transmission savings.

Reading `?untar` it looks like there are compatible implementations for Windows, which would be my other major concern.

kevinushey:
This PR is motivated by packrat -- it tries to obtain package sources (as a tarball) and stores them alongside the project; we want to use the `devtoools` download machinery when appropriate (as this can handle GitHub authentication) and this change should help make it easier to obtain package sources as tarballs from GitHub.

imanuelcostigan:
We have been having very strange issues using install_github at work where on some Windows machines, the zip file is showing up as corrupted an can't be unpacked, while on others there are no issues. Will be interesting to see whether this change makes this issue go away. If so, 👍

PR #1203, opened 2016-06-03
Pick https mirror if possible.

### LINKED ISSUE #1059: http:// vs https:// in hardcoded repositories
There are a few places where the RStudio CRAN mirror (`cran.studio.com`) is hard-coded, for convenience (e.g. functions that will get used non-interactively, where user may not have set repo in their .Rprofile).  Right now these are hard-coded as `http://...`, which causes various difficulties when using R-devel (which insists on `https://` in many important contexts) - specifically, dependencies may fail to be found when installing packages.   In particular, changing `http://` to `https://` in line 56 of `deps.R` fixes a lot of problems ...

hadley:
Fixes #1059 

cc @jimhester, @bbolker 

@jjallaire Can you please confirm that it's ok for me to re-license RStudio source in this way?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,7 +8,7 @@
   on your system (#1060)
 
 * devtools now uses https to access the RStudio CRAN mirror if it will work
-  on your system (#1060)
+  on your system (#1059)
 
 * `revdep_check()` will now create the `revdep/` directory if it does not
   already exist (#1178).
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,9 +5,6 @@
   `NEWS.md` (if present), and checking into git (if you use it) (#1076.)
 
 * devtools now uses https to access the RStudio CRAN mirror if it will work
-  on your system (#1060)
-
-* devtools now uses https to access the RStudio CRAN mirror if it will work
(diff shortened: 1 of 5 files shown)


jimhester:
I think the logic needs to be tweaked to include `capabilities("libcurl")` somewhere. Currently on my machine `download_method_secure() # FALSE` even though `capabilities("libcurl") #TRUE`

jjallaire:
In R 3.3 a download method of "auto" will actually result in "libcurl" on Unix alikes, form the docs:

> If method = "auto" is chosen (the default), on a Unix-alike method "libcurl" is chosen for https:// and ftps:// URLs and the "internal" method is chosen for other schemes.

So you may want to update the auto_download_method function to reflect that.

It's fine to relicense the source. You should keep the copyright notice but just change the license for that bit to whatever is used for the package (GPL-2 it looks like).

jjallaire:
Jim, I think the change I suggested might just pick that up  (i.e. if the
download method is "auto" then "libcurl" is used on Unix alikes).

hadley:
PTAL - I added a check for libcurl and removed the RStudio license block.

PR #1205, opened 2016-06-03
Call fetch() before determining sync status with Git remote in release()

### LINKED ISSUE #1128: Function to check that head is up to date from remote

krlmlr:
With new git_checks() which uses the same pattern as release_checks(). Failing to fetch, or trying to release a tag, doesn't abort the process, just shows an error:

```
> git_checks()
Running Git checks for devtools
Checking uncommitted files... OK
Checking if synched with remote branch...
ERROR: Error in 'git2r_remote_fetch': Failed to resolve address for github.com: Temporary failure in name resolution
```

Fixes #1128.


hadley:
LGTM

R/release.r
```diff
@@ -55,15 +55,9 @@ release <- function(pkg = ".", check = TRUE, args = NULL) {
...
-      if (yesno("Uncommited changes in git. Proceed anyway?"))
-        return(invisible())
-    }
-
-    if (git_sync_status(pkg$path)) {
-      if (yesno("Git not synched with remote. Proceed anyway?"))
-        return(invisible())
-    }
+    git_checks(pkg)
+    if (yesno("Were Git checks successful?"))
```
krlmlr: At some point we may want to avoid this question when we "know" that all checks succeeded.

hadley: It's currently so you can ignore if needed — i.e. I don't always actually care that there are uncommitted files but I want to know about it.

Also, I don't actually care if there are local changes that haven't been pushed to the remote; but I _do_ care if there are remote changes I don't have locally. How hard would it be to make that work?

krlmlr: Not hard at all, a simple tweak to git_sync_status(). Will change this.

krlmlr:
Tested cases: ahead, behind, and offline. Works for me, ready to merge.

Issue #596, opened 2014-09-24
Exports diagnostic


hadley:
To replace `missing_s3()`.

Should list all package contents broken down by type (functions/generics, S3 methods, S4 methods, S4 classes), and for each component, indicate whether or not it's exported.

hadley:
This no longer seems that important to me. It would basically print out the NAMESPACE.

hadley:
It's not important to me, but it seems like it would be helpful for other people. The key would be to show what's _not_ exported.

hadley:
It would be nice, but it's a lot of work and unlikely to get to the top of my priority list.

PR #1210, opened 2016-06-06
Do not export functions when running tests

### LINKED ISSUE #1201: Let users run `test` with `export_all=FALSE`
At the moment `test` quietly calls `load_all` with the default `export_all=TRUE`. This means:

- tests for S3 dispatch will pass when they shouldn't (e.g. `print.foo` will work even if it's not exported).
- internal package functions will be left in the search path after `test` has run.

jimhester:
Fixes #1201

### DIFF
--- a/R/test.r
+++ b/R/test.r
@@ -47,7 +47,7 @@ test <- function(pkg = ".", filter = NULL, ...) {
   # Run tests in a child of the namespace environment, like
   # testthat::test_package
   message("Loading ", pkg$package)
-  ns_env <- load_all(pkg, quiet = TRUE)$env
+  ns_env <- load_all(pkg, quiet = TRUE, export_all = FALSE)$env
 
   message("Testing ", pkg$package)
   Sys.sleep(0.05); utils::flush.console() # Avoid misordered output in RStudio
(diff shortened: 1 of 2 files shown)


R/test.r
```diff
@@ -47,7 +47,7 @@ test <- function(pkg = ".", filter = NULL, ...) {
   # Run tests in a child of the namespace environment, like
   # testthat::test_package
   message("Loading ", pkg$package)
-  ns_env <- load_all(pkg, quiet = TRUE)$env
+  ns_env <- load_all(pkg, quiet = TRUE, export_all = FALSE)$env
```
hadley: But won't this remove unexported functions from the interactive environment if they previously had used `load_all()`?

jimhester: Yeah it does (just checked). I wonder we could come up with an easy test for fully exported namespace and set `export_all` based on that...

jimhester:
https://github.com/hadley/devtools/pull/1210/commits/612c9d5d61fb63759361e728acfcce3e1f2dc564 will now preserve the prior export status when running `test()`.

hadley:
I'd rather tackle this meta-problem by providing better tools to detect which functions aren't important, i.e. #596

PR #1215, opened 2016-06-07
Use GitHub host

ijlyttle:
Addresses #1101, cleans up #1142 

(Thanks for continued patience!)

I reworked things a bit, to put all the url logic into `use_github()`, along with some comments that hopefully lay out the motivation. 

In short, the host argument to `use_github()` must contain a protocol and a hostname, and may include a path to describe the endpoint-root - per your last suggestion on the previous PR

My claims:

- by manual test, `use_github()` remains working for host = "https://api.github.com"
- by manual test, `use_github()` works for host = "https://github.hostname.com/api/v3"
- all the tests: `test(filter = "github")` remain successful (thanks @jennybc!)

Happy to amend as needed.


R/infrastructure-git.R
```diff
@@ -102,19 +107,65 @@ use_github <- function(auth_token = github_pat(), private = FALSE, pkg = ".",
...
+  # github_POST() needs a path and a host (protocol and hostname)
+  #
+  # In the default case, the supplied value for host is
+  # "https://api.github.com". For github_POST(), the host
+  # will not change and path will be "user/repos".
+  #
+  # If you are using enterprise github, you may have supplied a host that
+  # looks like "https://github.hostname.com/api/v3". For github_POST(),
+  # the host will be "https://github.hostname.com", and path will be
+  # "api/v3/user/repos".
```
hadley: It feels like this logic should be in `github_POST()` not here

R/infrastructure-git.R
```diff
@@ -102,19 +107,65 @@ use_github <- function(auth_token = github_pat(), private = FALSE, pkg = ".",
...
+  #
+  # If the hostname is "api.github.com", hostname becomes "github.com";
+  # other hostnames are not changed. The protocol is preserved.
+  #
+  # Any path in the url is removed.
+  #
+  # The resulting url, which should have only a protocol and hostname, is
+  # rebuilt and sent to use_github_links()
+  #
+  url <- httr::parse_url(host)
```
hadley: And this should be pulled out into a function like `github_url_from_host()` or similar.

ijlyttle:
Have added a couple of internal functions, and tests for them:

- `github_reg_host_path(host, path)` returns a regularized host and path
- `github_url_from_host(host)` returns the url for github links

`github_reg_host_path()` is implemented in `github_POST()` and in `github_GET()` (for completeness).

`github_url_from_host(host)` is implemented in `use_github()` to send the correct host to `use_github_links()`.

All claims (manual tests of `use_github()` and `test(filter = "github")`) re-checked.

Hopefully, this is closer. 

I'll have a look at why the CI is not returning green checkmarks.

R/git.R
```diff
@@ -135,10 +135,10 @@ github_remote_parse <- function(x) {
 
   if (grepl("^https", x)) {
     # https://github.com/hadley/devtools.git
-    re <- "github.com/(.*?)/(.*)\\.git"
+    re <- "github[^/]*/(.*?)/(.*)\\.git"
   } else if (grepl("^git", x)) {
     # git@github.com:hadley/devtools.git
-    re <- "github.com:(.*?)/(.*)\\.git"
+    re <- "github[^:]*:(.*?)/(.*)\\.git"
```
jimhester: If both of these regexes are exactly the same there is no need for these conditionals.

R/github.R
```diff
@@ -30,16 +30,59 @@ github_error <- function(req) {
...
+    path <- file.path(path_prefix, path)
+  }
+
+  list(host = host, path = path)
+}
+
+github_GET <- function(path, ..., pat = github_pat(),
+                       host = "https://api.github.com") {
+  # regularize the host and path
+  reg <- github_reg_host_path(host = host, path = path)
```
jimhester: `httr:GET()` can accept a url object directly, so this can be replaced by the following.
```r
url <- httr::parse_url(host)
url$path <- paste(url$path, path, sep = "/')
req <- httr::GET(url, github_auth(pat), ...)
```

R/git.R
hadley: Good point! I guess they used to be different?

R/infrastructure-git.R
```diff
@@ -190,10 +241,12 @@ use_git_ignore <- function(ignores, directory = ".", pkg = ".") {
 #' those fields already exist.
 #'
 #' @inheritParams use_git
+#' @param host GitHub host to use. Override with your GitHub enterprise
+#'    hostname, for example, "https://github.hostname.com".
 #' @family git infrastructure
 #' @keywords internal
 #' @export
-use_github_links <- function(pkg = ".") {
+use_github_links <- function(pkg = ".", host = "https://github.com") {
```
jimhester: I think this function should be rewritten to query the API for the repo and return the html_url and issues_url from the response (https://developer.github.com/v3/repos/#response-3) rather than special casing the api.github.com url.

The issues_url returns `https://api.github.com/repos/hadley/devtools/issues{/number}` so the `{/number}` needs to be stripped first.

```r
res <- github_GET("repos/hadley/devtools")
github_URL <- res$html_url
issues_URL <- sub("\\{/number\\}", "", res$issues_url)
```

ijlyttle: I'm happy to make the change for `github_URL`, but I don't think `issues_URL` is returning what we want. The hostname in `issues_URL` is `api.github.com`, where we want `github.com`

As a first approach, could I just append `"/issues"` to `github_URL`?

jimhester: Ah you are right, my apologies, I think appending '/issues' to `html_url` is the right thing then.

R/git.R
ijlyttle: There's likely something I am not seeing here - which are the regexes that are the same?

jimhester: L138 and L141

ijlyttle: Here's what I have for L138 and L141:

```R
    re <- "github[^/]*/(.*?)/(.*)\\.git" # L138
    re <- "github[^:]*:(.*?)/(.*)\\.git" # L141
```

I think the difference is in the delimiter, `/` for L138 and `:` for L141, to distinguish

```
https://github.com/hadley/devtools.git # L138 case
git@github.com:hadley/devtools.git     # L141 case
```

I think the two might be combined, as I don't think that either delimiter is legal part of a hostname in either case.

What do you think?

R/github.R
ijlyttle: Placeholder to say that I am noodling this out - I am just a bit slow :)

R/git.R
jimhester: Ah you are right about them being different, my mistake. 

If we are just ignoring everything but the `user/repo` slug they might as well be combined I think.

ijlyttle: Cool, I'll get on combining those.

R/github.R
jimhester: I am not seeing this behavior

``` r
host <- "https://api.github.com"
path <- "users/hadley/repos"

url <- httr::parse_url(host)
url$path <- paste(url$path, path, sep = "/")

httr::build_url(url)
#> [1] "https://api.github.com/users/hadley/repos"
httr::GET(url)
#> Response [https://api.github.com/users/hadley/repos]
#>   Date: 2016-06-08 20:20
#>   Status: 200
#>   Content-Type: application/json; charset=utf-8
#>   Size: 155 kB
#> [
#>   {
#>     "id": 40423928,
#>     "name": "15-state-of-the-union",
#>     "full_name": "hadley/15-state-of-the-union",
#>     "owner": {
#>       "login": "hadley",
#>       "id": 4196,
#>       "avatar_url": "https://avatars.githubusercontent.com/u/4196?v=3",
#>       "gravatar_id": "",
#> ...
```

ijlyttle: Yikes! I'll try again...

ijlyttle: I'll poke around - I would sure like to get your behavior...

jimhester: Looks like the difference is https://github.com/hadley/httr/commit/2f6baa0b98c9eaf632acc43ede3be2b187893ba3 (I am using devel httr)

ijlyttle: Good news for the future!

In that case, I'll look into  hadley/httr@2f6baa0  and see where the difference is, I'll propose a very small bit of code to cover the difference until the next CRAN release of httr (I'll leave a comment/issue to remove it from devtools).

Almost there!

R/git.R
ijlyttle: done in af11ad7

R/infrastructure-git.R
ijlyttle: done in 3263cee

R/github.R
ijlyttle: done in a8d5d29 - there is some code that can be taken out when the next httr is released

ijlyttle:
First - thanks @jimhester for your feedback and patience today

I think all the boxes are checked - I was able to delete a lot of code. 

There are a couple of lines in `github_GET()` and `github_POST()` that should be able to be deleted at the next release of httr.

All previous claims (manual test of `use_github()` on github.com and enterprise, `test(filter = "github")`) verified.

hadley:
LGTM. @jimhester are you ok with merging?

jimhester:
LGTM merge away

jimhester:
And thanks for your patience and work Ian!

hadley:
Thanks Ian!

Issue #1163, opened 2016-04-28
installing from git(hub) with submodules


jonkeane:
Currently git(hub) repositories with submodules can't be installed. There is a warning when using `install_github()`, but after the switch to `git2r` it looks like `args="--recursive"` are no longer supported in `install_git()`.

I've looked through git2r and I don't see any support for submodules there, but I might be missing it somewhere.

Would it be possible (and more importantly advisable) to add an option to check for a `.gitmodules` file when the repository is downloaded and then also download the submodules specified there before installing the package? I would be happy to take a stab at adding this and sending a pull request, but I want to make sure this isn't a terrible idea first.

This is an off shoot from Issue #751

hadley:
Yeah, I think that's a reasonable thing to add

vspinu:
@jonkeane, have you dug on this further? It's quite a headache I must say.

jonkeane:
I haven't looked at this since I sent the PR. If you check out the code at that point it should work, but let me know if there are any problems and I'll see what I can do to fix them.

jimhester:
Also a workaround with CRAN devtools is to use something like the following https://github.com/hadley/devtools/pull/751#issuecomment-215540807

rtobar:
I'm also eager to see these changes included in devtools. Thus, I rebased @jonkeane's patch on top of the current master, and added a small improvement on top of it. I commented on the original pull request about this to know whether I should create a new one or if we can reuse it to include my changes as well.

ronammar:
Has this feature been incorporated into `devtools` master?
Would the syntax appear as follows?:

```r
devtools::install_git(myRepoURL, args="--recursive")
```

PR #1222, opened 2016-06-12
Added functions to clone git submodules when they are present

jonkeane:
`install_git()` now has the option `submodules`. When it is `submodules = TRUE` the submodules that are listed in the `.gitmodules` file are cloned before the package is installed.

There were a number of functions that needed to be added for parsing the `.gitmodule` files, making sure ssh request urls are translated to https request urls. This does not (yet) include the submodule option for `install_github()`. Although I'm happy to add it, if this approach works for the authors. I've written tests for the one function that has no side effects.

This addresses #1163

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # devtools 1.11.1.9000
 
+* Added a `submodules` option to `install_git` to download the submodules if there
+  is a `.gitmodules` in the repository. Updated the documentation for `install_github()`
+  to reflect this option. (@jonkeane, #1163, #751)
+
 * Handle case when a GitHub request returns a non-JSON error response.
   (@jimhester, #1204, #1211)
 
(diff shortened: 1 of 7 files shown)


rtobar:
I have an updated version of this patch on top of the current master, plus a small improvement. Should I open a new pull request for it? Otherwise, how can I contribute to this one?

hadley:

bschilder:
@jonkeane Has the `submodules=` arg since been removed?

```R
devtools::install_git("https://github.com/RajLabMSSM/echofinemap.git", submodules=TRUE)
devtools::install_github("RajLabMSSM/echofinemap", submodules=TRUE)
```

```
Warning message:
Arguments in `...` must be used.
✖ Problematic argument:
• submodules = TRUE 
```

PR #1225, opened 2016-06-14
Fixes to ensure package2remote respects an updated .libPath()

jimhester:
packageDescription() loads the description from the namespace if it is
already loaded, which is not what we want for e.g. revdep_check().

This forces the lookup to be on the current .libPaths() only.

The ceiling parameter in gitr2::discover_repository() is used so that
use_git() only returns true if the input path is the base of the git
repository. This prevents test packages from being considered to be
using git.

### DIFF
--- a/tests/testthat/test-remotes.r
+++ b/tests/testthat/test-remotes.r
@@ -1,12 +1,4 @@
 context("remote_deps")
-with_temp_libpaths <- function(code) {
-  path <- tempfile("temp_libpath")
-  dir.create(path)
-  old <- .libPaths(path)
-  on.exit(.libPaths(old))
-  force(code)
-}
-
 test_that("remote_deps returns NULL if no remotes specified", {
   expect_equal(remote_deps("testTest"), NULL)
 })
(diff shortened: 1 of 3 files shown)


tests/testthat/test-remotes.r
```diff
@@ -82,3 +74,17 @@ test_that("remote_sha.github_remote returns NA if remote doesn't exist", {
 test_that("remote_sha.github_remote returns expected value if remote does exist", {
   expect_equal(remote_sha(github_remote("hadley/devtools@v1.8.0")), "ad9aac7b9a522354e1ff363a86f389e32cec181b")
 })
+
+test_that("packade2remotes looks for the DESCRIPTION in .libPaths", {
```
hadley: sp

hadley:
Sounds reasonable to me. Does this solve a current problem?

jimhester:
I was running into check errors when doing `revdep_check()` for packages with dependencies which are already loaded by devtools. For example digest was not being installed in the revdep_check library because the DESCRIPTION of the loaded namespace was being used instead of looking in the new library, where it was not yet installed. So checks were failing due to missing dependencies.

tests/testthat/test-remotes.r
jimhester: Thanks, fixed!

hadley:
Ooooooooh, that explains a lot of the problems I've been having too

hadley:
LGTM

Issue #1232, opened 2016-06-21
document() should check for the maximum version of R required by imported packages


seankross:
Would it be reasonable if `document()` updated the DESCRIPTION appropriately? I'm working on a PR right now.

hadley:
This feels out of scope for devtools, now that we're splitting it into smaller pieces. 

There isn't an obvious home for it yet, but it seems related to https://github.com/r-lib/pkgload/issues/42, by @billdenney

PR #1233, opened 2016-06-21
document() updates DESCRIPTION with minimum version of R (#1232)

seankross:
An attempt to solve #1232


hadley:
Checking seems too aggressive to me. I'd prefer to start with a function that just tells you what that version is.

Issue #1244, opened 2016-06-28
install_github fails when using devtools 1.12.0 on R 3.1


leonidb:
According to 1.12.0 description, R >= 3.0.2 is supported, however, when using devtools with R 3.1

```
devtools::install_github("...")
```

We get the following error:

```
    Downloading GitHub repo ...
    from URL ...
    Installing ...
    Error in if (capabilities("libcurl")) { : argument is of length zero
    Calls: <Anonymous> ... download_method_secure -> download_method -> auto_download_method
```

The check `if (capabilities("libcurl"))` was added in https://github.com/hadley/devtools/commit/eed074664b441ce34df1643691c0f81afb4c0004, which is part of 1.12.0 release.

When I execute `capabilities()`, `libcurl` indeed doesn't appear on the list, although
`libcurl` library is installed on the machine.

With devtools 1.11.1 the same command  works.

crarlus:
I can confirm this issue for R 3.1. 

```
Error in if (capabilities("libcurl")) { : argument is of length zero
```
Apparently this arises in older versions of R.

In R 3.3 
```
help("capabilities") 
```
states
```
libcurl	
is libcurl available in this build? Used by function curlGetHeaders and optionally by download.file and url. As from R 3.3.0 always true for Unix-alikes, and true for CRAN Windows builds.
```
The entry for libcurl is missing in older versions of R.

Furthermore, in https://cran.r-project.org/doc/manuals/r-release/NEWS.html it was announced
```
CHANGES IN R 3.2.0
	...
    capabilities("libcurl") reports if this is available.

```

Hence, in file [download-method.R](https://github.com/hadley/devtools/blob/master/R/download-method.R) one needs to check if the entry for "libcurl" exists at all, e.g.
```{R}
any(names(capabilities()) == "libcurl")
```

damesek:
That few lines solve the problem at my case:
sudo apt-get install libcurl4-openssl-dev libxml2-dev libxslt-dev libssl-dev -y
sudo apt-get install r-base r-base-dev -y

vext01:
The workaround I used is:
`options(download.file.method = "wget")`

hadley:
We'll be adding automated tests to ensure the devtools works with R 3.1.0 very shortly. We'll see if it's possible to push back to R 3.0.3.

PR #1250, opened 2016-07-04
solution for #1244

kiwiroy:
An option for keeping description correct WRT `install_github()`, especially on R 3.1.

An alternative for those users is:
```R
## method can be one of "curl", "internal", "libcurl" or "wget"
## rather than the default "auto"
options(download.file.method = "curl")
```


hadley:
`isTRUE()` would be a more compact fix

kiwiroy:
@hadley good call, done

hadley:
Thanks!

PR #1253, opened 2016-07-05
Github URLs may lack trailing `.git`

jennybc:
Apparently Github URLs don't require the trailing `.git`. I verified that you can use the plain browser URL in `git clone` and RStudio _New Project > Version Control ..._ and everything works just fine. Until you call `use_travis()`, `use_coverage()`, etc., where the current regex will fail to extract username and repo. The fact you can push and pull with the browser URL "for services like GitHub" even shows up in the Smart HTTP section of [Pro Git](https://git-scm.com/book/ch4-1.html).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # devtools 1.12.0.9000
 
+* Parse valid Git remote URLs that lack trailing `.git`, e.g. GitHub browser URLs (#1253, @jennybc).
+
 * Add a `check_bioconductor()` internal function to automatically install
   BiocInstaller() if it is not installed and the user wants to do so.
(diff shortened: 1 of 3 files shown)


jimhester:
You can avoid `perl = TRUE` with

```r
"github[^/:]*[/:]([^/]+)/(.*?)(?:\\.git)?$"
```

Which also works in all your test cases.

jennybc:
Wish I'd thought of that. OK I upgraded w/ your regex and dropped `perl = TRUE`.

hadley:

jennybc:
Bullet added (and rebased).

PR #1256, opened 2016-07-08
Don't load test helpers twice in test()

krlmlr:
requires testthat >= 1.0.3 (hadley/testthat#505).


hadley:
Thanks!

jimhester:
The testthat PR needs to be merged and testthat needs to add testthat to `Remotes:`, or that PR needs to be added to Remotes to avoid breaking devtools.

Issue #1247, opened 2016-07-01
Invalid comparison operator in dependency: >=


pierucci:
Builds on Travis are [now failing](https://travis-ci.org/pierucci/heemod/builds/141599194) with the following error that seems to come from `devtools`:

```
$ Rscript -e 'deps <- devtools::dev_package_deps(dependencies = TRUE);devtools::install_deps(dependencies = TRUE);if (!all(deps$package %in% installed.packages())) { message("missing: ", paste(setdiff(deps$package, installed.packages()), collapse=", ")); q(status = 1, save = "no")}'
trying URL 'https://cloud.r-project.org/src/contrib/diagram_1.6.3.tar.gz'
Content type 'application/x-gzip' length 466691 bytes (455 KB)
==================================================
downloaded 455 KB
Installing diagram
Error in FUN(X[[i]], ...) : 
  Invalid comparison operator in dependency: >= 
Calls: <Anonymous> ... install -> install_deps -> dev_package_deps -> lapply -> FUN
Execution halted
```

I did not change the DESCRIPTION file. I do not understand the cause of the error because if I read `package-deps.r` line 34 I can see the `>=` operator listed:

``` r
compare_nna %in% c(">", ">=", "==", "<=", "<")
```

HenrikBengtsson:
I'm coming here because I'm seeing the same for the [caper](https://cran.r-project.org/web/packages/caper/) CRAN package.  In this case it's because there's an extra space after the parentheses holding the version specification;

```r
> devtools:::parse_deps("R (>= 2.10), ape (>= 3.0-6) , MASS, mvtnorm")
Error in devtools:::parse_deps("R (>= 2.10), ape (>= 3.0-6) , MASS, mvtnorm") : 
  Invalid comparison operator in dependency: >= 
```
It'll work if one removes that extra space after the "ape" spec;
```r
> devtools:::parse_deps("R (>= 2.10), ape (>= 3.0-6), MASS, mvtnorm")
     name compare version
2     ape      >=   3.0-6
3    MASS    <NA>    <NA>
4 mvtnorm    <NA>    <NA>
```

Looking at the [diagram](https://cran.r-project.org/web/packages/diagram/) package, you'll find the same extra space in the DESCRIPTION:
```r
Depends:	R (≥ 2.01) , shape
```

[...]

PR #1257, opened 2016-07-08
BUG FIX: parse_deps() did not handle spaces after version specification

HenrikBengtsson:
This should resolve Issue #1247 

(I hereby make zero claim on copyright on this simple patch)


jimhester:
This should probably be `"[[:space:]]*,[[:space:]]*"`, tabs are valid in dcf files as well https://www.debian.org/doc/debian-policy/ch-controlfields.html

HenrikBengtsson:
Agree; updated PR.

I incorrectly though `[[:space:]]` required `perl = TRUE` and I didn't want to force that.  I've learned something today :)

Issue #1107, opened 2016-03-05
Let `install_version()` search for version in multiple repos


kenahoo:
TL;DR: I would like to change `install_version()` so that it will search for the requested version of the package in each of the repositories given, in sequence.

Motivation:  my company uses R internally, creating packages for various projects and publishing them to internal CRAN-alike servers for integration by other teams.  We have one “cran-stable” for official releases, and one CRAN-alike that updates with new snapshots every time someone pushes to a “develop” branch.

I would like our team to call `install_version(pkg, version, repos)`, which then iterates over each entry of `repos` until a suitable version is found.  

For instance, we might have the following:
-  http://myserver/cran-stable:
  - MyPackage_0.23.tar.gz
-  http://myserver/cran-dev:
  - MyPackage_0.24-283.tar.gz

```
repos <- c(Stable="http://myserver/cran-stable",
              Dev="http://myserver/cran-dev",
             CRAN="https://cran.revolutionanalytics.com")
install_version('MyPackage', '0.23', repos)  # Installs from ‘Stable’
install_version('MyPackage', '0.24', repos)  # Installs from ‘Dev’
```

Right now, it looks like `install_version()` expects exactly one version of a package to be officially published across all repositories, and it only knows how to look for alternate versions in `archive/` directories, under the assumption that the only alternate versions that could exist are previously-published versions.

Also, because packages in the “Dev” server may have versioned dependencies on packages from either repository, the same semantics would need to apply to the dependencies too, so `install_version()` would call itself recursively to handle them.

I would be happy to work this up as a pull request that keeps the existing semantics of finding packages in `archive/` directories too.

kenahoo:
Looking closer at `install_version()`, it looks like the `version` argument is an exact version required, not a minimum version.  So my example above would need to be:

```
repos <- c(Stable="http://myserver/cran-stable",
              Dev="http://myserver/cran-dev",
             CRAN="https://cran.revolutionanalytics.com")
install_version('MyPackage', '>= 0.23', repos)  # Installs from ‘Stable’
install_version('MyPackage', '>= 0.24', repos)  # Installs from ‘Dev’
```

and the `version` field would be extended to take a specification like the `Depends:` and `Imports:` fields in a `DESCRIPTION` file (and like those, multiple specs could be provided to be 'and'-ed together).

jimhester:
[...]

PR #1259, opened 2016-07-12
install_version() with multiple repositories

kenahoo:
Following up on my previous ticket #1107, this PR adds the ability for `install_version()` to search multiple CRAN[-like] repositories for a specific version of a package.  Currently, if it finds a package in an early repository but it's not the requested version, it will fail - this PR lets it continue searching the next repositories.

From the new docs I wrote for it:

> This function knows how to look in multiple CRAN-like package repositories, and in their \code{archive} directories, in order to find specific versions of the requested package.
> 
> The repositories are searched in the order specified by the \code{repos} argument.  This enables teams to maintain multiple in-house repositories with different policies - for instance, one repo for development snapshots and one for official releases.  A common setup would be to first search the official release repo, then the dev snapshot repo, then a public CRAN mirror.

Thanks.

### DIFF
--- a/R/install-version.r
+++ b/R/install-version.r
@@ -49,6 +49,12 @@ install_version <- function(package, version = NULL, repos = getOption("repos"),
     all(result)
   }
 
+  ## returns TRUE if 'pkg' is already installed and satisfies all version criteria 'criteria'
+  have <- function(pkg, criteria) {
+    v <- suppressWarnings(packageDescription(pkg, fields = "Version"))
+    !is.na(v) && satisfies(v, criteria)
+  }
+
   install_version_deps <- function(deps) {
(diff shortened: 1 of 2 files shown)


kenahoo:
I see this is failing the Travis & AppVeyor builds.  It seems to be a version thing on 'testthat', does it have something to with the change I made?

kenahoo:
Any thoughts on this?

jimhester:
I think this looks like a nice improvement, thank you for working on it. We will take a closer look at it nearer to the next devtools release with some more comments.

I apologize for the delayed response!

kenahoo:
Great, glad to hear it!  Thanks for taking a look.

kenahoo:
I just rebased this WRT master.

hadley:

Issue #1260, opened 2016-07-13
Non-interactive version of update_packages?


pkq:
Currently, calling `update_packages()` without providing a list of packages to update prompts the user for confirmation, meaning the function cannot be used non-interactively. Does it make sense to add an argument to allow this prompt to be overridden (e.g., `update_packages(override = TRUE)` )?

hadley:
This will be dealt with in remotes, particularly as we rethink how installation and updates should happen.

PR #1261, opened 2016-07-13
Allow update_packages() prompt to be overridden

pkq:
Following up on ticket #1260, this PR adds the ability to use `update_packages()` non-interactively by adding an `override` argument (i.e., `override = TRUE`). Currently, `calling update_packages()` without a list of packages prompts the user to confirm the updates - this PR allows the user to override this prompt.

FYI, I did read [guidelines for contributing](https://github.com/hadley/devtools/blob/master/CONTRIBUTING.md), as requested, but wasn't sure how or where to add a test for this.

Thanks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # devtools 1.12.0.9000
 
+* `update_packages()` allows for override of interactive prompt (#1260, @pkq).
+
 * Bugfix for installation of dependencies of dependencies (@jimhester, #1409).
 
 * `RCMD()`, `clean_source()`, `eval_clean()` and `evalq_clean()` have been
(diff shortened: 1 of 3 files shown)


jimhester:
I think it would be better to test if `pkgs` is `TRUE` and update all packages if so (without prompting).

pkq:
Is this what you had in mind?

pkq:
I've cleaned this pull request up a bit, PTAL. Any guidance regarding testing is appreciated. Thanks!

jimhester:
Thanks! Merged in aa5d4dabd7eca

pkq:
Thank you!

PR #1263, opened 2016-07-13
using authenticate to access Github package name. fixes #1262

### LINKED ISSUE #1262: Error resolving package name for private github repo
I'm running into some problems when installing and resolving dependencies for packages that imports/depends on packages in a private github repo.

``` r
> deps <- devtools::dev_package_deps(dependencies = TRUE)
Using GitHub PAT from envvar GITHUB_PAT
Using GitHub PAT from envvar GITHUB_PAT
Error in data.frame(package = package, installed = installed, available = available,  : 
  row names contain missing values
> 
```

The issue seems to be that we are not authenticating to Github when trying to resolve the package name. We then end up getting an error message when trying to construct a `data.frame` of named vectors with `NA` names.

I'm submitting a pull request for a solution where we:
1. Use the oath_token to authenticate before downloading the DESCRIPTION file.
2. Download the description from https://raw.githubusercontent.com/user/repo/... instead of https://github.com/user/repo/RAW/.... For some reason the latter approach doesn't seem to work.

eriknil:
The new version of devtools is unable to resolve remote package names for private github repositories.

Used same approach as in `remote_download.github_remote` for
authenticating, and changed the url used from
https://github.com/user/repo/RAW/... to
https://raw.githubusercontent.com/user/repo/...

This should fix #1262

### DIFF
--- a/R/install-github.r
+++ b/R/install-github.r
@@ -243,18 +243,27 @@ parse_git_repo <- function(path) {
 }
 
 #' @export
-remote_package_name.github_remote <- function(remote, url = "https://github.com", ...) {
+remote_package_name.github_remote <- function(remote, url = "https://raw.githubusercontent.com", ...) {
 
   tmp <- tempfile()
   path <- paste(c(
       remote$username,
       remote$repo,
-      "raw",
       remote$ref,
       remote$subdir,
       "DESCRIPTION"), collapse = "/")
 
-  req <- httr::GET(url, path = path, httr::write_disk(path = tmp))
+  if (!is.null(remote$auth_token)) {
+    auth <- httr::authenticate(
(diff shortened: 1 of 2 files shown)


R/install-github.r
```diff
@@ -302,3 +311,4 @@ download_github <- function(path, url, ...) {
   writeBin(httr::content(request, "raw"), path)
   path
 }
+
```
jimhester: Could you get rid of the trailing line here

eriknil:
of course

hadley:
@jimhester is this good to merge?

jalazawa:
I have the same issue

jimhester:
Thanks! Merged in a61d87fc995

Issue #879, opened 2015-07-15
install_git fails if branch argument is a tag


robinsonjj:
E.g.
```
> devtools::install_git('https://github.com/hadley/devtools', branch='v1.8.0')
Downloading git repo https://github.com/hadley/devtools
Error in .local(object, ...) : 'v1.8.0' did not match any branch
```

hadley:
This seems to be a git2r issue:

```R
library(git2r)

out <- tempfile()
r <- clone("https://github.com/hadley/devtools", out, progress = FALSE)
git2r::checkout(r, "v1.8.0")
```

Can you please follow up at https://github.com/ropensci/git2r/issues ?

ClaytonJY:
@robinsonjj posted this to `git2r` as https://github.com/ropensci/git2r/issues/176, which has been fixed quite recently, but I'm still getting a failure:

```r
> devtools::install_git("https://github.com/r-lib/httr.git", branch = "v1.3.1")
# Downloading git repo https://github.com/r-lib/httr.git
# Installation failed: 'head' is not an exported object from 'namespace:git2r'
```
(The second line used to be `Installation failed: 'v1.3.1' did not match any branch`.)

Can confirm that the checkout example posted on the `git2r` issue does work, which makes me think it may now be a `devtools` issue, though I'm far from certain. Can't see where `devtools` _or_ `git2r` refer to just `head`.

<details>
<summary>Session Info</summary>

```r
> devtools::session_info(c("devtools", "git2r"))
Session info ------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 setting  value                       
 version  R version 3.4.4 (2018-03-15)
 system   x86_64, linux-gnu           
 ui       RStudio (1.1.383)           
 language (EN)                        
 collate  en_US.UTF-8                 
 tz       America/New_York            
 date     2018-05-18                  

Packages ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 package    * version     date       source                         
 curl         3.2         2018-03-28 CRAN (R 3.4.4)                 
 devtools     1.13.5      2018-05-18 local                          
 digest       0.6.15      2018-01-28 CRAN (R 3.4.4)                 
 git2r        0.21.0.9002 2018-05-18 Github (ropensci/git2r@dcb3d91)
 graphics   * 3.4.4       2018-03-16 local                          
 grDevices  * 3.4.4       2018-03-16 local                          
[...]

PR #1264, opened 2016-07-14
fix install_git to allow installing from a specific tag

renejuan:
I think install_git does not allow installing from a tag as mentioned in the help. I saw a discussion about this in issue #879; maybe this will work.

### DIFF
--- a/R/install-git.r
+++ b/R/install-git.r
@@ -51,7 +51,13 @@ remote_download.git_remote <- function(x, quiet = FALSE) {
 
   if (!is.null(x$branch)) {
     r <- git2r::repository(bundle)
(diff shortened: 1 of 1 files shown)


hadley:
I'd prefer to leave this for now as this code is likely to be moving to a new home soon

malwinare:
@jimhester Since devtools is under going some changes and will be spilt into smaller packages, would it be maybe a good time to pick up this issue? 
We use `devtool::intall_git()` on a regular basis and it would be great if you could make installing from a tag possible, despite of still open issue about `git2r::checkout`. We would really appreciate it.  

Thank you in advance.

hadley:
We'll tackle this when we work on package installation in general.

PR #1268, opened 2016-07-18
Source single file gist without filename; fixes #1266

### LINKED ISSUE #1266: source_gist not working
Hi,

I am trying to use source_gist from the devtools package but I am encountering an error:

```
> library(devtools)
> source_gist("524eade46135f6348140")
Error in r_files[[which]] : invalid subscript type 'closure'
```

Thanks for any advice.

adamdsmith:
This fixes an issue in which an attempt to source a gist containing a single filename, but without specifying that filename, fails.  Fixes issue #1266.

### DIFF
--- a/R/run-source.r
+++ b/R/run-source.r
@@ -137,7 +137,7 @@ find_gist <- function(id, filename) {
     }
 
   } else {
-    if (length(r_files) = 1) {
+    if (length(r_files) == 1) {
       which <- 1
     } else {
       warning("Multiple R files in gist, using first.")


h8d5lo1:
The issue page hadley/devtools/issues/1266 seems to indicate this has been fixed, but I'm getting the same error as @Cauchy2. I've updated devtools and attempted to call gist_source() but to no avail.

```
> library(devtools)
> source_gist(9112634)
Error in r_files[[which]] : invalid subscript type 'closure'
```

I'm a bit of a novice with this stuff so apologies in advance if i'm missing something simple!

R/run-source.r
```diff
@@ -137,8 +137,8 @@ find_gist <- function(id, filename) {
     }
 
   } else {
-    if (length(r_files) > 1) {
-      warning("Multiple R files in gist, using first.")
+    if (length(r_files) >= 1) {
```
hadley: This is rather confusing logic. I think it would be better as:

```
if (length(r_files) == 1) {
   ...
} else {
   ...
}
```

Presumably zero files should also be an error.

adamdsmith:
Of course.  Thanks for your patience...

hadley:
Sorry, I forgot about your PR and ended up fixing by hand.

PR #1278, opened 2016-07-28
Add branch name to install message in install_git

rubak:
Output the name of the branch installed by `install_git()` to the user which may be helpful when browsing back through log.


hadley:
I'd rather leave this for now, as the package install code is likely to be moving to a new home

PR #1281, opened 2016-07-29
Handle case of uninstalled package being passed to session_info

jimhester:
Will now error with a more informative error message than previous (https://travis-ci.org/hadley/r4ds/builds/148333172#L4144)


hadley:
Could you please add a news bullet and rebase/merge?

PR #1283, opened 2016-08-01
Use Internal call rather than compiled code for the namespace registry

### LINKED ISSUE #1282: FR: Get rid of C code
This would allow building/upgrading devtools on Windows without Rtools and perhaps without `build_github_devtools()`.

It's only the `nsreg()` function which is called by `ns_registry()`; to me it looks like it does the same as `.Internal(getNamespaceRegistry())`. Getting the latter past the CRAN checks is possible with some eval-bquote trickery. This is a CRAN-able version of `.Internal()`:

```r
internal <- function(code) {
  Internal <- as.name(".Internal")
  code_sub <- substitute(code)

  eval(bquote(
    .(Internal)(.(code_sub))))
}
```

Putting `ns_registry()` into a separate package is also an option.

jimhester:
Fixes #1282


R/namespace-env.r
```diff
@@ -175,11 +175,11 @@ is_loaded <- function(pkg = ".") {
 
 
 # Returns the namespace registry
-#' @useDynLib devtools nsreg
 ns_registry <- function() {
-  .Call(nsreg)
+  (get(".Internal", envir = baseenv(), mode = "function"))(getNamespaceRegistry())
```
krlmlr: Would the following work without the globalVariables() call:

```r
internal <- (get(".Internal", envir = baseenv(), mode = "function"))
internal("getNamespaceRegistry"())
```

jimhester: No it still generates the  NOTE
```
ns_registry: no visible global function definition for
  ‘getNamespaceRegistry’
```

hadley:
That should simplify the install instructions in the readme, right?

jimhester:
Yeah that is true, want me to add that to this PR or will you modify it after merging?

hadley:
Can you add please? Might also be worth adding deprecation notice to `build_github_devtools()`

(But can you please verify it does actually work on windows)

jimhester:
Seems to be working fine on windows, and I updated the README as well.

R/build-github-devtools.r
```diff
@@ -32,6 +32,7 @@
 #' }
 #' @export
 build_github_devtools <- function(outfile = NULL) {
+  .Deprecated(msg = "`build_github_devtools()` is deprecated and will be removed in future versions of devtools")
```
hadley: I think the message would be more helpful if you told people what to do instead

jimhester: Ok, good idea! https://github.com/hadley/devtools/pull/1283/commits/eaa21c510c9c6cc3a344bfb847c7985bc03dfd8f#diff-16f1b7f2a08c02db5035892c22a81eeeR35

hadley:
Thanks!

PR #1285, opened 2016-08-02
Revdep e-mail improvements

krlmlr:
- new variables `my_version` and `you_cant_install`
- updated e-mail template, tested with the current DBI revdep check results

NEWS entry:

```
* Automated revdep check e-mails now can use the new `my_version` and
  `you_cant_install` variables. The e-mail template has been updated
  to use these variables (#1285, @krlmlr).
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # devtools 1.12.0.9000
 
+* Automated revdep check e-mails now can use the new `my_version` and
+  `you_cant_install` variables. The e-mail template has been updated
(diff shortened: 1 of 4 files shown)


krlmlr:
Use explicit version argument, version number at point of release unknown when revdep-checking.

krlmlr:
Change defaults to send only for problems.

krlmlr:
Changed defaults to send only for problems, and using explicit version argument. Tested with RSQLite.

PR #1286, opened 2016-08-03
add USE_RTOOLS: true to appveyor.yml

bborgesr:
this should fix the AppVeyor build error (I just had the same problem a couple of days ago and ended up here because of an unrelated bug, so decided to submit this since it's a very easy fix)


jimhester:
Appveyor was failing because of a devtools bug in the current release version, but `USE_RTOOLS:true` will be useful once #1283 is merged and devtools no longer has a `src/` directory, so thanks!

PR #1287, opened 2016-08-06
Take the best of both README templates

jennybc:
Unify the template for README.Rmd and README.md, so that both contain scaffold for

  * what does this package do?
  * how to install (if GitHub usage detected)
  * an example

I expanded the help. The use of whisker/mustache is very much "brute force" but it's simple to maintain.

The first commit should be taken, no matter what, as it fixes a general problem with `use_template()` & RStudio.

I'm happy to write tests (and have tested all 4 combinations of Rmd vs md * Github yes vs no), but I didn't see any others for this sort of thing.


hadley:
Looks great - thanks!

krlmlr:
This patch has been corrupted in master, working on restoring it.

krlmlr:
Luckily, it was only the master branch in my fork. Please ignore.

PR #1291, opened 2016-08-17
Avoid trying to move vignette sources twice

flying-sheep:
Sometimes multiple vignettes can share source files. This breaks the `file.remove` call which expects unique arguments.

---

the way i encountered this is that `tools::pkgVignettes` has two bugs:
1. it loops over all vignette engines (`tools::loadVignetteBuilder` results in a hardcoding sweave as last engine) and sets `engine` to the last one looped (=always sweave)
2. `tools:::find_vignette_product` is called with that engine and finds _all_ r scripts in the directory prefixed with that name.

this results in duplicated source entries if a vignette’s file name is a prefix of another.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -67,8 +67,10 @@
 
 * `test()` doesn't load helpers twice anymore (@krlmlr, #1256).
 
-* fix auto download method selection for `install_github()` on R 3.1 which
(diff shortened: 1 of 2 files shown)


hadley:

jimhester:
Thanks! Merged in 72b1926131

PR #1303, opened 2016-08-24
added remote_package_info

muschellij2:
Added `remote_package_info` for the ability to get information (if available) from the DESCRIPTION file other than just Version if DESCRIPTION file is used for parsing.  I added a generic for `remote_package_info`, so you may be able to remove `remote_package_name` as a generic if you want.


hadley:
I'd prefer to leave this for now, as this code is likely to be moving into another package in the near future.

muschellij2:
OK thanks - remotes or ghit by any chance?

hadley:
remotes

PR #1119, opened 2016-03-20
Add package spell checking

jeroenooms:
A quick way to find spelling errors in a package title, description and manual pages. More rigorous than CRAN's checks... you might find a few spelling errors in devtools :)

PR #1335, opened 2016-09-09
Add dictionary argument to spell_check

jeroen:
You can now use a custom language in `spell_check` for example:

``` r
spell_check(dict = "en_GB")
```

Also `hunspell` is now entirely self-contained (it has zero system dependencies) so I think we can just import it. Also added a feature that will automatically check documentation on release (as [suggested](https://github.com/hadley/devtools/pull/1119#issuecomment-200390768) by @hadley ).

Hopefully this will encourage users to eliminate typo's from documentation :)

### DIFF
--- a/R/release.r
+++ b/R/release.r
@@ -39,8 +39,10 @@
 #'   release it.
 #' @param args An optional character vector of additional command
 #'   line arguments to be passed to \code{R CMD build}.
+#' @param spelling language or dictionary file to spell check documentation.
+#' See \code{\link{spell_check}}. Set to \code{NULL} to skip spell checking.
 #' @export
-release <- function(pkg = ".", check = TRUE, args = NULL) {
+release <- function(pkg = ".", check = TRUE, args = NULL, spelling = "en_US") {
   pkg <- as.package(pkg)
   # Figure out if this is a new package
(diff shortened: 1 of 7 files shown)


jimhester:
@hadley may disagree, but I would prefer keeping this in `Suggests:`, while it is self-contained it still requires compilation and will be unused functionality for the vast majority of builds on the CIs, where we are trying to lighten the devtools dependency load.

jeroenooms:
Right, I don't disagree with that. Though it would be nice to run a full documentation spell check by default during CI builds and `release()`. It works really well :-)

hadley:
Let's leave it in suggests for now, and we can move into imports once we've extracted the stuff CI tools needs into a separate package.

jeroenooms:
OK it's back in suggests now.

R/release.r
```diff
@@ -60,6 +62,14 @@ release <- function(pkg = ".", check = TRUE, args = NULL) {
       return(invisible())
   }
 
+  if (length(spelling)) {
```
hadley: Can you please move this to `check-devtools.r`, following the template in `release_checks()`?

jimhester:
Thanks! Merged in 3af9ccab

Issue #1243, opened 2016-06-28
devtools::install and install_github do not pass on Additional_repositories on when installing deps?


cboettig:
If I have package A that depends on package B, which depends on package C; and B & C are in the listed `Additional_repositories` drat repository, it seems `devtools::install()` fails to find package C when trying to install B (and thus fails to install A & B).  A simple `install.packages("A", repo="my.drat.repo")` works fine.

For a concrete example:

``` r
install.packages("RNeXML", dep=TRUE, repos=c("https://cran.rstudio.com", "http://packages.ropensci.org"))
```

Works just fine.  But 

``` r
devtools::install_github("ropensci/RNeXML", deps=TRUE)
```

fails by failing to install the suggested dependency `rrdf` because `rrdflibs` is "not available", even though both are on the drat repo.

eddelbuettel:
Also, the discussion [in the remotes vignette](https://github.com/hadley/devtools/blob/master/vignettes/dependencies.Rmd) fails to mention `drat` use.  Given that CRAN supports it, and other packages use it, you may want to mention this (as a non-devtools alternative).

jimhester:
@cboettig I just opened a PR that should address your issue (as long as you specify your repos appropriately)

@eddelbuettel That vignette was written to explain the remotes feature of devtools, not as a general review of available CRAN alternatives. However if you would like `drat` to be mentioned in it I am sure we would be happy to merge a pull request with the change.

cboettig:
Thanks @jimhester , very nice!

eddelbuettel:
@jimhester could do if you think it fits -- I was mostly coming from the 'well remotes is not the only way to do this' angle of actually mentioning the fact that CRAN now formally supports Additional_repositories.  The remotes documentation, focused as it is, seems to imply it was the only game in town.

PR #1339, opened 2016-09-12
These changes address uses non-standard CRAN like repositories e.g. (drat)

jimhester:
This PR addresses https://github.com/hadley/devtools/issues/1243.

There were two issues, first the prior behavior of forcing `type = binary` was causing drat repositories without binary packages to fail, second the `repos` argument was not being passed onto `update()`.

With these changes the following correctly installs `rrdf` and `rrdflibs` as expected by @cboettig.

``` r
devtools::install_github("ropensci/RNeXML", dep=TRUE, repos = c("https://cran.rstudio.com", "http://packages.ropensci.org"))
```

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -18,10 +18,7 @@
 #' @param quiet If \code{TRUE}, suppress output.
 #' @param upgrade If \code{TRUE}, also upgrade any of out date dependencies.
 #' @param repos A character vector giving repositories to use.
-#' @param type Type of package to \code{update}.  If "both", will switch
-#'   automatically to "binary" to avoid interactive prompts during package
-#'   installation.
-#'
+#' @param type Type of package to \code{update}.
 #' @param object A \code{package_deps} object.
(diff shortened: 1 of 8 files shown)


R/deps.R
```diff
@@ -47,9 +44,6 @@
 #' }
 package_deps <- function(pkg, dependencies = NA, repos = getOption("repos"),
                          type = getOption("pkgType")) {
-  if (identical(type, "both")) {
```
hadley: How is `type = "both"` handled now?

jimhester: devtools always forces a type of `binary` https://github.com/hadley/devtools/pull/1339/files#diff-e7118181d76027ea30ccf5babc54f550L51, which means you never have a interactive dialog asking if you want to install.

This PR sets `option("install.packages.compile.from.source" = "never")`, which effectively means the same as forcing a type of "binary" if type = 'both', _unless_ there is no binary available (such as drat repositories), in which case it will compile the package.

hadley:
Thanks!

PR #1352, opened 2016-09-23
Make github_remote an exported function

connectedblue:
This PR proposes to make `github_remote` an exported function of the `devtools` package.

I find this function to be useful, together with `remote_download.github_remote` to download zipballs from a github site for processing locally (so not installing packages, just getting any project locally).

I'd rather be calling an exported function using `::` than an internal one which could change any time.

It does occur to me that the function I use (`download_github()`) might be a useful addition to `devtools`.  If you think so, I'm happy to close this PR and raise that instead for consideration?

rgds
Chris


hadley:
I'd rather not do this here - instead it should be a feature of the remotes package, which devtools will use in the future.

PR #1355, opened 2016-09-28
Added quiet argument to install_bitbucket

### LINKED ISSUE #1345: install_bitbucket() needs a default 'quiet' value
Hi,

Using the example on the `install_bitbucket` help page:

```
install_bitbucket("sulab/mygene.r@default")
```

Fails with:

```
Error in value[[3L]](cond) : argument "quiet" is missing, with no default
```

I think the problem is that the arguments of `install_bitbucket` are passed to `try_install_remote`, which requires a value for `quiet`:

```
install_bitbucket(repo, username, ref = "master", subdir = NULL, auth_user = NULL, 
   password = NULL, ...) 
-> install_remotes(remotes, ...) 
-> try_install_remote(..., quiet = quiet)
```

If we add `quiet` to install_bitbucket, the problem is solved: 

```
install_bitbucket("sulab/mygene.r@default", quiet = FALSE)
## works as expected
```

Adding a default `quiet` value to the `install_bitbucket` definition, as is the case in `install_github`, would fix this.

Best,

Tyler

plantarum:
Resolves #1345


hadley:
Can you please add a bullet to NEWS?

plantarum:
Apparently I can't. I'm not familiar with the continuous-integration checks, but it appears I've broken something to do with the code that tests the help files? I don't see anything obviously connected to this in the three files I've changed in this pull request?

jimhester:
@plantarum The failure with R-devel is an unrelated issue, you can safely ignore it.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # devtools 1.12.0.9000
 
+* added a `quiet` argument to `install_bitbucket()`, with a default value
+  of `FALSE` (fixes issue #1345, @plantarum).
```
hadley: Need another newline here

jimhester:
Thanks! Merged in 3dbef06d6c7

PR #1365, opened 2016-10-05
Check and install bioconductor

jimhester:
And install it if the user requests to do so.

### DIFF
--- a/R/deps.R
+++ b/R/deps.R
@@ -23,6 +23,8 @@
 #'   installation.
 #'
 #' @param object A \code{package_deps} object.
+#' @param bioconductor Install Bioconductor dependencies if the package has a
+#' BiocViews field in the DESCRIPTION.
 #' @param ... Additional arguments passed to \code{install_packages}.
 #'
 #' @return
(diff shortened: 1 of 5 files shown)


hadley:
I hate sourcing a random file, but it's clearly the best solution 😄

R/deps.R
```diff
@@ -109,7 +109,7 @@ dev_package_deps <- function(pkg = ".", dependencies = NA,
   deps <- unlist(lapply(parsed, `[[`, "name"), use.names = FALSE)
 
   if (is_bioconductor(pkg)) {
-    check_suggested("BiocInstaller")
+    check_bioconductor()
```
hadley: It's slightly annoying that there's an interactive prompt here, because it might get triggered arbitrarily later after starting `revdep_check()`. But I don't know how to avoid it

hadley: VOMITING IN MY MOUTH

hadley: Is it necessary to stop here on failure? There appear to be CRAN packages for which `is_bioconductor()` is true

jimhester: They are likely packages that were rejected for Bioconductor for whatever reason and the authors did not remove the `BiocViews` field.

We could add a bioconductor argument to `dev_package_deps()` which would allow you to turn off the check like we have in `check_cran()`.

hadley: Would you mind adding that? It'd help with my revdep checks.

jimhester: Sure, done in 7d7e6af

hadley:
Thanks!

PR #1366, opened 2016-10-05
Revdep check improvements

krlmlr:
@hadley: Please review and merge this before modifying revdep check code. See individual PRs for more fine-grained changes. Your most recent changes are included. My other PRs seem unrelated.

Closes #1285.

Closes #1300.

Closes #1319.

Closes #1336 (which itself contains #1320, #1321, #1322).

Closes #1338 (which contains #1337).

With NEWS.

Sequence of changes:

- Revert c3457d2a0506
- Merge all PR branches
- Manually reapply c3457d2a0506 (see 9245481)
- Improve handling of skip argument (ccf35e6)

I see how `skip` is different from `ignore`, perhaps `ignore` should be removed?

PR #1368, opened 2016-10-06
Don't pass skip to check_cran()

krlmlr:
Follow-up to #1366.

@jcheng5: Can you please try:

```r
devtools::install_github("hadley/devtools#1368")
```

@hadley: I broke revdep check, this PR fixes it.


hadley:
Thanks!

PR #1371, opened 2016-10-06
Run R CMD check on r-hub

gaborcsardi:
First shot at it.


R/r-hub.R
```diff
@@ -0,0 +1,67 @@
+
+#' Build and check packag on r-hub
```
jimhester: Missing an e in package.

R/r-hub.R
```diff
@@ -0,0 +1,67 @@
...
+                       interactive = FALSE, ...) {
+
+  pkg <- as.package(pkg)
+
+  built_path <- build(pkg, tempdir(), quiet = !interactive, ...)
+  on.exit(unlink(built_path), add = TRUE)
+
+  status <- rhub::check(
+    built_path,
+    email = maintainer(pkg)$email,
```
jimhester: Maybe we should allow setting the email as an extra parameter? I know there have been a handful of times I wanted to check a package on win-builder I was not the normal maintainer for.

It also would be handy for checking PR before sending them.

DESCRIPTION
```diff
@@ -26,7 +26,8 @@ Imports:
     jsonlite,
     stats,
     git2r (>= 0.11.0),
-    withr
+    withr,
+    rhub
```
jimhester: Don't know how heavy-weight rhub is, but maybe this should be in `Suggests()` with a `devtools::check_suggested()`.

R/r-hub.R
gaborcsardi: Thanks! My son jumped on the 'e' key and it is half-broken now. :)

R/r-hub.R
gaborcsardi: I thought about that, and `rhub::check()` allows that. But `build_win()` does not, so I went with a similar API. But I can easily add this.

DESCRIPTION
gaborcsardi: Good point, I will do this, too. It is pretty lightweight, except for packages that are only needed by devtools: `httr` and `curl`. But if `testthat` is not required, then `rhub` should not be, either.

gaborcsardi:
TODO list:

- [ ] add `email` argument
- [ ] make `rhub` package optional

gaborcsardi:
Thanks for the suggestions, all should be fixed now.

hadley:

DESCRIPTION
```diff
@@ -26,7 +26,7 @@ Imports:
     jsonlite,
     stats,
     git2r (>= 0.11.0),
-    withr
+    withr,
```
hadley: Trailing comma?

R/r-hub.R
```diff
@@ -0,0 +1,73 @@
+
```
hadley: Leading nl

jimhester:
I merged this manually at https://github.com/hadley/devtools/commit/9a24447aad6ffe45c7143759ef510b0161c3c230, incoporating Hadley's comments.

PR #1312, opened 2016-08-30
Add encoding support

hansharhoff:
This pull request an encoding argument to the  test_dir call in test(). This is related to an issue #1306 
A separate pull request has been sent to hadley/testthat (https://github.com/hadley/testthat/pull/518)

PR #1378, opened 2016-10-19
fix load_all in windows with UTF8 codes

shrektan:
If the source file contains nonASCII chars, like Chinese Chars, `devtools::load_all()` would fail with error as below, but `Build & Reload` the package works fine.

I think it's due to not explicitly setting the `encoding` param in `readLines` of `source_one`, see https://github.com/hadley/devtools/blob/master/R/source.r#L21
### The error message sample

```
Loading package_abc
Error in parse(text = lines, n = -1, srcfile = srcfile) :
  invalid multibyte character in parser at line 12
In addition: Warning messages:
1: In grepl("\n", lines, fixed = TRUE) :
  input string 12 is invalid in this locale
2: In grepl("\n", lines, fixed = TRUE) :
  input string 14 is invalid in this locale
3: In grepl("\n", lines, fixed = TRUE) :
  input string 15 is invalid in this locale
```
### Notes
1. add `encoding = "UTF-8"` in `readLines()` to fix the reading from the file
2. convert the lines to `native` encoding, so the printing of the function displays correctly

### DIFF
--- a/R/source.r
+++ b/R/source.r
@@ -1,4 +1,4 @@
-source_many <- function(files, envir = parent.frame()) {
+source_many <- function(files, envir = parent.frame(), fileEncoding = "UTF-8") {
   stopifnot(is.character(files))
   stopifnot(is.environment(envir))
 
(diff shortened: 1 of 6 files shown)


hansharhoff:
I believe this is closely related to issue #1312

R/source.r
```diff
@@ -18,7 +18,8 @@ source_one <- function(file, envir = parent.frame()) {
   stopifnot(file.exists(file))
   stopifnot(is.environment(envir))
 
-  lines <- readLines(file, warn = FALSE)
+  lines <- readLines(file, warn = FALSE, encoding = "UTF-8")
+  lines <- enc2native(lines)
```
hadley: Are you sure this line is necessary?

shrektan: I'm pretty sure this is necessary, because I just reproduced this to confirm.  

1. Let's say we are working on a windows platform, whose native encoding is not `utf-8`,
1. If we have a function (e.g., `futf8()`) that contains nonASCII chars in some package ,
1. Without `lines <- enc2native(lines)`, `devtools::load_all()` performs good and the function `futf8()` works well,
1. But, if you `print(futf8)`, you will find mess in the console,
1. It will be solved by adding this line.

### without the line

image

### with this line

image

shrektan:
@hadley Oh, just notice that the `load_all()` has been moved to https://github.com/r-pkgs/pkgload 

However, I just added support for reading encoding from DESCRIPTION as well as a unit test.

shrektan:
Close because related code has been moved to https://github.com/r-pkgs/pkgload

PR #1382, opened 2016-10-28
Debug on 'install-github.r' code

### LINKED ISSUE #1381: `install_github` fails when a private github repository is specified in the remote dependency.
#### [Description]
- `install_github` fails when a private GitHub repository is specified in the remote dependency.
  - The remote dependency is described in `DESCRIPTION` file as following
    ```R
    Remotes: 
        github::PRIVATE_ORG/PRIVATE_PKG
    ```
#### [Steps to Reproduce]
- This issue is related to a private github repository, try following steps with your own private one. (Be prepared with the `Personal Access Token (PAT)`)
- I'll give the instruction with the assumption that you have two private repositories named `banana` and `monkey`, and your GitHub username is `zookeeper`.
1. Put `Remotes: github::zookeeper/banana` in the `DESCRIPTION` file in the `monkey` R package.
2. Try `devtools::install_github("zookeeper/monkey")`

(Expected)
- First install `zookeeper/banana` and then `zookeeper/monkey`

(Actual)
- Fail to install `zookeeper/banana`
#### [Screenshots]
- This is an actual case of mine
- My private R package `JediETL` has a remote dependencies which contains several private GitHub repositories.

image

#### [Related Issues]
- hadley/devtools#1262

#### [Related Pull Requests]
- hadley/devtools#1263

dy-kim:
## 
### The philosophy on this update
- Use `github_GET` rather than `httr::GET` directly when accessing `GitHub API`
  - `github_GET` is well abstracted
    - It aptly handle authentication and decode the response of `GitHub API`
  - Otherwise, using `httr::GET` directly can not systematically prevent omitting authentication
    - Present `remote_package_name.github_remote` show the exact problem mentioned above
    - hadley/devtools#1263 fix this with `if (!is.null(remote$auth_token)) { ... }`, but it still uses `httr::GET` directly
- Refer an official `GitHub API` documentation
  - For example, `remote_package_name.github_remote` includes the process of downloading a raw content from `GitHub`
    - `GitHub` provides [`get-contents API`](https://developer.github.com/v3/repos/contents/#get-contents) and this API can meet the case
    - So let's try to make use of this reliable tool rather than a custom approach
- Make the modified codes more readable, especially in the perspective of,
  - "Making functions smaller. "
  - "Don't mix different levels of abstractions. "
### Changes

(What has actually changed in the code)
- `R/install-git.r`
  - Modified: `remote_package_name.github_remote()`
    - Abstract the codes for downloading a `DESCRIPTION` file from `GitHub` as a new function `github_contents` ~~`get_contents_using_github_api`~~
      - In this design, we can reuse `get_contents_using_github_api` to get any contents in given `remote` object (not only just `DESCRIPTION`)
    - Remove argument `url` (It becomes dispensable so deprecated)
  - Added: `github_contents()` ~~`get_contents_using_github_api()`~~
    - Directly wraps [`'Get contents' GitHub API`](https://developer.github.com/v3/repos/contents/#get-contents)
      - This API returns the download URL of a content (in `download_url` field)
    - To download a raw content it uses a new function `download_file_using_wget`
  - ~~Added: `download_file_using_wget`~~
    - ~~Wrapper of `utils::download.file` with proper and useful options~~
    - ~~When a download fails, simply run the same code again, then it will continue not start from the beginning~~
      - ~~The `extra = "--continue"` option for `utils::download.file()` makes this possible~~
  - Modified: `remote_sha.github_remote`
    - `remote_sha.github_remote` is not working when `remote` is a private
      - It seems that `git2r::remote_ls` is not working for private `GitHub` repositories
      - This issue cause unnecessary re-installation for the already instead private R packages
    - Directly wraps [`'Get the SHA-1 of a commit reference' GitHub API`](https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference)
      - Access to this API using `github_GET`
      - Previous if-clause in line 270 ~ 271 has been abstracted to `has_sha.github_remote()`

### DIFF
--- a/R/install-github.r
+++ b/R/install-github.r
@@ -253,17 +253,6 @@ remote_package_name.github_remote <- function(remote, ...) {
   read_dcf(tmp)$Package
 }
 
-download_file_using_wget <- function(url, destfile) {
-  utils::download.file(
-    url = url,
-    destfile = destfile,
-    cacheOK = TRUE,
-    method = "wget",
-    quiet = TRUE,
-    extra = "--continue"
-  )
-}
-
 #' @export
 #' @references https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference
 remote_sha.github_remote <- function(remote, ...) {
--- a/R/install-github.r
+++ b/R/install-github.r
@@ -243,48 +243,67 @@ parse_git_repo <- function(path) {
 }
 
 #' @export
-remote_package_name.github_remote <- function(remote, url = "https://github.com", ...) {
-
+remote_package_name.github_remote <- function(remote, ...) {
   tmp <- tempfile()
-  path <- paste(c(
-      remote$username,
-      remote$repo,
-      "raw",
-      remote$ref,
-      remote$subdir,
-      "DESCRIPTION"), collapse = "/")
-
-  req <- httr::GET(url, path = path, httr::write_disk(path = tmp))
-
-  if (httr::status_code(req) >= 400) {
-    return(NA_character_)
-  }
+  on.exit(unlink(tmp))
+  get_contents_using_github_api(remote,
+                                content_path = "DESCRIPTION",
+                                path_to_save = tmp)
 
   read_dcf(tmp)$Package
 }
 
+#' @references https://developer.github.com/v3/repos/contents/#get-contents
+get_contents_using_github_api <-
+  function(remote, content_path, path_to_save) {
+    owner <- remote$username
+    repo <- remote$repo
+    target_path <-
+      file.path("repos", owner, repo, "contents", content_path)
+    response <- github_GET(path = target_path)
+
+    download_file_using_wget(url = response$download_url,
+                             destfile = path_to_save)
+  }
+
+download_file_using_wget <- function(url, destfile) {
+  utils::download.file(
+    url = url,
+    destfile = destfile,
+    cacheOK = TRUE,
+    method = "wget",
+    quiet = TRUE,
+    extra = "--continue"
+  )
+}
+
 #' @export
-remote_sha.github_remote <- function(remote, url = "https://github.com", ...) {
-  # If the remote ref is the same as the sha it is a pinned commit so just
-  # return that.
-  if (!is.null(remote$ref) && !is.null(remote$sha) &&
-    grepl(paste0("^", remote$ref), remote$sha)) {
+#' @references https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference
+remote_sha.github_remote <- function(remote, ...) {
+  if (has_sha.github_remote(remote)) {
     return(remote$sha)
   }
 
-  tryCatch({
-    res <- git2r::remote_ls(
-      paste0(url, "/", remote$username, "/", remote$repo, ".git"),
-      ...)
+  owner <- remote$username
+  repo <- remote$repo
+  ref <- remote$ref
+  target_path <- file.path("repos", owner, repo, "commits", ref)
 
-    found <- grep(pattern = paste0("/", remote$ref), x = names(res))
-
-    if (length(found) == 0) {
+  tryCatch(
+    expr = {
+      response <- github_GET(path = target_path)
+      return(response$sha)
+    },
+    error = function(e) {
       return(NA_character_)
     }
+  )
+}
 
-    unname(res[found[1]])
-  }, error = function(e) NA_character_)
+has_sha.github_remote <- function(remote) {
+  return(!is.null(remote$ref) &&
+           !is.null(remote$sha) &&
+           grepl(paste0("^", remote$ref), remote$sha))
 }
 
 #' @export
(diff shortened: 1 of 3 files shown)


R/install-github.r
```diff
@@ -243,48 +243,67 @@ parse_git_repo <- function(path) {
...
-  tryCatch({
-    res <- git2r::remote_ls(
-      paste0(url, "/", remote$username, "/", remote$repo, ".git"),
-      ...)
+download_file_using_wget <- function(url, destfile) {
+  utils::download.file(
+    url = url,
+    destfile = destfile,
+    cacheOK = TRUE,
+    method = "wget",
```
jimhester: `method = 'wget'`and `method = 'curl'` require that those executables be installed on the users system and on the `PATH`. This is not the case on windows unless the user has Rtools installed and the Rtools directory on the path. We should just be calling `utils::download.file` without the method set to respect the users choice of preferred download method.

If you personally like being able to continue downloads you can set `options(download.method.file = "wget", download.method.file.extra = "--continue')` in your `.RProfile` and have this happen for all calls of `download.file()`.

R/install-github.r
```diff
@@ -243,48 +243,67 @@ parse_git_repo <- function(path) {
...
   tmp <- tempfile()
-  path <- paste(c(
-      remote$username,
-      remote$repo,
-      "raw",
-      remote$ref,
-      remote$subdir,
-      "DESCRIPTION"), collapse = "/")
-
-  req <- httr::GET(url, path = path, httr::write_disk(path = tmp))
```
jimhester: This was deliberately done using a regular http request rather than using the GitHub API because we were regularly hitting the unauthenticated GitHub API rate limits on Travis. That problem has since been ameliorated, so this is probably a reasonable change...

R/install-github.r
```diff
@@ -243,48 +243,67 @@ parse_git_repo <- function(path) {
...
+  function(remote, content_path, path_to_save) {
+    owner <- remote$username
+    repo <- remote$repo
+    target_path <-
+      file.path("repos", owner, repo, "contents", content_path)
+    response <- github_GET(path = target_path)
+
+    download_file_using_wget(url = response$download_url,
+                             destfile = path_to_save)
   }
```
jimhester: This function should be in `R/github.R` and should be named `github_contents()`. It also needs to pass along the ref parameter from the remote, the code as written assumes retrieval from the HEAD of the repository (see the [ref parameter](https://developer.github.com/v3/repos/contents/#parameters)).

R/install-github.r
```diff
@@ -243,48 +243,67 @@ parse_git_repo <- function(path) {
...
+    method = "wget",
+    quiet = TRUE,
+    extra = "--continue"
+  )
+}
 
-    found <- grep(pattern = paste0("/", remote$ref), x = names(res))
+#' @export
+#' @references https://developer.github.com/v3/repos/commits/#get-the-sha-1-of-a-commit-reference
+remote_sha.github_remote <- function(remote, ...) {
```
jimhester: The code in devtools predates the introduction of this endpoint.

However https://developer.github.com/changes/2016-02-24-commit-reference-sha-api/ mentions a couple of things this PR does not do.

1. The Accept header needs to be set to `application/vnd.github.VERSION.sha`.
2. ETags should be used with the local SHA, the response will return a 304 Not Modified if the SHA is the same. ie.
```bash
curl "https://api.github.com/repos/Homebrew/homebrew/commits/master" \
  -H "Accept: application/vnd.github.chitauri-preview+sha" \
  -H "If-None-Match: \"814412cfbd631109df337e16c807207e78c0d24e\""
```

Using 2. prevents this call from decrementing the rate limit and reduces load on GitHub's servers.

dy-kim: I've checked the references again. I missed some important points as you mentioned. I think the first method would be appropriate. The latter one requires ` local SHA` which is not considered in the present `remote_sha.github_remote`. Moreover, the role of `remote_sha.github_remote` should be simply returning `remote SHA` and  doing something more(compare and return logical info).

dy-kim: So I've tried the first method, by passing the extra argument `httr::add_headers(Accept = "application/vnd.github.VERSION.sha")` to `github_GET`. In this case the line `github_response(req)` inside of the definition of `github_GET` raise an issue and I've documented it ([hadly/devtools#1384](https://github.com/hadley/devtools/issues/1384)).

dy-kim: To sum these up, I'd like to approach with the first suggested method, and it carries an issue rooted from the restriction in the `github_GET`, but I don't want to avoid this problem by not using `github_GET`. Instead, I think it is better to extend `github_GET` together in this time.

dy-kim:
@jimhester Recently, I've updated what you suggested. Please review the changes.

deanchen:
@jimhester we're maintaining our own branch of devtools with the PR since reference private repos doesn't work without this. would appreciate it if you can take some time to review and merge this in.

hadley:

dy-kim:
@hadley Yes, I've heard those story from @jimhester at the latest `useR` conference. I would review the up-to-date `remotes` and make issues and PRs if needed.

PR #1388, opened 2016-11-03
Build on 3.0.3

jimhester:


hadley:
Still fails due to lack of `trimws()`

jimhester:
It is working now, but the check is failing because suggested packages are not available

krlmlr:
I'm switching to using `remotes` in r-appveyor, perhaps r-travis could do the same?

hadley:
Should we take another pass at this, or go for R 3.1?

hadley:
Superseded by #1560

PR #1397, opened 2016-11-07
Fully install the package before revdep checking

yihui:
During revdep checking, I found some packages actually had tests that depended on `system.file('docs', package)`. If the package is not fully installed, these tests will fail.


hadley:
Thanks - this should be resolved in revdepcheck (which does a much better job installation generally)

PR #1400, opened 2016-11-21
fix up casing with RStudio

kevinushey:


jimhester:
Thanks, merged in 15b5410c

PR #1405, opened 2016-11-29
Fix typos in ?revdep_check

mvkorpel:
Simple typo fix. Thanks.


hadley:
Thanks! Unfortunately `revdep_check()` is going away in favour of the new revdepcheck package so this is no longer relevant. Sorry for taking so long to get back to you!

PR #1407, opened 2016-11-30
Fix reported list of dependencies to be installed

mvkorpel:
List dependencies to be installed, not packages to be checked


hadley:
Thanks - this is also fixed in the revdepcheck package

PR #1413, opened 2016-12-07
Documentation:  typo fix in doc of use_package()

wamserma:
"Suggests" was given twice as possible value for the type parameter.


jimhester:
Thanks! Merged in a5ad54c4290d6

PR #1430, opened 2017-01-02
BUG FIX: print.spellcheck(x) gave error for length(x) == 0

HenrikBengtsson:
Example of error:

```r
> library(devtools)
> structure(list(), class = "spellcheck")
Error in sprintf(fmt, "  WORD") : 
  invalid format '%--Inf'; use format %s for character objects
In addition: Warning message:
In max(nchar(words)) : no non-missing arguments to max; returning -Inf
```


hadley:
Thanks!

PR #1441, opened 2017-01-14
Replace httr download with download.file

brada4:
httr on Windows does not inherit 'internal' proxy authentication nor initializes wininet.dll authentication dialog. This change should sharply rise corporate windows user :::download() success rates.

### DIFF
--- a/R/utils.r
+++ b/R/utils.r
@@ -154,10 +154,11 @@ first_upper <- function(x) {
 }
 
 download <- function(path, url, ...) {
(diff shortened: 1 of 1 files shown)


hadley:
We can't simply make that change since it breaks all the error handling.  Regardless, this code will be moving to remotes where we will consider other approaches.

gaborcsardi:
I believe that `remotes` handles proxy configuration better on windows. Please open an issue there if you have problems.

Issue #944, opened 2015-10-02
I don't understand how to use Remotes with travis


jennybc:
I'm trying to understand how to express that a package depends on the development/github version of another package. Such the right things happen for a user who does `devtools::install_github("jennybc/whatever")` and that it build successfully on travis for me.

I hoped I could specify a non-CRAN version for the dependency in DESCRIPTION and add the new `Remotes` field. But that leads to a failure on travis. Here's the possibly most informative part of the log:

```
[0K$ Rscript -e 'options(repos = "http://cran.rstudio.com"); tryCatch({   deps <- devtools::install_deps(dependencies = TRUE) }, error = function(e) {   message(e);   q(status=1) }); if (!all(deps %in% installed.packages())) {  message("missing: ", paste(setdiff(deps, installed.packages()), collapse=", "));  q(status = 1, save = "no") }'
Malformed remote specification 'hadley/readr'travis_time:end:03ddefb0:start=1443816228668380316,finish=1443816228819430449,duration=151050133
[0K
[31;1mThe command "Rscript -e 'options(repos = "http://cran.rstudio.com"); tryCatch({   deps <- devtools::install_deps(dependencies = TRUE) }, error = function(e) {   message(e);   q(status=1) }); if (!all(deps %in% installed.packages())) {  message("missing: ", paste(setdiff(deps, installed.packages()), collapse=", "));  q(status = 1, save = "no") }'" failed and exited with 1 during .[0m

Your build has been stopped.
```

Note this: `Malformed remote specification 'hadley/readr'`.

I get a successful build when I remove `Remotes` from DESCRIPTION and instead put this in `.travis.yml`:

```
r_github_packages:
- hadley/readr
```

Am I just missing how the various moving parts in DESCRIPTION and `.travis.yml` interact?

Here's a little test package [jennybc/jane](https://github.com/jennybc/jane) and my exploration of failure on travis [jennybc/jane](https://travis-ci.org/jennybc/jane/builds).

cc @jimhester

jimhester:
This is unfortunately currently broken until https://github.com/hadley/devtools/pull/936 is merged.

If you do
```yaml
r_github_packages:
  - hadley/devtools#936
```
It should work without without having to specify `hadley/readr` in `.travis.yml`.

jennybc:
[...]

PR #1455, opened 2017-02-07
remind users that REMOTES still need IMPORT and friend entries

maxheld83:
The `REMOTES` field *feels* so similar to the `IMPORTS` and friends fields in the `DESCRIPTION` that I thought a package would have to be called in only one of the two.
@jimhester reminded me in https://github.com/hadley/devtools/issues/944#issuecomment-278013085 that ain't so.
Hope this it clearer for other newbies.

### DIFF
--- a/vignettes/dependencies.Rmd
+++ b/vignettes/dependencies.Rmd
@@ -52,6 +52,11 @@ A type of 'github' can be specified, but is not required
 Remotes: github::hadley/ggplot2
 ```
 
+Remember that the `Remotes:` field only specifies the *source* of a package, but it does not actually make it available to the R session.
+You still also have to call the package in `Imports:`, `Suggests:` `Depends:` or  `LinkingTo:`.
+In the above example, where the *package name* and the *repo name* are identical, you still need `Imports: ggplot2` in your `DESCRIPTION`.
+
+
 ### Other sources
 
 All of the currently supported install sources are available, see the 'See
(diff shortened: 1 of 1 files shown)


vignettes/dependencies.Rmd
```diff
@@ -52,6 +52,11 @@ A type of 'github' can be specified, but is not required
 Remotes: github::hadley/ggplot2
 ```
 
+Remember that the `Remotes:` field only specifies the *source* of a package, but it does not actually make it available to the R session.
+You still also have to call the package in `Imports:`, `Suggests:` `Depends:` or  `LinkingTo:`.
+In the above example, where the *package name* and the *repo name* are identical, you still need `Imports: ggplot2` in your `DESCRIPTION`.
+
```
jimhester: I think something like the following would be clearer. The distinction between package name and repo name while correct is fairly rare and just muddies the waters.

`Remotes:` specifies the *source* of a development package, the package still needs to be listed in `Imports:`, `Suggests:` `Depends:` or  `LinkingTo:`. e.g. `Imports: ggplot2` in the above example.

vignettes/dependencies.Rmd
```diff
@@ -52,6 +52,10 @@ A type of 'github' can be specified, but is not required
 Remotes: github::hadley/ggplot2
 ```
 
+Remember that `Remotes:` specifies the *source* of a development package, so the package still needs to be listed in `Imports:`, `Suggests:` `Depends:` or `LinkingTo:`.
+For example, the above `DESCRIPTION` would also require an `Imports: ggplot2` line for `ggplot2` to actually be imported.
```
hadley: That's not really what "imports" does though.

hadley:
I think this is already covered in the second paragraph, so if more clarification is needed, it should go there.

PR #1463, opened 2017-02-20
Match depends correctly

thomasp85:
This change ensures that only exact matches in imports will get ignored. Currently adding e.g. `plyr` to a package that already imports `dplyr` will silently fail. This will happen for all packages that are substrings of packages already ignored.

The proposed change adds regex "fences" around the name of the package to import to circumvent this...

### DIFF
--- a/R/infrastructure.R
+++ b/R/infrastructure.R
@@ -319,7 +319,7 @@ add_desc_package <- function(pkg = ".", field, name) {
     new <- name
     changed <- TRUE
   } else {
-    if (!grepl(name, old)) {
+    if (!grepl(paste0('\\W', name, '(,|$|\\s)'), old)) {
       new <- paste0(old, ",\n    ", name)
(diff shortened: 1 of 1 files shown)


jimhester:
The proper way to do this is to use [word boundary anchors](http://www.regular-expressions.info/wordboundaries.html) around the package name.

``` r
re <- "\\btest\\b"
grepl(re, "test")
#> [1] TRUE
grepl(re, "ttest")
#> [1] FALSE
grepl(re, "testt")
#> [1] FALSE
grepl(re, " test")
#> [1] TRUE
grepl(re, " test ")
#> [1] TRUE
grepl(re, " test, ")
#> [1] TRUE
```

thomasp85:
Oh, didn't know that one... Much cleaner

jimhester:
Thanks! Merged in f7d4c020

PR #1475, opened 2017-03-14
Check for hunspell, give helpful error if missing

jimvine:
Per guidance in http://r-pkgs.had.co.nz/description.html


jimhester:
We have a helper in devtools to check for suggested packages. This can be just

`check_suggested("hunspell")`

jimhester:
Thanks! Merged in c7a71dba9

PR #1486, opened 2017-03-30
Check for suggested package (testthat) before use.

jimvine:
Inspired by recent Dirk Eddelbuettel post on suggests != depends I checked my packages to see whether I was checking for presence of suggested packages before using them. Spotted the absence of a check in the testthat setup generated by devtools.

http://dirk.eddelbuettel.com/blog/2017/03/22/#suggests_is_not_depends

I have no strong opinions on whether this is necessary / appropriate for tests or, if it is, whether I have suggested the best approach.

### DIFF
--- a/inst/templates/testthat.R
+++ b/inst/templates/testthat.R
@@ -1,4 +1,10 @@
-library(testthat)
-library({{{ name }}})
+if (requireNamespace("testthat", quietly = TRUE)) {
+  library(testthat)
+  library({{{ name }}})
 
-test_check("{{{ name }}}")
+  test_check("{{{ name }}}")
+} else {
+    warning("Tests not run. testthat package not present.", call. = FALSE)
+}
+
+


gaborcsardi:
I don't think there are any situations in which this check is useful.

If you want to run the tests, then a warning about `testthat` being unavailable is hardly what you want, no?

If you don't want to run the tests, then just don't run them. :)

jimvine:
Thanks for looking at this, @gaborcsardi. If you want to run the tests but `testthat` **is** unavailable then what do you think would be more useful than a warning? Should it be a `stop()` error condition instead?

I agree that not running the tests is also fine, which is presumably why devtools sets `testthat` as `Suggests` rather than `Depends`.

gaborcsardi:
>  If you want to run the tests but testthat is unavailable then what do you think would be more useful than a warning? Should it be a stop() error condition instead?

Yes, an error is more appropriate I think. Which you'll already get if testthat is not available, without your check.

> I agree that not running the tests is also fine, which is presumably why devtools sets testthat as Suggests rather than Depends.

What I mean is that not running the tests is fine if *you* don't want to run the tests.

`devtools` sets `testthat` as `Suggests`, because `testthat` is *not* needed for *using* the package. It is only needed if you want to run the tests. But then it is needed and not optional.

gaborcsardi:
In other words, I don't think it ever happens that you want to run the tests if `testthat` is available, but just want to ignore them if not. What would be this good for?

hadley:
Yeah, I also don't think this is needed.

Issue #1491, opened 2017-04-06
Bug in `install()` - Wrong argument passing in `pkgbuild::rcmd_build_tools()` usage


dy-kim:
### Problem
- `opts` object became obsolete since the commit [`Integrate callr and pkgbuild`](https://github.com/hadley/devtools/commit/2e3c4b692e3aecd7a4fd746e3000db1f4aa706bd#diff-b4729bd8d077a10ed051487dfd8e39eaL151)

### To do
- Change `arg` to `opts` inside the [`pkgbuild::rcmd_build_tools()` usage](https://github.com/hadley/devtools/blob/master/R/install.r#L158)
- I'll add this fixation to the PR #1481

PR #1493, opened 2017-04-10
No shQuote for new pkgbuild/callr

gaborcsardi:
New callr uses processx, so don't need to quote
paths any more. (It is actually a bug to quote
them.)

This also requires pinning pkgbuild to HEAD,
unfortunately. This should be temporary.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -50,6 +50,7 @@ Suggests:
 Remotes:
   hadley/testthat,
   r-pkgs/pkgload,
+  r-pkgs/pkgbuild@HEAD,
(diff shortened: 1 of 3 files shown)


gaborcsardi:
Btw. here:
https://github.com/hadley/devtools/blob/d8ab190cd05e4bd2a3688a1eab3324109060b111/R/install.r#L158
`args` should be `opts`, no? Otherwise `opts` is not used at all...

jimhester:
Yep you are right, I made the change

dy-kim:
I'll close the precedent but now duplicated issue #1491

PR #1495, opened 2017-04-13
Escape special characters in GitHub ref

yutannihilation:
Currently, `install_github()` fails with branches that contain special characters. A reprex is here:

``` r
devtools::install_github("yutannihilation/devtools@escape-#")
#> Using GitHub PAT from envvar GITHUB_PAT
#> Downloading GitHub repo yutannihilation/devtools@escape-#
#> from URL https://api.github.com/repos/yutannihilation/devtools/zipball/escape-#
#> condition in stop(github_error(request)): 404: Not Found
#>  (404)
```

The URL should be `https://api.github.com/repos/yutannihilation/devtools/zipball/escape-%23`, in which the special character `#` is escaped to `%23`. This PR fixes the issue by escaping it with `curl::curl_escape()`.

### DIFF
--- a/R/install-github.r
+++ b/R/install-github.r
@@ -174,7 +174,8 @@ github_resolve_ref <- function(x, params) UseMethod("github_resolve_ref")
 
 #' @export
 github_resolve_ref.default <- function(x, params) {
-  params$ref <- x
+  # ref can contain special characters like "#"
+  params$ref <- curl::curl_escape(x)
   params
 }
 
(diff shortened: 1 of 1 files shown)


hadley:
Would you mind moving this to remotes? (if it's not already fixed there). That's where all package downloading code is moving.

yutannihilation:
Thanks. I found this issue is already fixed in remotes by this commit: https://github.com/r-lib/remotes/commit/9bb69f18023fc189603609166ee9cbfb68b61a35

Looking forward the time when remotes is ready!

PR #1508, opened 2017-05-18
Support R 3.1

lionel-:
R 3.1 is still supported by the tidyverse. I got a devtools failure when I tried to add that platform to Travis (see https://github.com/tidyverse/rlang/pull/149).

This PR adds 3.1 to devtools's Travis and a compatibility function for `trimws()` to fix the failure.


gaborcsardi:
This code runs at install time, which can be potentially dangerous. E.g. if they change `trimws` in the future. Unlikely, but still, in general it is just better to avoid install time code in packages.

I wonder if we want to use the `backports` package, that creates the polyfill functions at load time, and very well maintained: https://github.com/mllg/backports

lionel-:
Ah nice, I didn't know about that package. I see it has a `trimws()` backport.

lionel-:
@gaborcsardi is the code in the NAMESPACE file executed at install time or at loading time?

gaborcsardi:
To my surprise, it seems that it is executed as install time, which sucks, and makes `backports` much less useful in its current state.

:( 

OK, no `backports` then.

gaborcsardi:
Anyway, I think we could still 
* inject missing functions as load time, or
* just use our own implementation, always.

gaborcsardi:
https://github.com/mllg/backports/issues/12

lionel-:
rlang could provide an exported function to conditionally assign our backports in a given environment as a function of the current R version. Then we'd call that it in the `.onLoad()` hook to populate our package namespaces.

lionel-:
@krlmlr might have other ideas

gaborcsardi:
Personally I would be happy if Michel "fixed" backports, and then we don't have to maintain it. :)

And for the current problem, I think we can just use our own replacement functions.

lionel-:
I'd have to keep hardcoding backports in rlang because it shouldn't depend on any other package. Maybe backports could be an exception though.

gaborcsardi:
Yeah, not easy to choose between dependency-free and DRY.

krlmlr:
Could we use backports by simply copying its R files?

gaborcsardi:
@krlmlr If GPL is not a problem. Ideally I would keep things in `backports`, so that we don't need to maintain it and my non-GPL packages can use it. But I understand that being dependency-free is an important virtue.

krlmlr:
rlang is GPL-2, backports is GPL-3, what am I missing?

krlmlr:
Now reading up, dtplyr (and others) have run time registration of S3 methods, we might be able to do something similar.

lionel-:
> Now reading up, dtplyr (and others) have run time registration of S3 methods, we might be able to do something similar.

That would be my preference as well.

```r
.onLoad <- function(libname, pkgname) {
  backports::import(pkgname)
}
```

With an optional argument to supply a character vector of backport names.

jimhester:
We already do assignments `.onLoad` in devtools / pkgload https://github.com/r-pkgs/pkgload/blob/master/R/aaa.r, https://github.com/r-pkgs/pkgload/blob/master/R/zzz.r anyway, pretty sure we can just do `onload_assign("trimws", backports::trimws)` anywhere in `R\` files and it should work how you intend.

lionel-:
@jimhester that would be a bit cumbersome because you'd need an else branch returning the base function. Please have a look at the last commit, I think that's the most straightforward solution.

jimhester:
I actually do not want to use backports for this, the implementation is not the same as the R version. 

It uses `gsub` rather than sub and `[[:space:]]` rather than `[ \t\r\\n]`, we should just define our own version with the base R definition.

lionel-:
Yeah I was thinking the same actually. And stringi is a heavy dep. I switched to our own implementation.

R/zzz.r
```diff
@@ -60,5 +60,26 @@ NULL
...
+      both = trimws(trimws(x, "left"), "right")
+    )
+  }
+
+  environment()
+})
+
+import_backport <- function(nm) {
+  pkg_env <- parent.env(environment())
+  assign(nm, backports[[nm]], envir = pkg_env)
```
jimhester: `assignInMyNamespace` might be a better alternative here.

jimhester:
I am pretty sure devtools >= 1.13.1 now works with R 3.1

Issue #1055, opened 2016-01-29
Description Remotes can not have spaces after the second line


lucacerone:
I am not sure whether this is actually an R (v3.2.3) issue or a devtools ( v 1.10.0.9000) issue,
so I apologize now if it is not relevant.

If in my `DESCRIPTION` file I have something like:
```
Depends:
 R (>= 3.2.3)
Imports:
 devtools (>= 1.10.0)
Remotes:
 hadley/ggplot2
```
then everything works fine.

However if I change `DESCRIPTION` to be (note: I have now two github remotes)

```
Depends:
 R (>= 3.2.3)
Imports:
 devtools (>= 1.10.0)
Remotes: 
  hadley/ggplot2
  hadley/memoise
```

then `install()` fails with error:

```
Downloading GitHub repo hadley/ggplot2
hadley@master
from URL https://api.github.com/repos/hadley/ggplot2
hadley/zipball/master
Error in curl::curl_fetch_memory(url, handle = handle) : 
  URL using bad/illegal format or missing URL
```

I have experienced similar issues using git:: url:: and so on, although now I am not able to reproduce the issue with them.

jimhester:
`Remotes: ` works the same as `Imports: ` etc, you need commas in between each entry.

```diff
@@ -3,5 +3,5 @@
 Imports:
  devtools (>= 1.10.0)
 Remotes:
-  hadley/ggplot2
+  hadley/ggplot2,
   hadley/memoise
```

lucacerone:
thanks (embarassed)

PR #1511, opened 2017-05-18
Throw error if commas missing in remotes

ianmcook:
If commas are missing between entries on separate lines under `Remotes:` in DESCRIPTION, the resulting error is:
```r
Error in curl::curl_fetch_disk(url, x$path, handle = handle) : 
  URL using bad/illegal format or missing URL
```
This does not provide much help to the user.

I think omitting these commas is a common mistake. For example: https://github.com/hadley/devtools/issues/1055

With the changes in this pull request, devtools throws an informative error in this case. The code checks for a newline in any of the remotes entries after splitting them on commas followed by optional space characters (which includes newlines).


jimhester:
Thanks! 

I merged this manually (https://github.com/hadley/devtools/commit/668a580f8c0b068df960863f8ee3b25b42b37f050) so GitHub did not close this PR automatically.

PR #1512, opened 2017-05-22
Fix use_readme_rmd() so that a pre-commit hook is registered if pkg uses git

PeteHaitch:
AFAICT, this small bug meant that the registration of the pre-commit hook was never triggered because the file path was not properly constructed.

- [x] Test added
- [ ] `NEWS.md` updated (will do if change is approved)
- [ ] Re-run roxygen2 (I did this but it introduces some doc changes unrelated to this PR, so I didn't commit these)


PeteHaitch:
Appveyor failure seems unrelated to PR

hadley:
Can you please move to [usethis](https://github.com/r-lib/usethis)? That's where all this code is moving

PR #840, opened 2015-06-08
Add forward slashes to Contributor Covenant URL in CONDUCT.md

karawoo:
The link to [contributor-covenant.org](http://contributor-covenant.org/) had no forward slashes and so doesn't show up as a link on GitHub.

PR #1515, opened 2017-05-23
Added forward slashes to URL in CONDUCT.md

seankross:
Like #840 but for the other copies of CONDUCT.md


jimhester:
This should probably go in https://github.com/r-pkgs/usethis, which is where the `use_*()` functions are migrating to.

seankross:
Thanks @jimhester, I'll do that.

PR #1520, opened 2017-05-25
Add revdep_check(..., reset = FALSE) to preserve revdep/checks/ folder

HenrikBengtsson:


hadley:
revdepcheck has better defaults for this

PR #1531, opened 2017-06-21
revdep_check: use quiet argument in revdep_check_from_cache()

ck37:
This updates revdep_check() to pass its quiet_check parameter into revdep_check_from_cache(), allowing revdep_check_from_cache() to display extra output if desired.

This is helpful for example in Travis-CI, where the current version can fail for not displaying any output for 10 minutes. It also is helpful to debug why a Travis build may fail during revdep_check().

Example output from current devtools where Travis fails due to lack of output during revdep_check(): https://travis-ci.org/ck37/SuperLearner/jobs/245155188#L2275

screen shot 2017-06-21 at 8 23 07 am

Example output from a patched devtools: https://travis-ci.org/ck37/SuperLearner/jobs/245161356#L3082

screen shot 2017-06-21 at 8 26 58 am

### DIFF
--- a/R/revdep.R
+++ b/R/revdep.R
@@ -175,7 +175,7 @@ revdep_check <- function(pkg = ".", recursive = FALSE, ignore = NULL,
   )
   saveRDS(cache, revdep_cache_path(pkg))
 
-  revdep_check_from_cache(pkg, cache)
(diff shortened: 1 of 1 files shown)


hadley:
This code has moved to revdepcheck which generally has a much better strategy for installing packages. Thanks for the code, and my apologies for not getting in touch sooner.

PR #1535, opened 2017-06-29
Remove hard-coded NA for dependences

qpcg:
Hard-coding dependencies = NA into the install_packages function within install_remote appears to cause the error:
"Installation failed: missing value where TRUE/FALSE needed"
when using install_git


hadley:
Can you please start by filing an issue that describes where this occurs?

PR #1538, opened 2017-07-07
Update use_cran_badge to https

alistaire47:
The link to the `use_cran_badge` image was redirecting to https anyway, and causing check notes like [this](https://juliasilge.com/blog/how-i-stopped/), so this PR changes `http` to `https`.

The rest of the diffs came from rebuilding the docs (and maybe an RStudio setting on trailing spaces). Ignore them if you wish.


gaborcsardi:
That check warning also happened with HTTPS, without the redirection. It was a bug in one of the Haskell packages that Pandoc used.

CRAN is using an updated Pandoc now, which should not have this problem.

As you said, the badge page redirects from HTTP to HTTPS, so it still makes sense to link the HTTPS version, to avoid the extra hop.

Changing the other files, however, in the same PR, is not very good imo.

alistaire47:
> CRAN is using an updated Pandoc now, which should not have this problem.

Good to know. 

> Changing the other files, however, in the same PR, is not very good imo.

Yeah, I was startled when I looked at the diff; I literally only typed one letter. There's a lot of automation going on with the rest, which I think is mostly the result of running `document()`. It appeared it hadn't been run in a while which is why I left the changes, but updated with a commit without them.

I found `StripTrailingWhitespace: Yes` was set in the project options (`devtools.Rproj`), which is where the other change in `infrastructure.R` came from. I turned it off, saved, and turned it on again, so the commit is now literally one letter.

hadley:
This code now lives in [usethis](https://github.com/r-lib/usethis). Can you please refile there if you're still interested?

PR #1539, opened 2017-07-10
Trying to fix #1235

### LINKED ISSUE #1235: install_github() failing on Windows due to long path names
We have a package available for install in a sub directory `R Package/ffanalytics` at `dadrivr/FantasyFootballAnalyticsR`. Users are instructed to install this via `devtools::install_github(repo = "dadrivr/FantasyFootballAnalyticsR", subdir = "R Package/ffanalytics")`. However, Windows users are reporting that the install fails with following error

```
Error in utils::unzip(src, exdir = target) :
cannot open file ‘C:/Users/<username>/AppData/Local/Temp/RtmpMZfFU0/devtools63c36d21285/dadrivr-FantasyFootballAnalyticsR-6c6c4a9/R Markdown/SubcriptionAccuracy/subscriptionAccuracy_cache/markdown_phpextra+backtick_code_blocks/unnamed-chunk-1_af5e2a896358b9ccbc8241e876b9ce5c.RData’: No such file or directory
```

The file mentioned is not part of the package directory, so I suspect that `devtools::install_github` is downloading the entire repo and then parsing it to find the subdir with the package. Since the repo files are extracted into a temp directory (`C:/Users/<username>/AppData/Local/Temp......`) the file names get rather long when it has to go through several levels of folders. 
The interesting thing is that you can get the command to complete after a number of tries - which may be when the file paths become short enough (<256 characters)

muschellij2:
I've added a `tmpdir` argument to `install_remote`, which is passed through to `source_pkg`, `source_pkg_info`, and `is_source_pkg` (and subsequently `tempfile`), which allows users to specify a "temporary" `tmpdir`, in hopes of shortening the length of the path, to help combat issue #1235.

I added `remote_package_info` previously in Nov 2016 and @hadley said most of that was going into the `remotes` package, so you may want to disregard that, or I can send another PR with a clean branch.


hadley:
I'd rather solve this by teaching the user how to set a better tempdir path

muschellij2:
I agree, that is not always wanted as a permanent tempdir path.  Also, it needs to be reset before the R session starts.  Is there harm putting this in however?   

I have a specific example that I need because CMake complains about the tempdir being too long to compile on Windows and liked this simple workaround. 

I'm teaching R right now so that may bleed over in the idea of making it "too easy" for a new user vs. them having to do the redefinition of TEMPDIR/TMP/etc in their environment variable.

hadley:
Oh I forgot that it's set before the process starts up.   But I'd prefer some other fix because there's unlikely to be devtools release for some time, and most of this code is moving into remotes.

PR #1542, opened 2017-07-16
Fix error if no misspellings found

HughParsonage:
If `spell_check` is run on a package where no words have been misspelled, the function returns an arcane error message

> Error in sprintf(fmt, "  WORD") : 
>   invalid format '%--Inf'; use format %s for character objects
> In addition: Warning message:
> In max(nchar(words)) : no non-missing arguments to max; returning -Inf

I propose that when no bad words are encountered, the function should return `NULL`. In particular, it should not error.

- [x] Style: I have kept to the style guide. (The `for(` parenthesis spacing has been left as it was.)
- [x] Added bullet point to NEWS.md
- [ ] Write a test. I was unable to write a short test. I note this function is not the subject of a unit test https://codecov.io/gh/hadley/devtools/src/master/R/spell-check.R I believe the only way to test this function or this extension would be to download a package free of spelling errors. Doing this seemed excessive for this pull request.
- [x] New parameters or new function: Not applicable.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # devtools 1.13.2
 Workaround a regression in Rcpp::compileAttributes.
 Add trimws implementation for R 3.1 support.
+* Fix arcane error message in `spell_check` when package has no spelling errors (@hughparsonage)
 
 # devtools 1.13.1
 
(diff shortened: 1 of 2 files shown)


hadley:
Duplicate of #1430

PR #1550, opened 2017-08-01
Use rcmdcheck to run and parse R CMD check output

jimhester:


R/check.r
```diff
@@ -160,29 +148,9 @@ check_built <- function(path = NULL, cran = TRUE,
...
-  }
-
-  # Record successful completion
-  writeLines("OK", file.path(package_path, "COMPLETE"))
-
-  log_path <- file.path(package_path, "00check.log")
-  parse_check_results(log_path)
+  withr::with_envvar(env_vars, action = "prefix", {
+    rcmdcheck::rcmdcheck(path, quiet = quiet, args = args)
+    })
```
hadley: Bad indent?

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # devtools 1.13.3.9000
 
+Use rcmdcheck to run and parse R CMD check output (#1153).
```
hadley: Bullet?

R/check.r
```diff
@@ -121,15 +113,11 @@ check <- function(pkg = ".",
...
 #' @param args Additional arguments passed to \code{R CMD check}
 #' @param env_vars Environment variables set during \code{R CMD check}
-#' @param needs_build_tools If \code{TRUE}, will ensure build tools are
-#'   available when checking the package.
-#' @param check_dir the directory in which the package is checked
 #' @param quiet if \code{TRUE} suppresses output from this function.
 check_built <- function(path = NULL, cran = TRUE,
                         check_version = FALSE, force_suggests = FALSE,
                         run_dont_test = FALSE, manual = FALSE, args = NULL,
-                        env_vars = NULL, needs_build_tools = TRUE,
```
hadley: Thinking about this a bit more, how does rcmdcheck handle build tools?

gaborcsardi: Minimal:
https://github.com/r-lib/rcmdcheck/blob/master/R/build.R

Maybe best would be to build, and then call `rcmdcheck` on the built tarball. Or to add build support to it. That should be just a `pkgbuild` call, no?

hadley: Yeah, I like the idea of using the pkgbuild functions. That pushes all the build tools related errors to one home.

gaborcsardi: https://github.com/r-lib/rcmdcheck/issues/31

jimhester: So is this good to merge then? The changes in rcmdcheck would be unrelated to this PR I think, since we are always explicitly building the package before checking.

gaborcsardi: Maybe wait until that is done?

gaborcsardi: So, if I do this, then `pkgbuild` needs to go on `CRAN` before the `rcmdcheck` (and `revdepcheck`) release. I assume that OK.

hadley: Yeah, that's fine. We'll probably start doing releases of those packages next week.

PR #1551, opened 2017-08-01
Use usethis for infrastructure functions.

jimhester:


hadley:
Should devtools just depend on usethis?

jimhester:
Maybe, that would save busy work importing and exporting.

hadley:
I think that's the only downside? Otherwise we'll have to keep devtools in sync, and the documentation is a pain.

jimhester:
Yeah I am fine moving it to Depends.

jimhester:
Ok switch to using Depends. I guess one more issue with this is that code which previously used `devtools::use_xyz()` will no longer work, which may break some things.

jimhester:
Looks like there are [68 usages on CRAN package with `devtools::use_*()`](https://github.com/search?utf8=%E2%9C%93&q=user%3Acran+%22devtools+use%22&type=Code)

hadley:
Oh darn. That means we need to deprecate some how. But I don't think we have a good existing strategy for deprecating functions that are moving from one package to another.

jimhester:
We can do like we did with the `with_` functions (https://github.com/r-lib/pkgload/blob/793c8245dac9c6b260470a160b53c87364a786f0/R/shims.r#L102).

We could probably remove them completely now as well, they have been deprecated for awhile at this point.

hadley:
Yeah, lets remove the withr wrappers, and take the same approach for the `use_` functions.

jimhester:
I added deprecation messages to all the `use_` functions and forwarded their arguments to the **usethis** equivalents.

`use_data()` broke using this approach because it uses basic NSE, so I simply aliased the function rather than using a wrapper. We don't have a deprecation message then, but I don't think there is a way to fix the problem short of using rlang / tidyeval.

hadley:
The plan is also to add to `Depends`, right?

jimhester:
Eventually yes, but if we put **usethis** in `Depends` in this PR we get masked warnings about all the use_ functions.

```
The following objects are masked from ‘package:usethis’:

    use_appveyor, use_build_ignore, use_code_of_conduct, use_coverage, use_cran_
badge, use_cran_comments, use_data, use_data_raw, use_dev_version, use_git,
    use_git_hook, use_github, use_github_links, use_gpl3_license, use_mit_licens
e, use_news_md, use_package, use_package_doc, use_rcpp, use_readme_md,
    use_readme_rmd, use_revdep, use_rstudio, use_test, use_testthat, use_travis,
 use_vignette
```

Once we remove the deprecated functions fully we can just put **usethis** in `Depends`.

hadley:
We could set `.conflicts.OK <- TRUE` in usethis

jimhester:
TIL about `.conflicts.OK`, sure we can do that.

PR #1560, opened 2017-08-02
Try building on R 3.0, 3.1

jimhester:


gaborcsardi:
With so many dependencies, I think this will be tricky. But let's try.

jimhester:
Failing on 3.0 due to **pkgbuild** and **pkgload**. **pkgbuild** is failing becuase **desc** is not available (https://github.com/r-lib/desc/blob/master/DESCRIPTION#L15). **pkgload** is failing because it depends on **pkgbuild** and **rlang** (https://travis-ci.org/hadley/devtools/jobs/260342442#L1473-L1477). 

https://github.com/tidyverse/rlang/issues/143

hadley:
Ok, I think 3.0.3 is going to be too much work, so lets stick with 3.1.0.  

@gaborcsardi could you do a CRAN release of desc so we get the 3.1.0 support?

gaborcsardi:
Sure, NP, I'll do it tomorrow.

I don't even know why it needs 3.1.x any more. I should really make a note about these dependencies somewhere....

hadley:
@gaborcsardi I vaguely remember some discussion we had about this but I can't remember where, or what the conclusion was. But there was definitely something that was going to be hard to work around

gaborcsardi:
@hadley: I found it, it is in the roxygen repo, of course :/ https://github.com/klutometis/roxygen/issues/582#issuecomment-278310358

So the biggest issue is probably that `testthat` does not work on 3.0.x. Anyway, I'll make a `desc` release in a minute.

gaborcsardi:
OK, `desc` submitted.

hadley:
I think we can make a push to get 3.1 everywhere and then we'll hold it there for a while and see how it goes.

Issue #1238, opened 2016-06-24
Proposal of new function: build_manual()


twolodzko:
As far as I know, devtools do not offer possibility to easily and directly build pdf manual. Maybe something like the function below could be added?

```
#' Create package manual
#' 
#' @param pkg package description, can be path or package name.  See
#'   \code{\link{as.package}} for more information
#' @param path path in which to produce package.  If \code{NULL}, defaults to
#'   the parent directory of the package.
#' 
#' @seealso \code{\link{RdUtils}} 
#' @export

build_manual <- function(pkg = ".", path = NULL) {
  pkg <- as.package(pkg)
  if (is.null(path)) {
    path <- dirname(pkg$path)
  }
  name <- paste0(pkg$package, "_", pkg$version, ".pdf", collapse = " ")
  system(paste0("R CMD Rd2pdf --force --output=", path, "/", name, " ", shQuote(pkg$path), collapse = " "))
}
```

AndreasChristianHill:
After installing the latest devtools release, I also just ran into the issue that running "check" or "build" does no longer produce a pdf-manual. It says --no-manual and I don't know how to change this. Is this a new issue of devtools or did i might have changed some settings?

twolodzko:
@AndreasChristianHill that is actually one of the reasons I started using this function... AFAIK this default behavior cannot be changed.

AndreasChristianHill:
@twolodzko Spending a bit more time reading the devtools-manual, I figured out that all options (including producing a pdf-manual) can actually be set if you call the respective functions "check" and "build" in the command line. E.g. checking with producing a manual goes like `check(pkg = ".", document = TRUE, build_args = NULL, ...,
  manual = TRUE, cran = TRUE, check_version = FALSE,
  force_suggests = FALSE, run_dont_test = FALSE, args = NULL,
  env_vars = NULL, quiet = FALSE, check_dir = tempdir(), cleanup = TRUE)`.
However, it seems to me that these options cannot be changed using the "configure build tools" in RStudio anymore. That would be worth a request @Hadley in my opinion.

twolodzko:
@AndreasChristianHill sorry, this is what I meant - Rstudio. Running `check(manual=TRUE)` works fine, but needs you to wait until compilation, while `R CMD Rd2pdf` works instantly. As about Rstudio, I did not found any way to revert the default behavior to produce pdf's.

hadley:
[...]

PR #1561, opened 2017-08-02
build_manual function

twolodzko:
Added build_manual function as discussed in  #1238.


R/build-manual.R
```diff
@@ -0,0 +1,19 @@
+
```
hadley: Can you please eliminate the extra newlines here and after the docs?

R/build-manual.R
```diff
@@ -0,0 +1,19 @@
...
+#'   \code{\link{as.package}} for more information.
+#' @param path path in which to produce package manual.
+#'   If \code{NULL}, defaults to the parent directory of the package.
+#'
+#' @seealso \code{\link{RdUtils}}
+#' @export
+
+build_manual <- function(pkg = ".", path = NULL) {
+  pkg <- as.package(pkg)
+  if (is.null(path)) {
```
hadley: This can be written more succinctly as:

```R
path <- path %||% dirname(pkg$path)
```

R/build-manual.R
```diff
@@ -0,0 +1,19 @@
...
+    path <- dirname(pkg$path)
+  }
+  name <- paste0(pkg$package, "_", pkg$version, ".pdf", collapse = " ")
+  system(paste0("R CMD Rd2pdf --force --output=", path, "/", name, " ", shQuote(pkg$path), collapse = " "))
```
hadley: Can you please use `callr::rmd()` here?

twolodzko: I made the two other edits you suggested, but I don't understand the syntax for `callr::rcmd` and I can't make it work.

hadley: It should look something like 

```R
callr::rcmd("Rd2pdf", args = c(
  "--force", 
  paste0("--output=", path, "/", name), 
  pkg$path
))
```

twolodzko: @hadley thanks, done.

hadley:
Thanks for all your work on this!

lfmcmillan:
I'm really glad there is a build_manual() function as a separate function, because when I run check(manual=TRUE) and build(manual=TRUE) from the command line in Windows 10, they don't actually create the manual, and running build_manual() seems to be the only way to successfully create the PDF manual

Issue #1568, opened 2017-08-06
The uses_git() function calls git2r with an error


potterzot:
I get an error when issuing `devtools::install()`:

    Error in git2r::discover_repository(path, ceiling = 0) : 
      unused argument (ceiling = 0)

All of my packages are up to date, and the help for `git2r::discover_repository()` just has a `path` argument, with no `check` argument.

By removing the 'check = 0' part of the code I was able to build devtools and then successfully run `devtools::install()`. Submitting a PR in a bit.

jimhester:
ceiling has been an argument in git2r since [version 0.12.1](https://github.com/ropensci/git2r/commit/2a52747392de31bd6db5cf8eaa5ae41bf7e86bf4). What version of git2r do you have installed?

```r
utils::packageVersion("git2r")
```

It seems the dependency version needs to be bumped in devtools (https://github.com/hadley/devtools/blob/master/DESCRIPTION#L30).

potterzot:
Thanks, I was at 0.11.0. Manually updating git2r (now 0.19.0) does the trick. Would you like me to do a PR with the DESCRIPTION change you highlighted?

jimhester:
Sure!

PR #1569, opened 2017-08-06
remove 'check = 0' argument passed to git2r::discover_repository().

potterzot:
This PR fixes the error in [issue #1568](https://github.com/hadley/devtools/issues/1568) by removing the `check = 0` argument passed to `git2r::discover_respository` located [here](https://github.com/hadley/devtools/blob/d5d9846ed2b0d9ae521b40a53678dd1e20fa84cc/R/git.R#L2).


hadley:
Per discussion in the issue, this seems unlikely to be the right fix

Issue #1454, opened 2017-02-07
run_examples() doesn't always respect run


bdwyer2:
When the `fresh` argument is `TRUE` the value of `run` seems to have no impact
```
run_examples(pkg = ".", run = FALSE, fresh = TRUE)
```

hadley:
Can you please provide a minimal reprex (reproducible example)? The goal of a reprex is to make it as easy as possible for me to recreate your problem so that I can fix it: please [help me help you](https://nypdecider.files.wordpress.com/2014/08/help-me-help-you.gif)! 

If you've never heard of a reprex before, start by reading "[What is a reprex](http://reprex.tidyverse.org/articles/reprex.html)", and follow the advice further down the page. Please make sure your reprex is created with the reprex package as it gives nicely formatted output and avoids a number of common pitfalls.

bdwyer2:
I think it would be difficult to create a reproducible example because I would have to create a whole package skeleton but it's easy to see the bug if you run `View(devtools::run_examples)` and look at lines 20-29.

hadley:
Maybe you'd like to do a PR instead?

hadley:
@Neil-Schneider I think you did the PR on the wrong repo.

hadley:
Oh I see you did it in the right place and @jimhester merged it

Neil-Schneider:
Yes he did. Thanks!

PR #1580, opened 2017-08-10
Refactor run examples 1454

Neil-Schneider:
# Description

This PR adds the necessary parameters to the secondary call of `run_examples()`, if `fresh = TRUE`. These parameters were not being passed into the secondary as pointed out in #1454.  

# Outstanding Issues

I have been having an issue with the `!missing(show)` warning being printed even when `show` should be missing. This only happens when `fresh = TRUE`.


Neil-Schneider:
It appears the test-examples still fails due to differences in quoting characters. (`'` vs `"`)

jimhester:
Thanks!

I merged this manually with https://github.com/hadley/devtools/commit/e215ebf00164ceafba385164c90d17603f842ddd.

You may want to add the email used in the commits `schneidernw <Neil.Schneider@crowehorwath.com>` as one of the emails associated with your GitHub username, so they are properly linked on GitHub.

Issue #1576, opened 2017-08-09
Review rstudioapis


hadley:
* Save all open files before `test()`, `load_all()` etc
* Open project with `create()`

jimhester:
Should the saving be configurable with an option? Or is there one set in the IDE we could query?

hadley:
I don't think its necessary - generally saving is not a dangerous option, and it's more frustrating to forget to save and then wonder why loading the package didn't change anything. The keyboard shortcuts do already save, so it's mostly a matter of bringing the CLI up to spec.

jimhester:
Closed by https://github.com/hadley/devtools/commit/5517fdb9f4b655eaeffdc7e371db42cddcc679e4, https://github.com/hadley/devtools/commit/ebe4cbfc65f8c2fe50a00c9cfe00571b78d29e90

PR #1584, opened 2017-08-16
Save open files in the IDE for command functions

jimhester:
I did not do `install()` as it gets called eventually for any of the `install_xyz` functions and it could be unexpected if installing a GitHub package saves your files.


jimhester:
This should close https://github.com/hadley/devtools/issues/1577, https://github.com/hadley/devtools/issues/1576 once it is merged.

PR #1587, opened 2017-08-16
bugfix: quote missing dependency package name

mtmorgan:
Confusing for some package name, e.g., 'fields' producing error message

```
Dependency package fields not available
```

### DIFF
--- a/R/package-deps.r
+++ b/R/package-deps.r
@@ -53,7 +53,7 @@ parse_deps <- function(string) {
 #' @keywords internal
 check_dep_version <- function(dep_name, dep_ver = NA, dep_compare = NA) {
   if (!requireNamespace(dep_name, quietly = TRUE)) {
-    stop("Dependency package ", dep_name, " not available.")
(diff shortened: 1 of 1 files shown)


hadley:
Thanks - we prefer to avoid `sQuote()` because it's behaviour differs from system to system making tests more challenging and leading to spurious diffs in files. But @jimhester can make the same change by hand, no need to revise the PR.

jimhester:
We actually just made a very similar change in **pkgload** (including adding quoting of package names) https://github.com/r-lib/pkgload/pull/54. I think we should export `check_dep_version()` in **pkgload** and possibly import and re-export it in devtools.

Thank you for the PR!

hadley:
Funnily enough @jennybc was just complaining about the same problem in roxygen2

mtmorgan:
See other uses of sQuote in R/utils.r

jimhester:
Other use of `sQuote()` was by me I am afraid, clearly I was not following the style guide. Sorry for the confusion, will update those as well!

PR #1595, opened 2017-08-23
Custom completions for `install_github()`.

jimhester:
These work in both command line R 
screenshot 2017-08-23 10 19 50

as well as recent versions of the IDE.
screenshot 2017-08-23 10 23 02

It queries the R GitHub package API by @gepuro.


R/completion.R
```diff
@@ -0,0 +1,16 @@
+install_github_completer <- function(env) {
+  if (completeme::current_function(env) == "install_github" && completeme::inside_quotes(env)) {
+    res <- jsonlite::fromJSON(paste0("http://rpkg-api.gepuro.net/rpkg?q=", env$token))
```
hadley: I think for safety this needs a timeout, and to be wrapped in a `tryCatch()`

hadley: I wonder if we should also support "backwards" completion so that `completeme` could autocomplete to `jimhester/completeme`

jimhester:
Ok I added a timeout and a `tryCatch()`. I set the timeout for 5 seconds, not sure what you had in mind or if this should be user configurable with a global option...

R/completion.R
```diff
@@ -0,0 +1,22 @@
+install_github_completer <- function(env) {
+  if (completeme::current_function(env) == "install_github" && completeme::inside_quotes(env)) {
```
hadley: An early return would make this code a bit simpler. Might be worth providing completeme helper? Should it also check the argument name?

hadley:
5 seconds seems long to me for something that needs to be run interactively. 

Will Escape/Ctrl + Break stop the execution of a slow completion? Does it block UI events? (i.e. can you keep typing while it looks up?)

R/completion.R
jimhester: Yeah that makes sense, the query wouldn't have to be altered for this, just the regular expression tweaked a little

jimhester:
I cleaned up the conditional and it now checks for the argument name (or being the first argument) before trying the completion.

It also completes on any part of the repo name, not just the prefix like before.

screenshot 2017-08-25 07 00 04

gepuro:
Can I support to make this function with improving a response time?

jimhester:
This would require more work into completeme to be really functional, so I am going to close this for now.

PR #1600, opened 2017-08-30
Spell checking has been moved into a separate package.

jeroen:
The spell checker has been factored out into a separate package. It now supports more advanced features like parsing and spell checking vignettes, and creating package specific wordlists.


R/release.r
```diff
@@ -40,9 +40,10 @@
 #' @param args An optional character vector of additional command
 #'   line arguments to be passed to \code{R CMD build}.
 #' @param spelling language or dictionary file to spell check documentation.
-#' See \code{\link{spell_check}}. Set to \code{NULL} to skip spell checking.
+#' See \code{\link[spelling:spell_check_package]{spelling}} package. Set to
+#' \code{NULL} to skip spell checking.
 #' @export
-release <- function(pkg = ".", check = TRUE, args = NULL, spelling = "en_US") {
+release <- function(pkg = ".", check = TRUE, args = NULL, spelling = "en_GB") {
```
hadley: Is there a reason to change this default?

R/spell-check.R
```diff
@@ -1,79 +1,7 @@
...
-#' @param ignore character vector with words to ignore. See
-#'   \code{\link[hunspell:hunspell]{hunspell}} for more information
-#' @param dict a dictionary object or language string. See
-#'   \code{\link[hunspell:hunspell]{hunspell}} for more information
-spell_check <- function(pkg = ".", ignore = character(), dict = "en_US"){
-
-  check_suggested("hunspell")
-
+spell_check <- function(pkg = ".", vignettes = TRUE, lang = "en_GB"){
+  .Deprecated("spelling::spell_check_package()", package = "devtools")
```
hadley: Is there a good reason to deprecate this, rather than soft-deprecating it?

jeroen: No, I just followed the example from the `use_this` functions. We can also keep it there as a wrapper if you'd like.

R/release.r
jeroen: To match the default in the `spelling` package.

R/spell-check.R
hadley: We deprecated the usethis functions because we're going to attach usethis so generally you'll get the new versions (the deprecation is just for people importing via namespace)

R/release.r
hadley: It's probably best to keep as is here, and as much as I hate to say it, it's probably to default to US english even in the spelling package (and make it an option)

R/spell-check.R
jeroen: OK i have updated the PR to not deprecate.

jeroen:
OK squashed and pushed.

hadley:
Thanks!

PR #1633, opened 2017-10-20
Fix the test failures / errors and add a warning for pkg arguments that are not `.`.

jimhester:
Would like suggestions on the warning message.

https://github.com/hadley/devtools/compare/master...jimhester:github-connection-tests?expand=1#diff-475284c83f91b5f64ec0e4afdbd4c955R191


hadley:
Probably should say `usethis::proj_set()` instead of `setwd()`.

Issue #1592, opened 2017-08-23
Unused argument error in `pkgbuild::build`


ShanSabri:
```
==> devtools::check()

Updating TestPkg documentation
Loading TestPkg
Setting env vars --------------------------------------------------------------
CFLAGS  : -Wall -pedantic
CXXFLAGS: -Wall -pedantic
Building TestPkg ---------------------------------------------------------------
Error in pkgbuild::build(pkg$path, tempdir(), quiet = quiet, args = build_args,  : 
  unused argument (check_dir = "/Users/some/parent/directory/to/TestPkg")
Calls: <Anonymous> -> <Anonymous> -> force
Execution halted

Exited with status 1.
```

If I run `devtools::check()` from the RStudio console, then it returns `Error: Function documentSaveAll not found in RStudio`. 

Session info posted below. I've just upgraded my R to the latest version: 
```
R version 3.4.1 (2017-06-30)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS Sierra 10.12.6

Matrix products: default
BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.4/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] devtools_1.13.3.9000 usethis_0.0.0.9000  

loaded via a namespace (and not attached):
[1] compiler_3.4.1        tools_3.4.1           withr_2.0.0           rstudioapi_0.6.0.9000 memoise_1.1.0         pkgload_0.0.0.9000   
[7] digest_0.6.12         pkgbuild_0.0.0.9000   rlang_0.1.2.9000     
```

jimhester:
Make sure you are using fully upgraded versions of the packages.

```r
devtools::install_github("hadley/devtools")
```

Should upgrade all of them for you automatically.

The `documentSaveAll()` looks like it needs IDE version 1.1.287 or later to work. We will need to add code to devtools which checks that these functions are available.

fishnerdsunite:
[...]

Issue #337, opened 2013-08-05
Can I override the `check` path?


yoni:
I'm running continuous integration builds on a Jenkins server and need the `R CMD check` output to remain in the project working directory. Otherwise, I can't  diagnose the issues on build failures.

I've tried a couple of approaches, including passing `check(..., args=c('--output', 'my_check_dir'))`. The problem is, I'm pretty sure `devtools` overrides the path separately, resulting in a conflict.

PR #1634, opened 2017-10-20
Add check_dir argument for backwards compatibility

jimhester:
This adds back the check_dir argument, although it is now ignored. The RStudio IDE calls `devtools::check()` with this argument, which caused #1592.

The `check_dir` argument was originally added for #337, however rcmdcheck does support the `--output` argument to `R CMD check`, so you can specify the check directory by passing `args = "--output=/foo/bar"` (as I mentioned in the help for `check_dir`.

Alternatively we could modify `rcmdcheck()` to take a `check_dir` argument or add the `args = "--output"` if `check_dir` is not `NULL`. Not sure what the best approach is.


hadley:
Thanks!

PR #1635, opened 2017-10-20
Add covr helper

### LINKED ISSUE #1628: Add covr helper

hadley:
Fixes #1628


R/test.r
```diff
@@ -68,6 +62,19 @@ test <- function(pkg = ".", filter = NULL, ...) {
...
 
+#' @export
+#' @rdname test
+test_coverage <- function(pkg = ".", ...) {
+  pkg <- as.package(pkg)
+
+  check_suggested("covr")
+
+  coverage <- covr::package_coverage(pkg$path, ...)
+  covr::shine(coverage)
```
jimhester: I would use `covr::report()`. I think I am going to be removing the `shine()` alias in the near future, as it no longer produces a shiny app, so doesn't really make sense.

PR #1644, opened 2017-10-26
Reinstate package coercion in check_win family

jennybc:
Otherwise we get:

```
check_win_release()
#> Error in pkg$package : $ operator is invalid for atomic vectors 
```


hadley:
Thanks!

PR #1665, opened 2017-12-03
Add function create_github_pat to make generating PATs super easy.

jeroen:
Homebrew has something similar and it is actually very convenient. Sends you directly to the form to generate the token with prefilled fields.


jimhester:
I think this is good, but I think it really should be in [usethis](https://github.com/r-lib/usethis/pulls) rather than here.

jeroen:
Are you sure? The function is not changing anything in R or in the package. It just sends the user to a webpage to generate a token. I thought it would be nice to accompany the `github_pat()` function.

What would it be called in usethis?

jimhester:
`use_github_pat()`?

jeroen:
But that implies that the github pat is actually installed? That is actually a different feature. I would expect you would have to call `use_github_pat("2b7a27ced93ca519")` or something and then it would modify your Rprofile or so?

The PR above only opens a browser window.

hadley:
This feels like a usethis function to me - maybe `browse_github_pat()`?  We could eventually have `use_github_pat()` once we have more wrappers around modifying `.Renviron`

jimhester:
Closing this here, I opened an usethis issue for it https://github.com/hadley/devtools/pull/1665

PR #1672, opened 2017-12-08
Use foghorn to report results of CRAN checks during release process 

fmichonneau:
As suggested by Hadley in fmichonneau/foghorn#15, I included the details of the CRAN check results during the release process.

Instead of having: "Have you fixed all existing problems at <URL>?", `release()` now shows the output of the CRAN check results directly in the terminal before asking the question. See below an example for `purrrlyr`. PTAL.

screenshot from 2017-12-08 15-13-08


jimhester:
Thanks! Unfortunately foghorn has an implicit dependency on stringr, (through rvest->selectr->stringr) so we cannot directly depend on it in devtools. You will need make this code conditional, e.g. `if (requireNamespace("foghorn")) { ... }` in `devtools::release()` and put foghorn in `Suggests:` instead.

Also please add a note to NEWS.md with the change and mentioning the issue number and your GitHub username.

fmichonneau:
Thanks @jimhester! I updated the pull request to make the use of foghorn conditional. The travis tests fail but it looks unrelated to my changes. Let me know if the logic I use to only display the "have you fixed all existing problems at..." message when:
- foghorn is absent OR
- if foghorn detects that there are things to fix

seems reasonable. PTAL.

jimhester:
Looks great, thanks!

PR #1678, opened 2017-12-22
[my_unzip] add exception

theGreatWhiteShark:
Whenever R is compiled with no 'unzip' package installed in the underlying operation system, the getOption( 'unzip' ) variable is set to an empty string. (At least for R-3.4.0 on a vanilla Ubuntu16.04

Since the empty string doesn't trigger the if clause and its usage of the R interal utils::unzip function, I added another exception.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -59,6 +59,8 @@
 * In order to not run test helpers in `document()`, the `helpers` argument of
   `load_all()` is set to `FALSE` (@nbenn, #1669)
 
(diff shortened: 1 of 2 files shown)


jimhester:
Thanks! This is fine, you just need to use the non-vectorized `||` function.

Make that change and add a note to NEWS.md mentioning this change, the PR number and your GitHub handle and I can merge this.

theGreatWhiteShark:
Alright, done.

jimhester:
Thanks!

PR #1712, opened 2018-02-20
Fix build_win() alias to check_win()

lionel-:
Also the convention in the tidyverse is to soft-deprecate before deprecating for at least one minor release cycle (i.e. middle number). Should we follow the same convention in r-lib? cc @hadley


hadley:
We're a little more aggressive in devtools because people should only be using these functions interactively; we don't need to care so much about breaking CRAN revdeps.

lionel-:
This is breaking ESS but I understand if you don't care about it ;)

jimhester:
Thanks!

PR #1713, opened 2018-02-20
Add an error_on argument to check and check_built

jimhester:
Fixes https://github.com/rstudio/rstudio/issues/2235

### DIFF
--- a/R/check.r
+++ b/R/check.r
@@ -65,7 +65,9 @@ check <- function(pkg = ".",
                   env_vars = NULL,
                   quiet = FALSE,
                   check_dir = tempdir(),
-                  cleanup = TRUE) {
+                  cleanup = TRUE,
+                  error_on = c("never", "error", "warning", "note")) {
+
   pkg <- as.package(pkg)
   withr::local_options(list(warn = 1))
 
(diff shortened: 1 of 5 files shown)


jimhester:
@jennybc since you were the original impetus for this saga I feel like it is only right I have you review this ;)

R/check.r
```diff
@@ -77,6 +79,11 @@ check <- function(pkg = ".",
     warning("`cleanup` is deprecated", call. = FALSE)
   }
 
+  if (missing(error_on) && !interactive()) {
```
jennybc: I am glad to pick up this pattern: how to reveal default args in signature and use `match.arg()`, while not necessarily defaulting to the first element.

jimhester:
Gah, git failure, thanks Jenny

R/check.r
jimhester: FWIW I generally think you _should_ default to the first argument, just because that is the normal convention. In this particular case I wanted to have different default behavior when run non-interactively, so needed to do this.

Maybe I should have used `error_on = NULL` so could have had the logic only in one place rather than two, I am not sure...

gaborcsardi: yeah, think about the kittens before you use `missing()` :)

jennybc: The exact thing just came up in usethis re: incrementing versions. It make sense to have "major", "minor", "patch", "dev", in that order. But in the absence of explicit info ... you probably don't want to increment the major version (in fact, we don't increment at all in that case). Also, I like revealing the possible values in the signature vs. documentation.

jimhester:
Thanks for the reviews!

PR #1719, opened 2018-02-24
Adjust to git2r's S4 --> S3 transition

jennybc:
WIP, more to start this conversation

git2r is transitioning from S4 to S3 and I discovered these problems with dev devtools and dev git2r

@stewid do you have any thoughts about packages that were using `@` to access slots in the S4 objects, now that you're returning S3 objects?

### DIFF
--- a/R/git.R
+++ b/R/git.R
@@ -4,13 +4,13 @@ uses_git <- function(path = ".") {
 
 # sha of most recent commit
 git_repo_sha1 <- function(r) {
-  rev <- git2r::head(r)
+  rev <- utils::head(r)
   if (is.null(rev)) {
     return(NULL)
   }
 
   if (git2r::is_commit(rev)) {
-    rev@sha
+    as.data.frame(rev)$sha
   } else {
     git2r::branch_target(rev)
   }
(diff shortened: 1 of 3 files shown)


stewid:
Maybe the transition of S4->S3 in git2r on CRAN should be a three-step process.  i) identify all slots that are used directly in packages and add an accessor function for each of them, for example, `sha`. ii) release a new S4 version of git2r on CRAN with those accessor functions. iii) About a month later, release the new S3 version of git2r.

hadley:
@stewid it's not clear to me that actually makes life substantially easier for developers. You'd still need a `packageVersion("git2r")` switch if you wanted to maintain backward compatibility.

stewid:
@hadley thanks for the advice. I have found six reverse dependency packages on CRAN using `@` mostly `@sha`. Maybe it's better to synchronize an update on CRAN. I'm not sure when I'm ready with a release candidate.  At the moment git2r does not build on CRAN because of library dependencies on github e.g. libssh2

hadley:
If it's only six, I wouldn't bother with multiple releases - just give authors a month's notice before you submit to CRAN.

jimhester:
I ended up finishing this with https://github.com/r-lib/devtools/commit/21fe55a912ca4eaa49ef5b7d891ff3e2aae7a370, which GitHub did not pickup as merging this PR, so will close this.

Thank you for getting it started @jennybc, it made it pretty easy to finish off!

PR #1732, opened 2018-03-14
Require git2r >= 0.21.0.9002.

jdblischak:
The current version with the S3 methods is [0.21.0.9002](https://github.com/ropensci/git2r/blob/d17df22adeba0340d9d198b03809c571765df6d2/DESCRIPTION#L7).

This is a tiny PR, so I don't think this requires any mention in the NEWS. Please let me know if you feel otherwise.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,7 +19,7 @@ Imports:
     callr,
     cli,
     digest,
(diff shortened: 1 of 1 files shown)


jimhester:
Whoops, typo on my part, thanks!

PR #1737, opened 2018-03-20
Prohibit winbuilder urls in cran-comments

HughParsonage:
Since R CMD check doesn't check `cran-comments.md`, but links to the win-builder logs are not permitted in a submission, their presence should be accompanied by a warning in `release()`. 

(Not sure how common this issue is, but I did get pinged by Uwe for this -- then almost made the same mistake next time.)

The R 3.2 error is I believe unrelated to this PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # devtools 1.13.3.9000
 
+* `check()` now emits a warning if `cran-comments.md` contains a link to
+  ephemeral `win-builder` results (@hughparsonage)
+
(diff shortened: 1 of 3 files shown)


jimhester:
Thank you for the pull request! However this seems a too specific issue to include in the general devtools checks. Maybe as an alternative you could add it to the `release_questions()` function in your package, to remind you to avoid them in cran-comments? See `?release` for details

> You can add arbitrary extra questions by defining an (un-exported)
     function called ‘release_questions()’ that returns a character
     vector of additional questions to ask.

PR #1743, opened 2018-03-24
Be explicit about the version checked

krlmlr:
in `check_win()`.

Are there other occasions where we want to show the version?


jimhester:
Not sure when you would not know the version you are checking, but I guess it doesn't hurt to include this.

krlmlr:
The winbuilder service allows submitting the same package twice if the version number is different.

PR #1745, opened 2018-03-28
Add manual = FALSE arg to check_win_xxx()

### LINKED ISSUE #1720: check_win_xxx() need manual = FALSE arg
to pass on to `build()` for DBI. Should this be fixed here or elsewhere?

krlmlr:
Closes #1720.

I have no idea why `check_win()` is exported in `NAMESPACE` but has no `\usage{}`. This is likely to trigger `R CMD check` errors.

I'm regularly using `check_win()`, would be great if it was exported but I'd resort to `:::` if it wasn't ;-)

### DIFF
--- a/R/check-win.r
+++ b/R/check-win.r
@@ -11,6 +11,8 @@
 #'   \code{\link{as.package}} for more information
 #' @param args Any additional arguments to pass to \code{\link[pkgbuild]{build}}
 #' @param quiet If \code{TRUE}, suppresses output.
+#' @param manual If \code{TRUE}, build manual. Required e.g. for dynamic
+#'   documentation using \code{\\}\code{Sexpr{}}.
 #' @export
 #' @family build functions
 #' @name check_win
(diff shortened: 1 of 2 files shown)


jimhester:
Because the `check_win_()` functions are being documented with name of `check_win`. Looks like a roxygen bug really, but I will probably just change the name of the Rd file.

Also the intent is not to export `check_win()`, just use the exported alternatives instead.

krlmlr:
Then it's `:::`. I can take care of it in this PR if you want.

jimhester:
I ended up doing this differently and just passed `...` to `pkgbuild::build()` so you can set whatever arguments you need.

Also the real reason that `check_win()` was being exported was there was a extraneous `#' @export`, https://github.com/r-lib/devtools/blob/7947bf6e620b4f3252191fade38d15910835e382/R/check-win.r#L14

krlmlr:
Thanks!

krlmlr:
The `...` is a much better solution indeed. Can we please have the same for `submit_cran()`? I need to submit with `compile_attributes = FALSE` because I need to stay out of sync with Rcpp's opinion for compatibility reasons (https://github.com/krlmlr/bindrcpp/commit/dc4c6f12c8e5eab4d5d550e9bbed8d3cf0c5f346).

Issue #1755, opened 2018-04-05
test_file helper


hadley:
It would be very useful to have a function that helps you test a single file, allowing you to iterate more rapidly when working on a single a file. The function should take a file path as an argument - it should be a path to a file in `R/` or `tests/testthat`. If omitted, it should default to using the RStudio active file.

It will need to:

* Save all open files
* Re-load the code
* Execute the single test file

Ideally it would be paired with a `test_coverage()` helper.

PR #1757, opened 2018-04-06
Add test_file function

jimhester:
This allows you to have tight iteration when editing a single R or test
file (#1755)


jimhester:
This basically just wraps `test()` and generates the `filter` argument to match the test / source file. Seems to work well and I didn't want to duplicate the package setup code in `test()`.

jimhester:
It also re-uses some logic from `usethis:::find_test_name()`.

jimhester:
Thanks! This is pretty nice actually, not sure why we didn't do this earlier :)

PR #1758, opened 2018-04-06
Add test_coverage_file to run covr::report on the currently active file

### LINKED ISSUE #1755: test_file helper

jimhester:
Fixes #1755 

This works pretty well, the only downside is you need to click through the first table even if there is only one file which is slightly cumbersome. The upside is if you pass more than one file it works nicely.

Fixing this would require some changes in covr.


jimhester:
Sounds good, did so https://github.com/r-lib/covr/issues/308

PR #1779, opened 2018-05-14
add pkgdown site

jayhesselberth:
roughly organize topics

activate `/docs` directory in settings and add URL to front github front page.

Preview here: https://xenodochial-davinci-070ddb.netlify.com/


jimhester:
Thanks, this is great!

PR #1793, opened 2018-06-04
Implement install_gitlab

jimhester:
With the recent news of Microsoft acquiring GitHub there will likely be
more demand for use of GitLab


jimhester:
This does not support the PR format like we do with GitHub, nor releases, as AFAIK there is no 1-1 equivalent of releases in GitLab. It _does_ support tags, SHAs and branches, e.g. all of the below work

```r
devtools::install_gitlab("jimhester/covr")

# tag
devtools::install_gitlab("jimhester/covr@v3.1.0")

# branch
devtools::install_gitlab("jimhester/covr@static_reports")
```

It also does the same thing as GitHub and only installs if the remote reference has changed

R/install-gitlab.r
```diff
@@ -0,0 +1,186 @@
+#' Attempts to install a package directly from GitLab
+#'
+#' This function is vectorised on \code{repo} so you can install multiple
+#' packages in a single command.
+#'
+#' @inheritParams install_github
+#' @param repo Repository address in the format
+#'   \code{username/repo[/subdir][@@ref|#pull]}. Alternatively, you can
```
hadley: Drop `#pull`?

R/install-gitlab.r
```diff
@@ -0,0 +1,186 @@
+#' Attempts to install a package directly from GitLab
+#'
+#' This function is vectorised on \code{repo} so you can install multiple
+#' packages in a single command.
```
hadley: Mention that we won't reinstall if sha is the same?

jimhester:
Re: tests yes agreed, I didn't mainly because there were none for other install_ functions.

R/install-remote.R
```diff
@@ -60,20 +60,8 @@ install_remote <- function(remote, ..., force = FALSE, quiet = FALSE,
           repos = repos, type = type)
 }
 
-try_install_remote <- function(..., quiet = FALSE) {
```
jimhester: I removed this mainly because it was useful for running revdep checks, it was added in https://github.com/r-lib/devtools/pull/1284 and we no longer use devtools to do this.

It also makes the user experience when installing a single repository worse, as there is no longer an explicit error on failure to download the package.

jimhester:
Ok I added tests, mostly adapting tests from remotes for install_github.

jimhester:
Thanks for the review, I will merge this for now, we can fix any additional issues that crop up as needed later.

PR #1804, opened 2018-06-26
Correctly detect RemoteSha of packages installed locally

lionel-:
So they are not reinstalled by install_deps()

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # devtools 1.13.3.9000
 
+* Fixed an issue where dependencies installed from a local source
+  would always be reinstalled.
+
 * `install_gitlab()` function added to install repositories from GitLab,
   (#1249, #716).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,8 @@
 # devtools 1.13.3.9000
 
-* Fixed an issue where dependencies installed from a local source
-  would always be reinstalled.
+* local remotes now use the package version as the RemoteSha unless they were installed from a git commit
+  with a clean working directory. This prevents unwanted reinstalls for development packages in possibly 
+  undesired states (@lionel-, #1804).
 
 * `install_gitlab()` function added to install repositories from GitLab,
   (#1249, #716).
--- a/R/install-remote.R
+++ b/R/install-remote.R
@@ -199,7 +199,14 @@ package2remote <- function(name, repos = getOption("repos"), type = getOption("p
       path = x$RemoteUrl,
       branch = x$RemoteBranch,
       subdir = x$RemoteSubdir,
-      sha = x$RemoteSha %||% x$Version,
+      sha = {
+        # Packages installed locally might have RemoteSha == NA_character_
+        sha <- x$RemoteSha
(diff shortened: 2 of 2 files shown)


jimhester:
The current behavior is by design. The SHA is recorded only if the local package was installed from a git directory with a clean working tree, otherwise it is NA, as we cannot be sure if the package has changed since the last time it was installed.

If the local package does have a clean working tree then the SHA will be recorded, and it will only be re-installed if the package has changed. See https://github.com/r-lib/devtools/pull/1027 for when this was added.

lionel-:
Oh, so the dependency gets reinstalled from the local source?

This is quite confusing. Should we print some hints to explain why and whence the package is reinstalled?

lionel-:
I'm not sure I like this behaviour. If I work on say Rcpp, install it from source, then work on it a bit more, it will get automatically reinstalled with the WIP changes which is often not (maybe never?) what I'd want.

Or am I missing something? I guess the issue is that I rarely have a clean working directory (I might have changed the maintainer email in DESCRIPTION to redirect winbuilder and rhub to my mailbox for instance). I wonder if we need an option in `devtools::install()` to install from last commit rather than the working tree, then it'd have the proper SHA.

jimhester:
I don't really think we need an option, but you could write a helper to stash the current changes and install if you prefer this behavior. Note git2r doesn't currently have an API for `git stash apply` (https://github.com/ropensci/git2r/issues/358), so you need to use command line git.

```r
install_last_commit <- function(...) {
  system("git stash")
  devtools::install(...)
  system("git stash pop")
}
```

lionel-:
Could you tell me more about the intended workflow for the current behaviour of local installs please? I don't understand it. Is it not dangerous to reinstall from working tree each time? Can a user reasonably expect this behaviour? Wouldn't they expect the intalled version to remain the same as the last explicit install?

jimhester:
With the option `upgrade_dependencies = TRUE` (the default) devtools upgrades all dependencies when installing a package. You can set this to `FALSE` to avoid this.

Local dependencies are treated like any other dependency, GitHub, CRAN or otherwise, if the remote has changed since the previous install than the package will be upgraded from the previous location assuming `upgrade_dependencies = TRUE`.

If you install local dependencies with a dirty working directory there is no way for us to know what the state of the package is when it is installed, so we always will reinstall it. If you install the package with a clean working directory then the SHA will be recorded, and it will only be re-installed if the SHA changes.

lionel-:
> You can set this to FALSE to avoid this.

I have, so there might be a bug.

> Local dependencies are treated like any other dependency.
>
> there is no way for us to know what the state of the package is when it is installed, so we always will reinstall it.

Despite the apparent consistency this does not seem to be the right thing to do. A working tree is not necessarily in working order. You can assume that the user knows that the state was good at time of installation, but not at random ulterior times such as when upgrading dependencies.

This behaviour seems unintuitive and dangerous to me. It should at least be told to the user in the output, something like "Reinstalling somepkg because it was installed with a WIP git working tree. Install from a clean working tree to avoid this."

PR #1815, opened 2018-07-06
Housekeeping edits to release.r

ntdef:
Original docs say that the package is released with FTP when in fact it's now down with a POST request to an HTTP endpoint. 

Also found a stray line that wasn't doing anything.


jimhester:
Thanks!

PR #1823, opened 2018-07-20
Forward path to usethis::use_rstudio()

lionel-:


jimhester:
Changing the working directory would only help if you had not called a usethis function to set the project yet. Once the usethis project is set the working directory is ignored. You need to use `usethis::proj_set()` to change the active project rather than changing the working directory.

Perhaps instead `warn_unless_current_dir()` should be changed to throw an error if `pkg$path` does not equal `usethis::proj_get()`

jimhester:
Ok I have added temporarily setting the active project to `warn_unless_current_dir()`, so we can preserve existing behavior (with a warning) for all usethis functions in devtools.

lionel-:
Thanks

PR #1824, opened 2018-07-20
Use better vignette workflow

jimhester:
This installs the built vignettes to doc/ rather than inst/doc and
builds a vignette index in Meta/vignette.rds. This means that you can
view development vignettes by loading the package with `load_all()` and
calling `browseVignettes(pkgname)` or `vignette(vignName)` and the devel
vignettes are found. The `doc` and `Meta` directories are added to
.Rbuildignore; so they will not be included in the built package.

This does not currently do anything with intermediates, doing so is tricky assuming we want generality with non-knitr / rmarkdown vignette engines.


R/vignettes.r
```diff
@@ -2,26 +2,33 @@
...
+#' files can be checked into version control, so they can be viewed with
+#' \code{browseVignettes()} and \code{vignette()} if the package has been
+#' loaded with \code{load_all()} without needing to re-build them locally.
 #'
 #' @param pkg package description, can be path or package name.  See
 #'   \code{\link{as.package}} for more information
 #' @param quiet If \code{TRUE}, suppresses most output. Set to \code{FALSE}
 #'   if you need to debug.
+#' @param install If \code{TRUE}, install the package before building
+#'   vignettes.
```
jennybc: This installs current dev version to a temporary library that is used just to render the vignette, right? I believe this happens elsewhere too now (README?), so it would be good to nail the wording and use in all places. Specifically, IIUC, the installation referred to here does not affect one's normal library and has no persistent effect. It makes the dev state of vignettes and README more parallel to docs below `man/`.

doc/dependencies.R
```diff
@@ -0,0 +1 @@
+
```
jennybc: Is this expected?

jimhester: Yes, the vignette has no R code, which is why there is nothing here

R/vignettes.r
jimhester: Yes, that is right. I added this option mainly because it was taking a while to install the package when I was writing / testing this. We could alternatively just remove the option...

hadley: Either way, need to describe the temporary install in the docs.

doc/dependencies.Rmd
```diff
@@ -0,0 +1,88 @@
+---
```
hadley: Do we want to check this in? How will we keep this up to date? Do you imagine these changes will also flow into the roxygen vignette roclet? 

Should we also create an md version for easy previewing on github?

jimhester: I don't know about whether to check in or not, maybe we should not, but the benefit would be users would not have to build vignettes themselves. You would have to update it with `devtools::build_vignette()`.

Intermediates like `.md` are challenging to deal with without special casing rmarkdown vignettes and rendering them separately. It is somewhat questionable how much utility they have particularly now that we have pkgdown to view devel vignettes.

jennybc: Good point re: pkgdown. Some of my vignette whining may no longer relevant. If I have a balky vignette (takes a non-trivial amount of time to render, calls internet, needs auth, etc.), I would turn it into an article now.

That just transfers some whining from devtools to pkgdown 😅 but I still think lowers the overall amount of pain. Why? Because pkgdown is designed to make the built product (the site) constantly available, whereas the official vignette workflow is much more tied to an official build/release event.

hadley:
I don't think this solves the whole vignette workflow, but it obviously improves things.

For tidyverse packages, I think there's a strong pressure to move towards articles and away from vignettes since the vignettes often need graphics and hence end up rather large. Downloading large vignettes that are more easily discoverable from the web doesn't seem like a big win.

hadley:
(To be clear I think you are good to merge)

ifellows:
Just a note on this. I now have to have an extra step in my workflow where I have to manually copy the doc directory to inst/doc. In many packages they take some time to run and prebuilt vignettes are a good thing. Plus I point people to the vignettes on github as part of the documentation (see for example https://github.com/statnet/lolog), so having them prebuilt and up-to-date is important. Not a huge deal, but definitely not an expected change.

I can't say that I understand the rational either. I've never run `vignette(vignName)` while developing a package, so that didn't seem like a pressing issue (at least for me). I would have thought it would be more important to put the files where CRAN expects them.

PR #1825, opened 2018-07-20
Add build_package_index to build a proper package index

jimhester:
This allows the RStudio help to work for development packages like it
does for installed packages

I am not entirely sure this is a good idea, but it does allow you to view the package index without installing the package, which is useful in RStudio and for finding package vignettes...

### DIFF
--- a/R/document.r
+++ b/R/document.r
@@ -45,5 +45,8 @@ document <- function(pkg = ".", clean = NULL, roclets = NULL, reload = TRUE) {
   )
 
   pkgload::dev_topic_index_reset(pkg$package)
+
+  build_package_index(pkg)
+
   invisible()
 }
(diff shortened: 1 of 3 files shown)


jimhester:
One issue with this is making sure to remove the lazy load database from memory before updating, otherwise you get `Error in fetch(key) : lazy-load database '.../devtools/help/devtools.rdb' is corrupt` errors when trying to re-load a help file that was previously loaded.

jennybc:
To record this thought from group meeting: I'm not sure how important it is to make the index available this way. From the roxygen workflow, I am already accustomed to the idea that I must properly build and install to have my docs fully "wired up", e.g. all links working. I am prepared to accept the analogous limitation for vignettes.

jimhester:
I think we are going to wait on this for now.

PR #1849, opened 2018-08-17
Rely on roxygen2 more

hadley:
Much of the previous code is no longer necessary as roxygen2 handles it internally. Once we can move roxygen2 back into imports, we can add a version dependency there and then delete the old code.

### DIFF
--- a/R/document.r
+++ b/R/document.r
@@ -20,20 +20,24 @@ document <- function(pkg = ".", roclets = NULL) {
     rstudioapi::documentSaveAll()
   }
 
-  # Refresh the pkg structure with any updates to the Collate entry
-  # in the DESCRIPTION file
-  roxygen2::update_collate(pkg$path)
+  if (packageVersion("roxygen2") >= "6.1.0") {
+    withr::with_envvar(r_env_vars(), roxygen2::roxygenise(pkg$path, roclets))
+  } else {
+    # Refresh the pkg structure with any updates to the Collate entry
+    # in the DESCRIPTION file
+    roxygen2::update_collate(pkg$path)
 
-  roclets <- roclets %||% roxygen2::load_options(pkg$path)$roclets
-  roclets <- setdiff(roclets, "collate")
+    roclets <- roclets %||% roxygen2::load_options(pkg$path)$roclets
+    roclets <- setdiff(roclets, "collate")
 
-  if (packageVersion("roxygen2") < "6.1.0") {
-    load_all(pkg$path, helpers = FALSE)
+    if (packageVersion("roxygen2") < "6.1.0") {
+      load_all(pkg$path, helpers = FALSE)
+    }
+
+    withr::with_envvar(r_env_vars(),
+      roxygen2::roxygenise(pkg$path, roclets = roclets, load_code = pkgload::pkg_ns)
+    )
   }
-
-  withr::with_envvar(r_env_vars(),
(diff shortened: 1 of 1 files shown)


jimhester:
Couldn't we just depend on roxygen 6.1.0 now that it is on CRAN, even with it still in Suggests? We already depend on roxygen 5.0.0 there.

hadley:
My memory is that the version requirements of suggested packages aren't enforced, but I might be wrong.

gaborcsardi:
> My memory is that the version requirements of suggested packages aren't enforced, but I might be wrong.

WRE has:

>  Version requirements can be specified but should be checked by the code which uses the package.

Kind of makes sense for a suggested package.

jimhester:
Right, but we do check this explicitly with [pkgload::check_suggested()](https://github.com/r-lib/pkgload/blob/eaae7d52e92823f6e4f7e4331f2442d03eeac8cb/R/utils.R#L32-L59) earlier in this function, although you get a warning rather than an error. (I artificially bumped the version in the DESCRIPTION as a test).

```r
check_suggested("roxygen2")
Warning in check_dep_version(package, version) :
  Need roxygen2 >= 7.0.0 but loaded version is 6.1.0
```

R/document.r
```diff
@@ -20,20 +20,24 @@ document <- function(pkg = ".", roclets = NULL) {
...
 
-  roclets <- roclets %||% roxygen2::load_options(pkg$path)$roclets
-  roclets <- setdiff(roclets, "collate")
+    roclets <- roclets %||% roxygen2::load_options(pkg$path)$roclets
+    roclets <- setdiff(roclets, "collate")
 
-  if (packageVersion("roxygen2") < "6.1.0") {
-    load_all(pkg$path, helpers = FALSE)
-  }
+    if (packageVersion("roxygen2") < "6.1.0") {
```
jimhester: We can remove this conditional I guess right, the only way you could be in this block is by being less than "6.1.0"

hadley: Doh

hadley:
Hmm, looks like it should be an error? I'm in favour of just bumping if you're ok with it.

jimhester:
Yeah just bumping sounds good to me.

hadley:
Appveyor failure seems unrelated; I think this is good to merge.

jennybc:
I am finding that I *need* this PR in order for `devtools::document()` to do the right thing with roxygen2 v6.1.0.

jimhester:
Thanks!

jennybc:
I think this is still worth pondering but I guess it's a roxygen2 matter now (?):

> A different consequence of this PR that might be undesirable is that the `helpers = FALSE` behaviour (in the sense of `load_all(pkg$path, helpers = FALSE)`) has been lost. Is that worth taking a stand on?

hadley:
Would you mind filing a roxygen2 bug so I don't forget?

jimhester:
Maybe we should be calling `pkgload::load_all(helpers = FALSE)` in roxygen2 and maybe `export_all = FALSE` to more closely mimic a normal package namespace? I am not sure what roxygen2 assumes in regards to exports.

jimhester:
Also potentially `recompile = FALSE`, which would be a way to fix https://github.com/klutometis/roxygen/issues/771

hadley:
Yeah, that seems like a good idea.

PR #1861, opened 2018-09-06
Use remotes for installing remotes

jimhester:


jimhester:
Still some work to be done, mainly opening this to test on CIs

PR #1870, opened 2018-09-14
Reload loaded packages as well as attached packages

jimhester:
Historically `reload()` only reloaded attached packages, this changes
the behavior to reload packages which are only loaded, not attached as
well.

@wch, @hadley Do either of you remember if there was a reason `reload()` only reloaded attached packages? It seems to me like it should reload loaded, but not attached packages as well.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # devtools 1.13.6.9000
 
+* `reload()` now reloads loaded but not attached packages as well as attached ones.
+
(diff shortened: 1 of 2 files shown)


hadley:
I suspect at the time it was written we might not have fully understood the difference between loaded and attached.

wch:
Yup, I think @hadley is right.

jimhester:
Ok thanks for taking a look!

PR #1873, opened 2018-09-19
check: pass check_dir to rcmdcheck

### LINKED ISSUE #1864: check() fails on Windows if check_dir has spaces
And probably also for other special characters that would need escaping. This is because `R CMD check` re-parses the arguments.

A fix will be implemented in rcmdcheck, by changing the working  directory and omitting the `--output` option. Then I'll submit a PR here to adapt `devtools::check()` to the new rcmdcheck argument.

gaborcsardi:
Instead of using the `--output` argument of `R CMD check`.
`--output` is impossible to quote correctly on windows,
because of the processx auto-quoting and/or R's
command line reparsing.

Closes #1864.

### DIFF
--- a/R/check.R
+++ b/R/check.R
@@ -162,7 +162,6 @@ check_built <- function(path = NULL, cran = TRUE,
 
   pkgname <- gsub("_.*?$", "", basename(path))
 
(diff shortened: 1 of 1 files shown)


jimhester:
Thanks!

PR #1898, opened 2018-10-23
Passed upgrade through to install_deps

Paxanator:
Not sure if I need to make an issue first, but the upgrade argument isn't being passed through to install_deps.

If there's guidance on writing a test I can do that as well


jimhester:
Yes you are correct, sorry for the oversight and thank you for the fix!

Paxanator:
Thanks @jimhester ! Is there a planned maintenance release in the future?

jimhester:
Yes, probably in the next week or so.

PR #1926, opened 2018-11-19
Fix default value for upgrade in install_deps and install_dev_deps

kiendang:
following r-lib/remotes@70959f2

Currently if using `remotes` master branch, running `devtools::install_deps` without specifying `upgrade` will lead to
```
Error in match.arg(upgrade, c("default", "ask", "always", "never")) :
  'arg' must be of length 1
```
caused by [this line](https://github.com/r-lib/remotes/blob/bd970d723facc6ede2ff0f54ce4ae917526a37f8/R/deps.R#L543)


jimhester:
Thanks!

PR #1930, opened 2018-11-23
Re-document after modifying install_deps

kiendang:
Sorry I forgot to re-document in #1926.


kiendang:
Already closed this because I saw 8a3d2c2. But then reopened because I saw the `@param upgrade` entries haven't been re-generated in doc. I guess because the latest `remotes` wasn't installed locally when re-documenting?

PR #1933, opened 2018-11-29
Update check-win.R

lbusett:
Update documentation: the Description was still using `build_win` (deprecated) instead than `check_win`


jimhester:
Thanks!

PR #1952, opened 2018-12-15
Build before confirming submission

krlmlr:
This allows reviewing the `.tar.gz` before even sending it to CRAN.


jimhester:
Thanks, I often like to inspect the built tarball myself, so this is a good idea!

PR #1988, opened 2019-02-14
Sort out title vs. description for install_deps() help; re-document()

jennybc:
Re-documenting in the presence of current remotes causes a few more positive changes, because of parameter inheritance.


jimhester:
Thanks!

PR #1989, opened 2019-02-15
Fix attaching usethis if lib path has changed

### LINKED ISSUE #1987: Weird installation error if installing to a different lib that is not on the lib path
For this you need the dev version (or just a different version, really) of usethis in a regular library, and then install usethis and devtools to another library.  E.g.

```r
dir.create(lib1 <- tempfile())
pak::pkg_install("r-lib/usethis", ask = FALSE, lib = lib1)
.libPaths(c(lib1, .libPaths()))

library(usethis)
dir.create(lib2 <- tempfile())
install.packages(c("devtools", "usethis"), lib = lib2)
```

```
** testing if installed package can be loaded
Error: package or namespace load failed for 'devtools':
 .onAttach failed in attachNamespace() for 'devtools', details:
  call: value[[3L]](cond)
  error: Package 'usethis' version 1.4.0 cannot be unloaded:
 Error in unloadNamespace(package) : namespace 'usethis' is imported by 'devtools' so cannot be unloaded

Error: loading failed
Execution halted
ERROR: loading failed
```

Is there a reason why devtools can't simply just depend on usethis?

gaborcsardi:
If the lib path has changed since loading usethis,
.onAttach potentially fails when it tries to attach
usethis as the new lib path might contain a different
usethis version.

Closes #1987.

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -68,5 +68,7 @@ devtools_default_options <- list(
 .onAttach <- function(libname, pkgname) {
   env <- as.environment(paste0("package:", pkgname))
   env[[".conflicts.OK"]] <- TRUE
-  suppressPackageStartupMessages((get("library", baseenv()))("usethis"))
+  usethis_lib <- dirname(getNamespaceInfo("usethis", "path"))
+  suppressPackageStartupMessages((
+    get("library", baseenv()))("usethis", lib.loc = usethis_lib))
 }


gaborcsardi:
I tested it, it does work.

jimhester:
Thanks! this looks good to me

PR #2012, opened 2019-03-22
Consitent order for yays and nays in release.R

layik:
Dear devtools team, I think pesudo-randomising "yays" vs "nays" serve no purpose, except confusion especially for newbies. But I would not expect it to be welcomed. Yet I will give it a go :)


jimhester:
Thanks for the PR. However the point of the function is to make you stop and think about your answers to make sure you do each step and not just blindly choose yes. So this is working as intended.

PR #2017, opened 2019-04-04
Fix typos

katrinleinweber:
See diff ;-)


jimhester:
Thanks, quite thorough!

katrinleinweber:
Partly [by reg-ex](https://github.com/katrinleinweber/PhD-thesis/blob/doctor/CHECKS-MD#L2) ;-)

PR #2061, opened 2019-06-30
Clearer links to new homes of deprecated functions

gustavdelius:
Just a small tweak to the "Details" section of the man page for the deprecated functions, making it easier for the user to follow links to the new home packages for these functions.

### DIFF
--- a/R/infrastructure.R
+++ b/R/infrastructure.R
@@ -1,5 +1,7 @@
 #' @details
-#' Instead of the use_xyz functions from devtools use \link[usethis]{use_testthat}.
+#' Instead of the use_xyz functions from devtools use the same function from
+#' the \href{https://github.com/r-lib/usethis}{usethis} package, for example
+#' \code{usethis::\link[usethis]{use_testthat}}.
 #' @rdname devtools-deprecated
 #' @importFrom usethis use_testthat
 #' @export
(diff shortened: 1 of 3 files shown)


jennybc:
Can you reword to be consistent with this?

https://r-pkgs.org/intro.html#intro-usage

The main point is that, in interactive use, people should attach devtools and just call these functions, e.g. `create_package()`, `use_testthat()`, etc. They should not be seeing deprecation messages in this scenario. devtools makes all exported usethis functions available.

The only time you should see a deprecation message is when you call a usethis function *qualified with the devtools namespace*, e.g. `devtools::use_testthat()`. That is the usage the message is intended to discourage.

Does this help clarify the usage we're [dis|en]couraging?

gustavdelius:
Yes, I understand. Perhaps the help text on devtools-deprecated should be changed as well. It currently somewhat misleadingly states: "These functions are Deprecated in this release of devtools, they will be marked as Defunct and removed in a future version."
@jennybc I think it would be best to leave the formulation up to you so that it is consistent everywhere. Many thanks for your work on devtools and the R packages book.

jennybc:
@gustavdelius Yeah, we've been in a weird state with these functions for a while, i.e. probably sending conflicting (or at least confusing) messages. @jimhester is releasing a version of devtools that should substantially straighten this out.

PR #2091, opened 2019-08-18
https URLs

hrbrmstr:
NOTE: I did not change the domain or path, just migrated `http://` to `https://` for URLs when I noticed that a few key ones (such as submitting to CRAN) would enable [MP]ITM attacks.


jimhester:
Thanks for paying attention to this! 
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`, then I would be happy to merge.

hrbrmstr:
Absolutely! Will do a bit later today.

NEWS.md
```diff
@@ -50,6 +56,8 @@
 
 * `test_file()` and `test_coverage_file()` now work with C
   and C++ files in the src/ directory as well.
+  
+*
```
jimhester: Looks like maybe an extraneous bullet got added here.

PR #2092, opened 2019-08-22
Special case document() when applied to roxygen2 itself

hadley:
To avoid crashing due to (presumably) the external pointers pointing to memory that has been reclaimed

(I didn't add a news entry since the only person this really affects is me)


jimhester:
Great work, this package is now better thanks to you!

Thanks!! You are da bomb!

PR #2130, opened 2019-10-01
Use user-specified stop_on_failure if provided. Fixes #2129

### LINKED ISSUE #2129: test() errors on user-supplied stop_on_failure
```
> devtools::test(".", stop_on_failure=TRUE)
Loading [...]
Testing [...]
Error in (function (path, filter = NULL, reporter = default_reporter(),  : 
  formal argument "stop_on_failure" matched by multiple actual arguments
```

Looks like this arose in #2099. 

Proposed fix: check for stop_on_failure in `...` and then set if not found? E.g. change current [line 61](https://github.com/r-lib/devtools/blob/master/R/test.R#L61) to something like

```
testthat_args <- list(test_path, filter = filter, env = env,  ... = ...)
testthat_args$stop_on_failure <- testthat_args$stop_on_failure %||% FALSE
```

infotroph:
Implementing the fix I proposed in #2129.

I also took a stab at some tests, but am not sure how robust my approach is -- please have a look and suggest improvements.


jimhester:
I question the need for this, but if you are going to do it I think adding `stop_on_failure` as a normal argument to `test()` is a better approach.

infotroph:
@jimhester Two questions:

(1) Can you expand on "question the need"? The current behavior makes it impossible to set `stop_on_error = TRUE`, which seems like a definite bug given `...` is documented to be passed on to `test_dir`.

(2) I'd be OK with a normal argument, but am surprised you prefer it! Given previous discussion in https://github.com/r-lib/testthat/issues/912 it sounded like "non-user-facing" was an important consideration. Was I wrong about that, or is there some addition calculus here (maybe adding it as an argument would help set up for a future change of default from FALSE to TRUE?)

jimhester:
Most users aren't going to want or need to change this default.

Being a normal argument is just much less magical, I don't think there is any need to hide this from the user.

infotroph:
Thanks for the explanation. I'll rework this as a standard argument and ping you when it's ready for re-review.

infotroph:
@jimhester This is ready for another look. I think the Travis failure is unrelated -- last few commits to master have all produced the same warning as this PR.

jimhester:
Looks good, thanks for working on this!

Great work, this package is now better thanks to you!

Thanks!! You are da bomb!

jchaskell:
I don't think the version on cran is getting this last change, so the function isn't backward compatible for people who were passing in `stop_on_failure` as an argument to the function.

donarus:
Any details about when this version will be vailable on CRAN?

Issue #2142, opened 2019-11-02
Fail to install packages dependencies in DESCRIPTION Remotes


c1au6i0:
I would like to use a dev version of a package in my own package.
As detailed in [the vignette]( https://cran.r-project.org/web/packages/devtools/vignettes/dependencies.html), I included the following line of code in the DESCRIPTION file of my package:

`Remotes: SciViews/svDialogs`

tried also:

`Remotes: SciViews/svDialogs@v1.0.2`

Unfortunately, that dependency is not installed  prior to the installation of my package. Running, `devtools::install_deps()` does not install that dependency, but installs any missing dependency in `Imports`.

On the other hand `devtools::install_github("SciViews/svDialogs")` works as expected and installs the dev version of the package.

Not sure if I am missing something important. Any ideas?

Here some other info on my session:
```sessionInfo()
#> R version 3.6.0 (2019-04-26)
#> Platform: x86_64-apple-darwin15.6.0 (64-bit)
#> Running under: macOS  10.15
#> 
#> Matrix products: default
#> BLAS:   /Library/Frameworks/R.framework/Versions/3.6/Resources/lib/libRblas.0.dylib
#> LAPACK: /Library/Frameworks/R.framework/Versions/3.6/Resources/lib/libRlapack.dylib
#> 
#> locale:
#> [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> loaded via a namespace (and not attached):
#>  [1] compiler_3.6.0  magrittr_1.5    tools_3.6.0     htmltools_0.4.0
#>  [5] yaml_2.2.0      Rcpp_1.0.2      stringi_1.4.3   rmarkdown_1.16 
#>  [9] highr_0.8       knitr_1.25      stringr_1.4.0   xfun_0.10      
#> [13] digest_0.6.22   rlang_0.4.1     evaluate_0.14
```

Thanks

jennybc:
Just to check: are you also including svDialogs in Imports in `DESCRIPTION`? You need to do that as well. It is not enough to only mention it in `Remotes:`.

c1au6i0:
No, I did not! 🤦‍♂️
Nailed, that was the problem! Thank you @jennybc

jennybc:
[...]

PR #2143, opened 2019-11-02
Vignette code examples include Imports

c1au6i0:
Just a minimal change of the vignette  examples to make immediately obvious to impatient readers that package needs to be listed in `Imports` too.
See Issue #2142


c1au6i0:
Ahh that’s right. I will just close it the

Issue #2152, opened 2019-11-26
`check_win` does not reset email to original


muschellij2:
May be related to https://github.com/r-lib/devtools/issues/1723.

When using `check_win_devel` with the `email` argument, this does not reset after submission.  I see https://github.com/r-lib/devtools/blob/master/R/check-win.R#L61 should indeed copy the original `DESCRIPTION` file over, but the example below shows that's not the case:

``` r
library(devtools)
#> Loading required package: usethis
library(desc)
path = tempfile()
create_package(path, fields = 
                 list(
                   `Authors@R` = 'person("Jane", "Doe", email = "jane@example.com", role = c("aut", "cre"),
                          comment = c(ORCID = "YOUR-ORCID-ID"))',
                   License = "MIT + file LICENSE",
                   Language =  "es"
                 ),
               open = FALSE)
#> ✔ Creating '/var/folders/1s/wrtqcpxn685_zk570bnx9_rr0000gr/T/Rtmpr1oav9/file98c05eb60133/'
#> ✔ Setting active project to '/private/var/folders/1s/wrtqcpxn685_zk570bnx9_rr0000gr/T/Rtmpr1oav9/file98c05eb60133'
#> ✔ Creating 'R/'
#> ✔ Writing 'DESCRIPTION'
#> Package: file98c05eb60133
#> Title: What the Package Does (One Line, Title Case)
#> Version: 0.0.0.9000
#> Authors@R (parsed):
#>     * Jane Doe <jane@example.com> [aut, cre] (<https://orcid.org/YOUR-ORCID-ID>)
#> Description: What the package does (one paragraph).
#> License: MIT + file LICENSE
#> Encoding: UTF-8
#> Language: es
#> LazyData: true
#> ✔ Writing 'NAMESPACE'
#> ✔ Setting active project to '<no active project>'
desc = desc::description$new(file.path(path, "DESCRIPTION"))
desc$get_maintainer()
#> [1] "Jane Doe <jane@example.com>"
devtools::check_win_devel(pkg = path, email = "bob@bob.com")
#> Building windows version of file98c05eb60133 (0.0.0.9000) for R-devel with win-builder.r-project.org.
#>   
   checking for file ‘/private/var/folders/1s/wrtqcpxn685_zk570bnx9_rr0000gr/T/Rtmpr1oav9/file98c05eb60133/DESCRIPTION’ ...
  
[...]

PR #2157, opened 2019-12-03
Added add = TRUE For `check_win`

muschellij2:
Fixing #2152 As per https://github.com/r-lib/devtools/issues/2152#issuecomment-559144986, by adding `add = TRUE` on the `on.exit/on.error` collector.


jimhester:
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Otherwise looks good, thanks!

jimhester:
Thanks a million!

PR #2171, opened 2019-12-22
install_dev_deps() passes formal args onto install_deps()

### LINKED ISSUE #2166: devtools::install_dev_deps() doesn't pass all args through to remotes::install_deps()
Hi all, I can see that all args besides `upgrade` and `...` aren't being passed through from `devtools::install_dev_deps()` to  `remotes::install_deps()`.

https://github.com/r-lib/devtools/blob/e463a32cb2ecb2f7903d73fe4b3739bf1fd131f9/R/install.R#L150-L168

If you agree I'm happy to raise a PR and add a test case to cover this.

Here's a reprex illustrating the bug 👇 

``` r
library(devtools)
#> Loading required package: usethis

# let's give it a package to install
pkg <- "docopt"

if (pkg %in% rownames(installed.packages())) {
  remove.packages(pkg)
}
#> Removing package from 'C:/Users/timjg/Documents/R/win-library/3.5'
#> (as 'lib' is unspecified)

directory <- tempdir()

create(directory, fields = list("Suggests" = pkg))
#> <U+2714> Setting active project to 'C:/Users/timjg/AppData/Local/Temp/RtmpGABqrf'
#> <U+2714> Creating 'R/'
#> <U+2714> Writing 'DESCRIPTION'
#> Package: RtmpGABqrf
#> Title: What the Package Does (One Line, Title Case)
#> Version: 0.0.0.9000
#> Authors@R (parsed):
#>     * First Last <first.last@example.com> [aut, cre] (<https://orcid.org/YOUR-ORCID-ID>)
#> Description: What the package does (one paragraph).
#> License: What license it uses
#> Suggests:
#>     docopt
#> Encoding: UTF-8
#> LazyData: true
#> <U+2714> Writing 'NAMESPACE'
#> <U+2714> Setting active project to '<no active project>'

# won't pass through our `type` arg to remotes::install_deps
install_dev_deps(directory, type = stop("oh no!")) 
#> 
#> docopt (NA -> 0.6.1) [CRAN]
#> Installing 1 packages: docopt
#> Installing package into 'C:/Users/timjg/Documents/R/win-library/3.5'
#> (as 'lib' is unspecified)
#> package 'docopt' successfully unpacked and MD5 sums checked
#> 
#> The downloaded binary packages are in
#>  C:\Users\timjg\AppData\Local\Temp\RtmpGABqrf\downloaded_packages

# expected behaviour would be to throw the error
remotes::install_deps(directory, dependencies = TRUE, type = stop("oh no!"))
#> Error in serialize(list(repos, type), NULL, ascii = TRUE): oh no!
```

<sup>Created on 2019-12-17 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

t-gibson:
Closes #2166

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -117,27 +117,63 @@ install <-
 #' `install_deps()` will install the
 #' user dependencies needed to run the package, `install_dev_deps()` will also
 #' install the development dependencies needed to test and build the package.
-#' @template devtools
-#' @inheritDotParams remotes::install_deps -pkgdir -dependencies
+#' @inheritParams install
+#' @inherit remotes::install_deps
 #' @export
-install_deps <- function(pkg = ".", ...) {
+install_deps <- function(pkg = ".",
+                         dependencies = NA,
+                         repos = getOption("repos"),
+                         type = getOption("pkgType"),
+                         upgrade = c("default", "ask", "always", "never"),
+                         quiet = FALSE,
+                         build = TRUE,
+                         build_opts = c("--no-resave-data", "--no-manual", " --no-build-vignettes"),
+                         ...) {
   remotes::update_packages("roxygen2")
 
   pkg <- as.package(pkg)
 
   check_dots_used(action = getOption("devtools.ellipsis_action", rlang::warn))
 
-  remotes::install_deps(pkg$path, dependencies = NA, ...)
+  remotes::install_deps(
+    pkg$path,
+    dependencies = dependencies,
+    repos = repos,
+    type = type,
+    upgrade = upgrade,
+    quiet = quiet,
+    build = build,
+    build_opts = build_opts,
+    ...
+  )
 }
 
 #' @rdname install_deps
 #' @export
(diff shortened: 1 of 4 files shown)


R/install.R
```diff
@@ -129,11 +129,14 @@ install_deps <- function(pkg = ".",
                          build = TRUE,
                          build_opts = c("--no-resave-data", "--no-manual", " --no-build-vignettes"),
                          ...) {
+  remotes::update_packages("roxygen2")
```
jimhester: Why did you add this?

t-gibson: I wanted to align the internals of `install_dev_deps()` and `install_deps()`. Happy to remove the changes to `install_deps()` if it shouldn't be done as part of this PR.

jimhester: It is in `install_dev_deps()` because you need roxygen2 when developing packages, but you definitely don't want to always install it whenever you are installing any dependency.

jimhester:
Thanks!

PR #2173, opened 2019-12-31
Suggest installing spelling when using spell_check

### LINKED ISSUE #2172: spell_check doesn't check if spelling is installed
When running `devtools::spell_check()` on a computer without `spelling` installed, `spell_check` fails due to the missing namespace:

```
> devtools::spell_check()
Error in loadNamespace(name) : there is no package called ‘spelling’
```

Other functions such as `devtools::check_rhub()` seem to make use of `check_suggested()` in interactive sessions to prompt users to install the package if needed, rather than just failing. Would it make sense for this function to do the same?

mikemahoney218:
Closes #2172 .

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # devtools 2.2.2
 
+* `spell_check()` now checks if `spelling` is installed and prompts users to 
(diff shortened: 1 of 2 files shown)


jimhester:
Thanks!

PR #2201, opened 2020-03-19
Generalise build_readme() to build_rmd()

hadley:
Also needed to make some changes to `install()` so that it doesn't actually reload when `reload = FALSE`

Code coverage is high because of an existing test.


jimhester:
LGTM, thanks!

PR #2221, opened 2020-05-05
Optionally open tarball in cran_submit()

krlmlr:
I do this more often than not, works on my system and should be portable thanks to xopen.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -57,7 +57,8 @@ Suggests:
     pkgdown (>= 1.5.1),
     rhub (>= 1.1.1),
     rmarkdown (>= 2.1),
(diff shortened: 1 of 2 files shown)


gaborcsardi:
I would probably do this myself, but I have no idea how many people feels the same way. Maybe we could make it opt-in via an option?

krlmlr:
We have a `yesno()`, we could make it default to `no`. I still think it's better to review before submitting, we could add an opt-out later if people are offended by the extra choice?

hadley:
This seems annoying to me. I have _never_ looked at a tarball prior to submission, and if I did want to look at it, I'd open it myself.

PR #2222, opened 2020-05-05
increase version number for foghorn dependency

fmichonneau:
the latest update to foghorn ensures that the results reported during `release()` are accurate.


jimhester:
Thanks!

PR #2238, opened 2020-06-24
Updates in anticipation of testthat 3.0.0

hadley:


R/utils.R
```diff
@@ -57,3 +57,24 @@ release_bullets <- function() {
...
+#' returns the path to them.
+#'
+#' @param package Name of the test package.
+#' @keywords internal
+#' @examples
+#' if (has_tests()) {
+#' devtest("testUseData")
+#' }
+#' @export
+devtest <- function(package) {
```
jimhester, line 73: Wonder if we should deprecate this...

hadley: Or just remove it? Doesn't look like anyone uses it (https://github.com/search?p=2&q=org%3Acran+devtest&type=Code)

jimhester: yeah, I would be fine with that as well, I'll remove it.

PR #2241, opened 2020-07-01
Added function to increase build version of package

wleoncio:
# Motivation

When an R package maintainer uses [semantic versioning](https://semver.org/), changing the build version of under-development work can be a painfully manual process involving opening the DESCRIPTION file and editing the number there. This may be the reason why many developers avoid using this feature (i.e., build version bumping) which is so useful for organizing the development process.

I've created a function that can transforms this laborious process into a 1-line solution

# Function usage

Assuming the working directory coincides with the root package directory, calling `increase_version()` will bump up the build version, e.g.:

image

(the first argument, file, defaults to "DESCRIPTION", but can point out to another location if need be)

# Disclaimer

- Changes to .Rd files related to other functions are a result of running `devtools::document()` on the latest master commit.
- Package passes tests related to new function, but fails on some unrelated tests
- Package check yields 2 errors, 1 warning and 1 note, all unrelated to the new function.


jimhester:
Thanks for taking the time to work on this. 

However we already use `usethis::use_version()` to handle bumping versions in devtools.

rmflight:
Depending on your use case, versioning by git commit may be useful, in which case you could use these pre or post commit hooks

https://gist.github.com/rmflight/8863882

wleoncio:
I understand. Thank you for your feedback.

PR #2246, opened 2020-07-10
Make yesno() work in noninteractive mode

krlmlr:
and return "yes" by default.


jimhester:
I don't think yesno should work in non-interactive mode, the purpose of the function is to provide friction so that people stop and think about the questions and are sure they have done all the steps.

PR #2250, opened 2020-07-21
Simplify document()

hadley:
Now that we require recent roxygen2


jimhester:
Deleting code, so satisfying!

PR #2255, opened 2020-07-22
Helper cleanup

hadley:


R/utils.R
```diff
@@ -131,74 +42,15 @@ is_attached <- function(pkg = ".") {
...
-
-# throws a warning if the argument is not the current directory.
-warn_unless_current_dir <- function(pkg, envir = parent.frame()) {
-  if (pkg != ".") {
-    warning("`pkg` is not `.`, which is now unsupported.\n  Please use `usethis::proj_set()` to set the project directory.", immediate. = TRUE)
-    local_proj(pkg, .local_envir = envir)
-  }
-}
-
-menu <- function(...) {
```
jimhester, line 155: I think these were here because they were being mocked in the tests, but it seems those tests are no longer used in devtools.

PR #2260, opened 2020-07-30
No-longer export inst()

### LINKED ISSUE #2218: Reconsider removing `inst()`
The [search](https://github.com/search?utf8=%E2%9C%93&q=user%3Acran+%22devtools%3A%3Ainst%22&type=Code) from the original issue #1643 shows four packages using `devtools::inst()` as of now:

 * NlsyLinks, last updated 4 years ago
 * Wats, last updated 5 years ago
 * teachingApps, last updated 2 years ago; [no longer using `devtools::inst()`](https://github.com/cran/teachingApps/blob/master/R/inst.R)
 * IalsaSynthesis, last updated 5 years ago

Since the three remaining packages with an actual dependency only have the dependency in their tests and none of them seem to be actively maintained, I would suggest to remove `devtools::inst()` from the package once again.

The upside of this would be that the auto-complete prefix for `devtools::install()` shortens to "i" instead of "insta", saving a lot of keystrokes for this very frequent function call.
Also, it's less confusing than to have two functions with very similar name but completely different functionality.

hadley:
Fixes #2218

This will break a couple of packages, but they can just use `pkgload::inst()` instead (or `system.file()` in most cases)


jimhester:
Perhaps we should email these maintainers now so that maybe they would update their packages and we could update devtools without manual CRAN intervention...

hadley:
Looks like there are three packages — two are on Github and the last is archived:

* IalsaSynthesis - https://github.com/IALSA/IalsaSynthesis/issues
* NlsyLinks - https://github.com/LiveOak/NlsyLinks/issues
* Wats - archived

I'll file issues.

hadley:
Issues at https://github.com/IALSA/IalsaSynthesis/issues/26 and https://github.com/LiveOak/NlsyLinks/issues/1

jimhester:
Thanks!

PR #2261, opened 2020-07-30
Delete check_failures

hadley:
Not used by any CRAN pacakages: https://github.com/search?q=org%3Acran+check_failures&type=Code


torbjorn:
aw..

I was using it for our automated package pipeline :(

PR #2270, opened 2020-08-24
Need another add = TRUE for on.exit on upload ftp for #2157 fix again

muschellij2:
Need another add = TRUE for `on.exit` on line https://github.com/r-lib/devtools/blob/master/R/check-win.R#L129, otherwise  it resets all the renaming/replacing that is above.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -13,6 +13,9 @@
 
 # devtools 2.3.1
 
+* `check_win_*()` function now resets the email to the original email after
+  execution, this bug was fixed and crept back in (@muschellij2, #2152).
+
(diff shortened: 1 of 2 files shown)


jimhester:
Are you sure? `on.exit()` calls are scoped per environment and there is only one in the body of that `upload_ftp()` function.

muschellij2:
I'm not sure, I'll try to get a reprex, but this current devtools was not
replacing the email.
Best,
John

jimhester:
Adding the add = TRUE won't hurt either way, so I will merge this.

Thanks a million!

PR #2271, opened 2020-08-31
Use rlang::inform() + glue() for messaging

hadley:
After doing this it occurred to me wonder if a cli function would've been better; let me know if you think it would be and I'll update.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -32,7 +32,6 @@ Imports:
     desc (>= 1.2.0),
     DT (>= 0.14),
     ellipsis (>= 0.3.1),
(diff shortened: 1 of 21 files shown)


jimhester:
yeah a cli function would be preferred.

hadley:
Which one?

jimhester:
I guess it depends on the context of the particular message, the `cli_alert_()` family I guess.

hadley:
Oh yeah, and then I can use inline markup and pluralisation as well. I'll take a look later today.

jimhester:
This looks good to me, thanks!

PR #2272, opened 2020-09-01
Redirect callr stderr to stdout

gaborcsardi:
To make sure that they are interleaved correctly.
Fixes https://github.com/r-lib/callr/issues/154


gaborcsardi:
The CI failures seem unrelated.

jimhester:
It is annoying that R is not more consistent on what messages go to stdout and stderr during an install, but it can't be helped :(

jimhester:
Thanks!

PR #2274, opened 2020-09-02
Only document in check if roxygen2 versions match

### LINKED ISSUE #2263: Be less aggressive about documenting in check()
Not sure how to detect this, but when you’re doing a PR to someone else’s repo, it’s annoying that check documents and introduces a bunch of spurious changes. 

Maybe a reasonable heuristic is to only rebuild if roxygen version is the same, otherwise emit a very clear warning.

hadley:
Fixes #2263

### DIFF
--- a/R/check.R
+++ b/R/check.R
@@ -39,9 +39,9 @@
 #'
 #' @return An object containing errors, warnings, and notes.
 #' @template devtools
-#' @param document If `NA` and the package uses roxygen2, will
-#'   rerun [document()] prior to checking. Use `TRUE`
-#'   and `FALSE` to override this default.
+#' @param document By default (`NULL`) will document if your installed
+#'   roxygen2 version matches the version declared in the `DESCRIPTION`
+#'   file. Use `TRUE` or `FALSE` to override the default.
 #' @param build_args Additional arguments passed to `R CMD build`
 #' @param check_dir the directory in which the package is checked
 #'   compatibility. `args = "--output=/foo/bar"` can be used to change the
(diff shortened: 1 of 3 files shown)


R/check.R
```diff
@@ -136,6 +133,28 @@ check <- function(pkg = ".",
...
+    return(FALSE)
+  }
+
+  installed <- packageVersion("roxygen2")
+  if (required != installed) {
+    cli::cli_rule()
+    cli::cli_alert_info(
+      "Installed roxygen2 version ({installed}) doesn't match required version ({required})"
+    )
+    cli::cli_alert_danger("check() will not redocument this package")
```
jimhester, line 149:
```suggestion
    cli::cli_alert_danger("check() will not re-document this package")
```

R/check.R
```diff
@@ -136,6 +133,28 @@ check <- function(pkg = ".",
...
 
+can_document <- function(pkg) {
+  required <- pkg$roxygennote
+  if (is.null(required)) {
+    # Doesn't use roxygen2 at all
+    return(FALSE)
+  }
+
+  installed <- packageVersion("roxygen2")
+  if (required != installed) {
```
jimhester, line 144: Don't know if this would be better, but maybe something like.

```r
same_version <- required == installed`

if(!same_version) {
 # show message
}

same_version
```

hadley: I just tried it and I don't think it's an improvement

jimhester:
Thanks!

PR #2282, opened 2020-10-02
test_files includes full path

hadley:
So extra with_dir not necessary? Not sure what assumptions covr is making, but this made it work again for me.

### DIFF
--- a/R/test.R
+++ b/R/test.R
@@ -92,10 +92,8 @@ test_coverage_file <- function(file = find_active_file(), filter = TRUE, show_re
   reporter$start_file(file, "test")
 
   env <- load_all(pkg$path, quiet = TRUE, export_all = export_all)$env
-  withr::with_dir("tests/testthat", {
-    testthat::with_reporter(reporter, {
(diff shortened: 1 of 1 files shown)


jimhester:
covr doesn't make any assumptions but testthat tests definitly assume the working directory is `tests/testthat`, I don't see how this would work.

hadley:
`test_files` includes the full path. So maybe I should trim that off rather than going in the opposite direction.

jimhester:
So why does it having the full path matter?

hadley:
Because without some change, running `test_coverage_file()` when the R file is open doesn't work for me.

jimhester:
Oh I see, we now set the working directory above with `local_test_directory()`, so yes we should remove the with_dir call.

jimhester:
But your commit message is misleading then, this has nothing to do with the path to the test files being absolute, just that we have already changed the working directory a few lines above.

If you could update the commit message then I can merge this.

hadley:
Oh yeah I was confused too. Can I just squash and merge with a better message?

jimhester:
yeah

PR #2285, opened 2020-10-08
Remove prompt to open package

hadley:
This is the only question you need to routinely answer no to and I find it super annoying


jimhester:
So are you back to using `devtools::release()`?

hadley:
No; but I've always used `submit_cran()`

jimhester:
Oh, I somehow didn't even notice this was in `submit_cran()` and not release.

Issue #2295, opened 2020-11-11
devtools::build() fails to include PDF of help pages when needed


dmurdoch:
The `rgl` package (https:://github.com/rforge/rgl/pkg/rgl) uses some `\Sexpr` expressions in its help files.  CRAN says this means the tarball should contain the PDF file for the help pages, and `R CMD build rgl` includes it.  However, if I ask to build the source package in RStudio, I get a tarball without the help PDF, and CRAN rejects my submission. 

RStudio is calling `devtools::build()`, and after saving files, it calls `pkgbuild::build` with lots of arguments set, including `manual = FALSE`, coming from a default in `devtools::build()`.  This causes `pkgbuild::build` to use the option `--no-manual` when it calls `R CMD build`.

If I run `R CMD build` on a package that doesn't need the PDF file, I don't get one, so I think the default should leave off any specification of the option.

jimhester:
If you use `devtools::submit_cran()` to submit your package it automatically sets `manual = TRUE` when building the package. The defaults for `build()` are designed for local purposes where you don't need a PDF manual or possibly even have LaTeX installed. I empathize with the issue, and wish there was an easier way to specify options to the package builder in RStudio. However I don't think we should change the default.

If you do want to submit your packages manually I would encourage you to call `devtools::build(manual = TRUE)` explicitly from the command line rather than relying on the build button.

dmurdoch:
You don't necessarily get a PDF manual with `manual = TRUE`. The lines that control it are here: https://github.com/wch/r-source/blob/b720f68627df32e974ce7a4b219525d114b81622/src/library/tools/R/build.R#L609.  It's only packages that have help pages that need to execute code that build it.

The issue that LaTeX might not be available is a good point.  Maybe the default should be `manual = pkgbuild::has_latex()`.

As an additional argument, I'm not the only one doing this.  Here's part of a conversation I had with CRAN about this:

> On 11.11.2020 00:04, Duncan Murdoch wrote:
> > > Sorry about that.  I was using the devtools package build; I'll submit a 
> > > bug report there.
> 
> Ah, it is a bug there? We see this quite frequently, but I was not aware 
> it comes from devtools.

jimhester:
Fixed by #2296

PR #2296, opened 2020-11-11
Build manual

dmurdoch:
This simple change fixes the problem in issue #2295 by changing the default to `manual = TRUE`.  The manual is only included when necessary, because `pkgbuild::build` just calls `R CMD build`, which makes the inclusion decision.

### DIFF
--- a/R/pkgbuild.R
+++ b/R/pkgbuild.R
@@ -2,10 +2,13 @@
 #' @param path Path in which to produce package.  If `NULL`, defaults to
 #'   the parent directory of the package.
 #' @inherit pkgbuild::build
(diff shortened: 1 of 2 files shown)


dmurdoch:
Instead of changing the default, I've changed it to add a note to the docs warning about CRAN issues.  A similar note could be added to the docs for `pkgbuild::build`, but that causes Roxygen issues when importing the docs for `devtools::build`:  see https://github.com/r-lib/roxygen2/issues/1135#issuecomment-726801449.

jimhester:
Changing the documentation seems worthwhile, thank you!

Issue #2073, opened 2019-07-07
submit_cran() email prompt - Creator vs Maintainer?


DavisVaughan:
My DESCRIPTION file looks like this:

```
Authors@R: c(
    person("Max", "Kuhn", , "max@email.com", c("aut", "cre")),
    person("Davis", "Vaughan", , "davis@email.com", c("aut")),
    person("RStudio", role = "cph"))
Maintainer: Davis Vaughan <davis@email.com>
```

when I run `submit_cran()`, it asks if it is okay to send the email to `max@email.com`, rather than `davis@email.com`. Is this intended? I figured it would make sense to send the email to the maintainer, but maybe I am wrong. It feels wrong to go in and change me to the `cre` just to get the email to switch to me.

DavisVaughan:
Especially considering when I actually did submit to CRAN it actually sent the email to me (the maintainer)!

DavisVaughan:
Okay upon reading R Packages I see I don't need Maintainer. I removed it and changed the `cre` field.

PR #2304, opened 2020-12-20
Check that the email is actually changed in change_maintainer_email()

emilsjoerup:
Earlier, I was running `check_win_release(email = 'myname@domain.dk')` on a package I am working on. Due to us using both Authors@R and the Maintainer field in our DESCRIPTION file, the report was sent to a different email address. Luckily it was someone I know and it is no problem.

It is bad form to send the reports to unsuspecting people, so I wrote a check for the `change_maintainer_email()` function such that this does not happen again. 

This issue has come up before in #2073 and I suspect that we are not the only people who have errouneously used both authors@R and Maintainer field.

I wrote a unit test for the change, and had to change a DESCRIPTION file or include a new package. I chose to change the testTest package's description file to use both Authors@R and Maintainer. This had no adverse effect on other tests or package checks.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,7 @@
 # devtools (development version)
 
 * Add a check to `change_maintainer_email()` to assess whether the email is actually changed.
-  If the email is not changed, the code now stops such that an email is not accidentally sent to the wrong recipient.
+  If the email is not changed, the code now stops such that an email is not accidentally sent to the wrong recipient. (@emilsjoerup, #2073)
 
 * `check()` only re-documents if you have a matching version of roxygen2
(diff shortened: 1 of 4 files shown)


jimhester:
Can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

emilsjoerup:
Just making sure you get pinged @jimhester

emilsjoerup:
What's the status on this. Do I still need to make changes?

jimhester:
Thanks a bunch!

PR #2325, opened 2021-02-18
Set env vars in test() and test_file()

hadley:
testthat is already setting it in `test_local()` but it seems better to be consistent


jimhester:
Thanks!

PR #2331, opened 2021-03-23
Switch to fs

### LINKED ISSUE #1826: Convert to use fs

malcolmbarrett:
This PR switches from base R file functions to fs. That was mostly a one-to-one switch (although I'm not 100% sure I translated all of the more esoteric base incantations correctly), but I did discover some subtle misbehavior in `clean_vignettes()` and its test expectations. This PR addresses those, as well: `clean_vignette()` now deletes empty `doc` and `Meta` folders 

Closes #1826


malcolmbarrett:
Looks like that mac check failure just farted out while setting up. Not my fault 😇

malcolmbarrett:
I swapped out a use of `file.path(tempdir(), ...)` for `path_temp(...)`, but I did not replace other uses of `tempdir()` as it seemed unnecessary. Happy to fully swap those if you think that's better

malcolmbarrett:
Awesome, thanks! Let me know if I can help with anything else 😄

PR #2347, opened 2021-04-28
Don't open projects with `create()`

### LINKED ISSUE #2346: RStudio New Project wizard opens duplicate sessions
When using RStudio's New Project wizard to create "R package using devtools", it always opens the package in duplicate sessions.
Perhaps the dcf uses `create_package()`, which has `open = rlang::is_interactive()` argument and it causes new session gets opened.

malcolmbarrett:
This PR disables `open` by default to avoid the duplicate sessions opened in the RStudio package creation gui

Closes #2346


jimhester:
Thanks a million!

PR #2350, opened 2021-05-01
Execute the addins in interactive mode

jennybc:
I prefer the user experience this offers. I feel better oriented and informed about what command I actually invoked and how things are going.

And even, morally, it seems fitting to use `Interactive: true` for these:

https://rstudio.github.io/rstudioaddins/#execution-modes

> The `Interactive` field within the addin registration describes whether the addin is interactive or non-interactive. It’s important to understand when an addin should be declared as interactive versus non-interactive, as this effects how RStudio will attempt to execute the addin:
>
> * Interactive addins are invoked by emitting a call to their function directly into the R console. For addins that display user-interface (e.g. using a Shiny application) this enables users to stop/interrupt them.
>
> * Non-interactive addins are run in the background and can not be interrupted, so it’s imperative that these addins complete execution quickly. Otherwise, it’s possible that your addin could freeze the user’s R session.
>
> The use case for non-interactive addins is typically simple text insertion or transformation, in which case users would be annoyed if each invocation resulted in code being inserted into the console.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # devtools (development version)
 
+* RStudio addins now run in interactive mode, rather than background mode (@jennybc, #2350)
+
 * `install(upgrade)` now defaults to 'default' rather than 'ask'. This allows you to control the default asking behavior with the `R_REMOTES_UPGRADE` environment variable (#2345)
 
 * `build_readme()` now uses the `path` argument, as designed (#2344)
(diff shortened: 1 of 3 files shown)


jennybc:
Also added small tweak to the error message. Probably influenced by the boatload of error messages I've written lately, even though I can't style it (yet).

jimhester:
Thank you!

Issue #2358, opened 2021-06-15
Suggestion: Add-in to run examples


pbreheny:
Just a suggestion, but I think it would be nice to offer an Rstudio add-in that allows the user to check all examples for a package with a keyboard shortcut. In other words, add this block to `inst/rstudio/addins.dcf`:

```
Name: Run examples
Description: Runs R code in examples using `devtools::run_examples()`
Binding: run_examples
Interactive: true
```

jimhester:
Sure, do you want to create a pull request to add this?

It would basically be what you have above plus a note in `NEWS.md` mentioning the new add-in

PR #2359, opened 2021-06-15
New addin for run_examples()

pbreheny:
Implementing idea suggested in #2358: An Rstudio addin to run all examples.


jimhester:
Thanks a million!

Issue #2139, opened 2019-10-26
tag ref as instructed by CRAN-RELEASE


randy3k:
I have been using the following function to tag ref as instructed by CRAN-RELEASE.
```
This package was submitted to CRAN on 2019-10-26.
Once it is accepted, delete this file and tag the release (commit 18503ae37c).
```

```r
tag_release <- function(pkg = ".") {
    pkg <- devtools:::as.package(pkg)
    release_file <- file.path(pkg$path, "CRAN-RELEASE")
    if (!file.exists(release_file)) {
        cat("CRAN-RELEASE file not found\n")
        return(invisible())
    }
    line <- readLines(release_file)[2]
    match <- regexec("commit ([a-z0-9]+)", line)
    sha <- regmatches(line, match)[[1]][2]
    if (is.na(sha)) {
        cat("Cannot parse CRAN-RELEASE\n")
        return(invisible())
    }
    semver <- paste0("v", pkg$version)
    if (devtools:::yesno(glue::glue("tag {sha} as {semver}"))) {
        return(invisible())
    }
    tag_sha <- gert::git_tag_create(semver, message = semver, ref = sha, repo = pkg$path)
    file.remove(release_file)
    invisible(tag_sha)
}
```

The code has not yet polished for deployment. Yet, I want to check if devtools devs are interested in it before I create a PR.

randy3k:
One potential problem is that the version number in DESCRIPTION may have changed so we might need to hard code that too in CRAN-RELEASE.

jimhester:
I believe `usethis::use_github_release()` tags the commit along with creating the draft GitHub release, isn't that correct @hadley?

randy3k:
Thanks for the ref to usethis. That feature is more ambitious than the goal of this and that one is github specific. However, as usethis has a similar code, should we move the discussion to usethis?

randy3k:
I see there are two major problems of `use_github_release`. 1. It is github specific, 2. It tags the HEAD ref instead of what is said in `CRAN-RELEASE`. To allow actuary tagging, we need to write both version number and sha to the file.

jimhester:
Yes, it would be great to discuss this in usethis, I think that is the right place for this.

jimhester:
Also I agree that `use_github_release()` should use the commit defined in `CRAN-RELEASE`

randy3k:
I think we still need some changes in devtools. For example, we need to store the tagged version number together with the commit sha in `CRAN-RELEASE`.

PR #2379, opened 2021-10-13
Put more info into CRAN-SUBMISSION and write it for computers, not humans

jennybc:
Relates to #2139 and https://github.com/r-lib/usethis/issues/1380

Example of what CRAN-RELEASE looks like with this PR:

```
DCF: TRUE
submitted_to_cran_on: 2021-10-13
version: 2.4.2
SHA: f45c182b3009e29763d790006b58a1415e5ef1ff
```

as opposed to the existing:

```
This package was submitted to CRAN on 2021-10-13.
Once it is accepted, delete this file and tag the release (commit e10658f5).
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # devtools (development version)
 
-`release()` and `submit_cran()` now record submission details using the Debian Control File format, for better machine-readability. This file now includes package version, in addition to full SHA and a timestamp.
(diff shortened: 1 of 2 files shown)


jennybc:
The `DCF: TRUE` line is a bit gross, but it will help me in usethis. I'm about to release and I will obviously need to support both the old and new formats for now. I want to close https://github.com/r-lib/usethis/issues/1380.

jimhester:
Do you really need `DCF: TRUE`, could we just reorder the fields so version if first and then check for `version:` instead?

jennybc:
I also switched to `Sys.time()`.

What this PR is currently producing:

```
Version: 2.4.2
Date: 2021-10-13 13:09:51
SHA: fbe18b5a22be8ebbb61fa7436e826ba8d7f485a9
```

PR #2387, opened 2021-11-16
Add check_mac_release function to submit to the mac builder

### LINKED ISSUE #2375: `check_mac_devel()`
Like `check_win_devel()` but on CRAN's new m1 macOS builder (https://mac.r-project.org/macbuilder/submit.html)

jimhester:
Fixes #2375


R/check-mac.R
```diff
@@ -0,0 +1,71 @@
...
+#' @param ... Additional arguments passed to [pkgbuild::build()].
+#' @family build functions
+#' @return The url with the check results (invisibly)
+#' @export
+check_mac_release <- function(pkg = ".", args = NULL, manual = TRUE, email = NULL, quiet = FALSE, ...) {
+  check_dots_used(action = getOption("devtools.ellipsis_action", rlang::warn))
+
+  pkg <- as.package(pkg)
+
+  if (!is.null(email)) {
```
jimhester, line 22: The mac builder doesn't seem to use the email or send you results via email, so it is possible we should just remove this functionality for this. What do you think?

R/check-mac.R
```diff
@@ -0,0 +1,71 @@
...
+#' @template devtools
+#' @inheritParams pkgbuild::build
+#' @param email An alternative email to use, default `NULL` uses the package
+#'   Maintainer's email.
+#' @param quiet If `TRUE`, suppresses output.
+#' @param ... Additional arguments passed to [pkgbuild::build()].
+#' @family build functions
+#' @return The url with the check results (invisibly)
+#' @export
+check_mac_release <- function(pkg = ".", args = NULL, manual = TRUE, email = NULL, quiet = FALSE, ...) {
```
jimhester, line 17: Currently it seems the only thing the public API allows is to check on R-release and the M1 macs. Though the response from the POST request indicates future plans to support more versions of R and architectures, e.g.

```
> httr::content(res)
$submitted
[1] TRUE

$id
[1] "1637073137-c5c2e7a4dcbedebf"

$package
[1] "glue"

$version
[1] "1.5.0.9000"

$rflavor
[1] "r-release"

$rfw
[1] "4.1-arm64"

$url
[1] "https://mac.R-project.org/macbuilder/results/1637073137-c5c2e7a4dcbedeb
f/"
```

R/check-mac.R
```diff
@@ -0,0 +1,71 @@
...
+    body = list(
+      pkgfile = httr::upload_file(built_path)
+    ),
+    headers = list(
+      "Content-Type" = "multipart/form-data"
+    ),
+    encode = "multipart"
+  )
+
+  httr::stop_for_status(res)
```
jimhester, line 58: Do you think we need a custom error message here, or is the default error for a bad status sufficient?

R/check-mac.R
```diff
@@ -0,0 +1,71 @@
...
+
+  response_url <- httr::content(res)$url
+
+  if (!quiet) {
+    time <- strftime(Sys.time() + 30 * 60, "%I:%M %p")
+
+    cli::cli_alert_success(
+      "[{Sys.Date()}] Check {.url {response_url}} for a link to results in 15-30 mins (~{time})."
```
jimhester, line 66: This time was based on win-builder times, it seems conservative for the typical mac-builder times, should we leave it or change it?

jimhester:
I haven't yet added support for custom additional dependencies, but this is a first pass at `check_mac_release()`.

R/check-mac.R
hadley: Yeah, just remove I think.

R/check-mac.R
hadley: You could use `task = "Uploading package"` as a compromise

R/check-mac.R
hadley: I think change it.

If the times are pretty short, is it worth adding a polling mode?

jimhester: I guess the other thing is the results page seems to auto-refresh itself, so we could alternatively just use `utils::browseURL()` to open it for the user? (maybe only if it is interactive and !quiet?)

PR #2401, opened 2022-01-07
Fix typo in documentation

karawoo:
Documentation for `uninstall()` references the `remove.package()` function, which should be `remove.packages()`.


jennybc:
Thanks! I made a few other updates, since we're touching this.

PR #2410, opened 2022-02-22
Adjust to new RStudio version numbering

### LINKED ISSUE #2397: `devtools::dev_sitrep()` fails after the change in RStudio version numbering
I was able to check this on 2 separate machines (both running macOS Monterey 12.1). Older machine running RStudio 1.4.1717 and an M1 Mac running RStudio v 2021.9.0.351. 

It looks like `devtools:::check_for_rstudio_updates()` errors with `HTTP status was 400` after the change in version numbering style.

```
devtools::dev_sitrep()
Error in utils::download.file(url, tmp, quiet = TRUE) : 
  cannot open URL 'https://www.rstudio.org/links/check_for_update?version=2021.9.0.351&os=darwin&format=kvp'
In addition: Warning message:
In utils::download.file(url, tmp, quiet = TRUE) :
  cannot open URL 'https://www.rstudio.org/links/check_for_update?version=2021.9.0.351&os=darwin&format=kvp': HTTP status was '400 Bad Request'
Error: [ENOENT] Failed to remove '/var/folders/r9/kvvr87vj4dngktkxksjf_5g80000gn/T/RtmpJhqLSV/filead4862b00739': no such file or directory
```

I was able to replicate this on an Intel Mac running RStudio 2021.9.372 (after updating from v 1.4.1717). `dev_sitrep()` worked fine before the update.
```
devtools::dev_sitrep()
Error in utils::download.file(url, tmp, quiet = TRUE) : 
  cannot open URL 'https://www.rstudio.org/links/check_for_update?version=2021.9.1.372&os=darwin&format=kvp'
In addition: Warning message:
In utils::download.file(url, tmp, quiet = TRUE) :
  cannot open URL 'https://www.rstudio.org/links/check_for_update?version=2021.9.1.372&os=darwin&format=kvp': HTTP status was '400 Bad Request'
Run `rlang::last_error()` to see where the error occurred.
Error: [ENOENT] Failed to remove '/var/folders/wx/zjw1jzk11s76vchhc2xym7xw0000gn/T/RtmpUcW8QW/file1056659aa557b': no such file or directory
Run `rlang::last_error()` to see where the error occurred.
```

Strangely, when trying to create a _reprex_, both reprexes seems to be running fine. 
M1 Mac:
``` r
library(devtools)
#> Loading required package: usethis

devtools::dev_sitrep()
#> ── R ───────────────────────────────────────────────────────────────────────────
#> • version: 4.1.2
#> • path: '/Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/'
#> ── devtools ────────────────────────────────────────────────────────────────────
#> • version: 2.4.3
#> ── dev package ─────────────────────────────────────────────────────────────────
#> • package: <unset>
#> • path: <unset>
#> ✓ All checks passed
```

<sup>Created on 2021-12-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

And older Intel Mac:
``` r
library(devtools)
#> Loading required package: usethis

devtools::dev_sitrep()
#> ── R ───────────────────────────────────────────────────────────────────────────
#> • version: 4.1.1
#> • path: '/Library/Frameworks/R.framework/Versions/4.1/Resources/'
#> • R is out of date (4.1.1 vs 4.1.2)
#> ── devtools ────────────────────────────────────────────────────────────────────
#> • version: 2.4.3.9000
#> • devtools or its dependencies out of date:
#>   'glue', 'xfun', 'knitr', 'pkgbuild', 'openssl'
#>   Update them with `devtools::update_packages("devtools")`
#> ── dev package ─────────────────────────────────────────────────────────────────
#> • package: <unset>
#> • path: <unset>
```

<sup>Created on 2021-12-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

jennybc:
Closes #2397

As per @jgutman advice:

* Use the long version number for RStudio, when available. Fall back to shorter version otherwise.
* URL encode the version number.
* Add `manual=true` to the URL we use when checking for updates.

Also makes this helper much less likely to throw an error, even if it does not succeed.

### DIFF
--- a/R/sitrep.R
+++ b/R/sitrep.R
@@ -15,8 +15,18 @@ check_for_rstudio_updates <- function(os = tolower(Sys.info()[["sysname"]]),
   )
 
   tmp <- file_temp()
-  on.exit(file_delete(tmp))
-  utils::download.file(url, tmp, quiet = TRUE)
+  withr::defer(file_exists(tmp) && nzchar(file_delete(tmp)))
+  suppressWarnings(
+    download_ok <- tryCatch({
+      utils::download.file(url, tmp, quiet = TRUE)
+      TRUE
+    }, error = function(e) FALSE)
+  )
+  if (!download_ok) {
+    # I'll take silent failure here over dev_sitrep() falling over completely
+    # if this download fails
+    return()
+  }
   result <- readLines(tmp, warn = FALSE)
 
   result <- strsplit(result, "&")[[1]]
--- a/R/sitrep.R
+++ b/R/sitrep.R
@@ -2,13 +2,17 @@
 #' @importFrom memoise memoise
 NULL
 
-check_for_rstudio_updates <- function(os = tolower(Sys.info()[["sysname"]]), version = rstudioapi::getVersion(), in_rstudio = rstudioapi::isAvailable()) {
-
+check_for_rstudio_updates <- function(os = tolower(Sys.info()[["sysname"]]),
+                                      version = rstudioapi::versionInfo()$long_version,
+                                      in_rstudio = rstudioapi::isAvailable()) {
   if (!in_rstudio) {
     return()
   }
 
-  url <- sprintf("https://www.rstudio.org/links/check_for_update?version=%s&os=%s&format=%s", version, os, "kvp")
+  url <- sprintf(
+    "https://www.rstudio.org/links/check_for_update?version=%s&os=%s&format=%s",
+    utils::URLencode(version, reserved = TRUE), os, "kvp"
+  )
 
   tmp <- file_temp()
   on.exit(file_delete(tmp))
--- a/R/sitrep.R
+++ b/R/sitrep.R
@@ -2,8 +2,16 @@
 #' @importFrom memoise memoise
 NULL
 
+rstudio_version_string <- function() {
+  if (!rstudioapi::isAvailable()) {
+    return(character())
+  }
+  rvi <- rstudioapi::versionInfo()
+  rvi$long_version %||% as.character(rvi$version)
+}
+
 check_for_rstudio_updates <- function(os = tolower(Sys.info()[["sysname"]]),
-                                      version = rstudioapi::versionInfo()$long_version,
+                                      version = rstudio_version_string(),
                                       in_rstudio = rstudioapi::isAvailable()) {
   if (!in_rstudio) {
     return()
--- a/R/sitrep.R
+++ b/R/sitrep.R
@@ -31,9 +31,9 @@ check_for_rstudio_updates <- function(os = tolower(Sys.info()[["sysname"]]),
     }, error = function(e) FALSE)
   )
   if (!download_ok) {
-    # I'll take silent failure here over dev_sitrep() falling over completely
-    # if this download fails
-    return()
+    return(
+      sprintf("Unable to check for RStudio updates (you're using %s).", version)
+    )
   }
   result <- readLines(tmp, warn = FALSE)
 
--- a/R/sitrep.R
+++ b/R/sitrep.R
@@ -18,7 +18,7 @@ check_for_rstudio_updates <- function(os = tolower(Sys.info()[["sysname"]]),
   }
 
   url <- sprintf(
-    "https://www.rstudio.org/links/check_for_update?version=%s&os=%s&format=%s",
+    "https://www.rstudio.org/links/check_for_update?version=%s&os=%s&format=%s&manual=true",
     utils::URLencode(version, reserved = TRUE), os, "kvp"
   )
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # devtools (development version)
 
+* `dev_sitrep()` has been updated for the calendar-based version number scheme adopted by the RStudio IDE in September 2021 (#2397, #2410).
+
 # devtools 2.4.3
 
 * New `check_mac_release()` function to check a package using the macOS builder at https://mac.r-project.org/macbuilder/submit.html (#2375)
--- a/tests/testthat/_snaps/sitrep.md
+++ b/tests/testthat/_snaps/sitrep.md
@@ -1,7 +1,7 @@
 # check_for_rstudio_updates
 
     Code
-      writeLines(check_for_rstudio_updates("mac", "0.0.1", TRUE))
+      writeLines(check_for_rstudio_updates("darwin", "0.0.1", TRUE))
     Output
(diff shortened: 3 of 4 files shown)


R/sitrep.R
```diff
@@ -2,13 +2,17 @@
 #' @importFrom memoise memoise
 NULL
 
-check_for_rstudio_updates <- function(os = tolower(Sys.info()[["sysname"]]), version = rstudioapi::getVersion(), in_rstudio = rstudioapi::isAvailable()) {
-
+check_for_rstudio_updates <- function(os = tolower(Sys.info()[["sysname"]]),
+                                      version = rstudioapi::versionInfo()$long_version,
```
jennybc, line 6: What's changed:

``` r
rstudioapi::getVersion()
#> [1] '2021.9.1.372'

rstudioapi::versionInfo()$long_version
#> [1] "2021.09.1+372"
```

R/sitrep.R
```diff
@@ -2,13 +2,17 @@
...
+                                      in_rstudio = rstudioapi::isAvailable()) {
   if (!in_rstudio) {
     return()
   }
 
-  url <- sprintf("https://www.rstudio.org/links/check_for_update?version=%s&os=%s&format=%s", version, os, "kvp")
+  url <- sprintf(
+    "https://www.rstudio.org/links/check_for_update?version=%s&os=%s&format=%s",
+    utils::URLencode(version, reserved = TRUE), os, "kvp"
+  )
```
jennybc, line 15: What's new in the URL:

``` r
version <- rstudioapi::versionInfo()$long_version
version
#> [1] "2021.09.1+372"
URLencode(version, reserved = TRUE)
#> [1] "2021.09.1%2B372"
```

jennybc:
Would appreciate a quick :eyes: @romainfrancois and/or @jgutman (I can't officially request a review from you).

jgutman:
This was my advice, but unfortunately I just checked some older RStudio versions, and in those versions we didn't have a long_version. Can we use $long_version if it exists, and otherwise fallback to $version ? I think that should work for all versions, from my initial testing

R/sitrep.R
```diff
@@ -2,17 +2,31 @@
...
 
   tmp <- file_temp()
-  on.exit(file_delete(tmp))
-  utils::download.file(url, tmp, quiet = TRUE)
+  withr::defer(file_exists(tmp) && nzchar(file_delete(tmp)))
```
jennybc, line 18: devtools already uses with a lot, so as I touch things, I intend to keep switching over. I also want to account for the case where the download has failed.

R/sitrep.R
```diff
@@ -2,17 +2,31 @@
...
+  suppressWarnings(
+    download_ok <- tryCatch({
+      utils::download.file(url, tmp, quiet = TRUE)
+      TRUE
+    }, error = function(e) FALSE)
+  )
+  if (!download_ok) {
+    # I'll take silent failure here over dev_sitrep() falling over completely
+    # if this download fails
```
jennybc, line 27: This captures why I've made this function a bit more robust. In general, I think "sitrep" functions should basically never throw an error, because you're often using them to bring clarity to weird, sub-optimal situations.

jennybc:
@jgutman can you give me concrete examples of the different version number forms I need to handle?

R/sitrep.R
jgutman: Two issues I see, unfortunately

1) Older versions of RStudio didn't have a `long_version`, only a `version`. That would be okay, as you could potentially fallback to `version` if `long_version` is not found, but
2) calling URLencode on a version number from `version` rather than `long_version` gives me an error. For example

```
rstudioapi::versionInfo()
$citation

To cite RStudio in publications use:

  RStudio Team (2021). RStudio: Integrated Development Environment for R. RStudio, PBC, Boston, MA URL http://www.rstudio.com/.

A BibTeX entry for LaTeX users is

  @Manual{,
    title = {RStudio: Integrated Development Environment for R},
    author = {{RStudio Team}},
    organization = {RStudio, PBC},
    address = {Boston, MA},
    year = {2021},
    url = {http://www.rstudio.com/},
  }


$mode
[1] "desktop"

$version
[1] ‘1.4.1717’

$release_name
[1] "Juliet Rose"

> URLencode(rstudioapi::versionInfo()$version, reserved=TRUE)
Error in strsplit(URL, "") : non-character argument
```

jgutman (changes requested):

jennybc:
The quoting here is a surprise:

```
$version
[1] ‘1.4.1717’
```

jennybc:
It looks like `rstudioapi::versionInfo()$version` is an instance of `c("package_version", "numeric_version")`, whereas `$long_version` is a simple string.

jennybc:
I'm under the impression that I can't / really shouldn't use `RStudio.Version()` because it's made available in an RStudio context via some sort of shim, whereas it's much safer to always go through rstudioapi.

jgutman:
> @jgutman can you give me concrete examples of the different version number forms I need to handle?

An older version
https://dailies.rstudio.com/rstudio/juliet-rose/desktop/macos/
Returns `$version
[1] ‘1.4.1717’` but no `long_version` available

A newer version
https://dailies.rstudio.com/version/2022.02.0+443/
$version
[1] ‘2022.2.0.444’
$long_version
[1] "2022.02.0+444"

A daily version
$version
[1] ‘2022.6.0.128’
$long_version
[1] "2022.06.0-daily+128"

jgutman:
> I'm under the impression that I can't / really shouldn't use `RStudio.Version()` because it's made available in an RStudio context via some sort of shim, whereas it's much safer to always go through rstudioapi.

Probably right, rstudioapi::versionInfo() will be equivalent!

R/sitrep.R
jgutman: Converting $version to as.character first should work though

jgutman: I tested your function out in RStudio 1.4.1717, and it doesn't return the expected results, but does if I add this to the URL: `&manual=true` after `format=kvp`-- when you click on `Help > Check for Updates` it calls the url with the 
`version={version}&os={os}&format=kvp&manual=true` format, so that may also be required

jennybc:
Here's how the IDE does this:

https://github.com/rstudio/rstudio/blob/main/src/cpp/session/modules/SessionUpdates.R

jgutman (approved):
LGTM

jennybc:
OK this was more than I bargained for, but I think it's done.

PR #2420, opened 2022-04-23
technical edit; added missing single quote character

rmsharp:
The Value section was missing a line of the roxygen2 comment in check_man.Rd


jennybc:
Thanks!

PR #2423, opened 2022-05-20
Mildly improve visual hierarchy of devtools::check()

hadley:
Works best in conjunction with https://github.com/r-lib/pkgbuild/pull/139


gaborcsardi (approved):
Thanks!

PR #2431, opened 2022-05-24
Add linting action

### LINKED ISSUE #2430: Add undesirable function linter
```R
devtools_undesirable_functions <- function() {
  funs <- c(
    # Base messaging
    "message" = "use cli::cli_inform()", 
    "warning" = "use cli::cli_warn()", 
    "stop" = "use cli::cli_abort()",
    # rlang messaging
    "inform" = "use cli::cli_inform()", 
    "warn" = "use cli::cli_warn()", 
    "abort" = "use cli::cli_abort()",
    # older cli 
    "cli_alert_danger" = "use cli::cli_inform()", 
    "cli_alert_info" = "use cli::cli_inform()", 
    "cli_alert_success" = "use cli::cli_inform()", 
    "cli_alert_warning" = "use cli::cli_inform()"
  )
  
  lintr::undesirable_function_linter(
    fun = funs,
    symbol_is_undesirable = FALSE
  )
}
```

hadley:
Fixes #2430

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -25,5 +25,8 @@ jobs:
           packages: r-lib/lintr
 
       - name: Lint
-        run: lintr::lint_package()
+        run: |
+          out <- lintr::lint_package()
+          print(out)
+          if (length(out)) stop("lints found")
         shell: Rscript {0}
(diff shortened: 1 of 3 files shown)


hadley:
@jennybc I think it's slightly better to put the lint config in `.lintr` because then it's also used when you call `lintr::lint_package()` locally.

jennybc:
> @jennybc I think it's slightly better to put the lint config in .lintr because then it's also used when you call lintr::lint_package() locally.

Oh I agree. I think usethis ended up the way it did because I couldn't find a working example of doing such a thing and I got frustrated and inlined the whole thing.

Also based on the usethis experience, I would push a commit introducing an undesirable function, just to make sure it gets caught and reported the way you expect.

PR #2436, opened 2022-06-07
Fix redundant `load_all()` call in `devtools::test()`

lionel-:
Currently I'm seeing:

```
> devtools::test('~/Sync/Projects/R/r-lib/rlang')
ℹ Loading rlang.
ℹ Testing rlang
ℹ Loading rlang.
```

With this PR this becomes:

```
> devtools::test('~/Sync/Projects/R/r-lib/rlang')
ℹ Testing rlang.
ℹ Loading rlang.
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # devtools (development version)
 
+* `test()` no longer calls `load_all()` twice.
+
 * `aspell_env_var()` does a better job of matching R CMD check behaviour,
   which is only to use `aspell`, not `hunspell` or `ispell` (#2376).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # devtools (development version)
 
-* `test()` no longer calls `load_all()` twice.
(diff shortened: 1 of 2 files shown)


jennybc (approved):
I have no idea how well covered this is by tests, but from reading the downstream testthat functions, this change certainly looks fine to me.

I guess we'll be the first to know if anything's wonky 😬

hadley:
FWIW you're seeing this now because of a pkgload bug: https://github.com/r-lib/pkgload/issues/213

hadley (approved):

lionel-:
Thanks for discovering my bug @hadley :)

Regarding the second commit which adds a trailing dot, I now have a doubt. In error messages we recommend ending messages with full stops. But what is our standard/recommendation for other UI alerts? Should they be consistent?

hadley:
It looks weird with a trailing period to me, but that might just be what I'm accustomed too. OTOH these are mostly sentence fragments, not full sentences so maybe `.` isn't needed?

PR #2437, opened 2022-06-08
Unregister package before installing instead of unloading

### LINKED ISSUE #2349: `devtools::install()` crashes with rlang
Because of dangling references. Unregistering the namespace instead of unloading fixes it: https://github.com/r-lib/devtools/compare/master...lionel-:fix-install-crash

lionel-:
Closes #2349


hadley:
Is it worth finishing this off now that pkgload has been released?

PR #2445, opened 2022-07-19
Update roxygen2

hadley:


jennybc:
I had just done this on `main` already (but seeing your diff made me update lintr locally and re-`document()`.

PR #2453, opened 2022-08-01
Remove unused internal utilities

IndrajeetPatil:

PR #2466, opened 2022-08-26
Remove unused internal utilities

IndrajeetPatil:
Follow up to #2453

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -32,6 +32,13 @@ is_testing <- function() {
   identical(Sys.getenv("TESTTHAT"), "true")
 }
 
+release_bullets <- function() {
+  c(
+    '`usethis::use_latest_dependencies(TRUE, "CRAN")`',
+    NULL
+  )
+}
+
 is_rstudio_running <- function() {
   !is_testing() && rstudioapi::isAvailable()
 }
--- a/R/utils.R
+++ b/R/utils.R
@@ -28,15 +28,15 @@ vcapply <- function(x, FUN, ...) {
   vapply(x, FUN, FUN.VALUE = character(1), ...)
 }
 
-is_testing <- function() {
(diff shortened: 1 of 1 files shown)


R/utils.R
```diff
@@ -3,8 +3,6 @@ compact <- function(x) {
   x[!is_empty]
 }
 
-"%||%" <- function(a, b) if (!is.null(a)) a else b
```
IndrajeetPatil, line 6: `{rlang}` exports this operator, and `{rlang}` is a hard dependency of `{devtools}`, so this shouldn't be needed.

R/utils.R
```diff
@@ -30,13 +28,6 @@ vcapply <- function(x, FUN, ...) {
   vapply(x, FUN, FUN.VALUE = character(1), ...)
 }
 
-release_bullets <- function() {
-  c(
-    '`usethis::use_latest_dependencies(TRUE, "CRAN")`',
-    NULL
-  )
-}
-
```
jennybc, line 39: We do NOT want to remove this.

This function is consulted by `devtools::release()` and `usethis::use_release_issue()`.

IndrajeetPatil: Ah, sorry I missed that `{usethis}` is accessing these internal objects from `{devtools}`.

Is this where this function is used?

https://github.com/r-lib/usethis/blob/8f045dc0a88eac82a602df56e2dd62df90971a04/R/release.R#L133-L135

At any rate, I have reverted this change.

R/utils.R
jennybc: But atm devtools does not import rlang's entire namespace, so calls to `%||%` don't actually work with this PR. It's regrettable that the tests don't reveal this, but it's true.

jennybc:
Thanks for this but I'm not doing this sort of housekeeping on devtools atm, so I'm going to close this, since it's not a quick drive-by "merge".

PR #2471, opened 2022-09-20
Set package in `test_active_file()`

### LINKED ISSUE #2470: Set `TESTTHAT_PKG` in `test_active_file()`
Or otherwise figure out how to pass the current package to testthat so it can do it.

hadley:
Fixes #2470

### DIFF
--- a/tests/testthat/test-test.R
+++ b/tests/testthat/test-test.R
@@ -31,8 +31,19 @@ test_that("devtools::test_active_file works", {
   expect_equal(length(out), 1)
 })
 
-test_that("TESTTHAT_PKG environment varaible is set", {
-  test_test(test_path("testTest"), filter = "envvar")
+test_that("TESTTHAT_PKG environment variable is set", {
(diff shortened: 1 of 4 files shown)


tests/testthat/test-test.R
```diff
@@ -31,8 +31,19 @@ test_that("devtools::test_active_file works", {
...
+
+  test_test(
+    test_path("testTest"),
+    filter = "envvar",
+    stop_on_failure = TRUE
+  )
+  test_active_file(
+    test_path("testTest/tests/testthat/test-envvar.R"),
+    stop_on_failure = TRUE
+  )
```
jennybc, line 45: So this is one of those occasions where I would use `expect_without_error()`, but I know that's not how you roll (yet) :slightly_smiling_face: Which explains the mysterious `expect_true(TRUE)`.

I can't pass up this opportunity to keep making the case.

jennybc (approved):

PR #2411, opened 2022-03-01
Add useful development dependencies

hadley:
Fixes #2388 

Along the way I did:

* `use_tidy_description()`
* `use_latest_dependencies(TRUE, "CRAN")`
* Moved packages from suggests to imports without changing which packages will actually be installed by carefully consulting `pak::local_deps_explain()`.

PR #2503, opened 2023-03-07
use_tidy_dependencies

ateucher:
This uses `use_tidy_dependencies()`, which most consequentially adds glue to Imports, and adds brings in `import-standalone-purrr.R`

To start to make use of these I've replaced most instances of `paste0()` with `glue()`. I've also used the purrr-like functionality to remove the DIY `compact()`, replaced `vcapply()` with `map_chr()` and `vapply()` calls with their `map_` equivalent.

The addition of glue takes the number of non-default packages to 21 which triggers a NOTE about excessive dependencies. Although not used, I see that profvis, miniUI, and bench are included in Imports - presumably so the user has them installed so they can use them without fuss? 

Looking at `pak::pkg_deps_explain("devtools", "cli")`, I wonder if cli would be a good candidate to move to Suggests and use `cli::` throughout? It has many dependency pathways to ensure that it will always be installed, and it is easy to qualify with its namespace (and appears to be mostly called that way already in devtools).


R/release.R
```diff
@@ -71,13 +71,13 @@ release <- function(pkg = ".", check = FALSE, args = NULL) {
...
         summary(cran_details)
         cat_rule()
       }
-      cran_url <- paste0(
-        cran_mirror(), "/web/checks/check_results_",
-        pkg$package, ".html"
+      cran_url <- glue(
+        "{cran_mirror()}/web/checks/check_results_",
+        "{pkg$package}.html"
       )
```
jennybc, line 81: I suspect this fits on one line now.

NAMESPACE
```diff
@@ -80,9 +80,11 @@ export(uses_testthat)
 export(wd)
 export(with_debug)
 import(fs)
+import(rlang)
```
jennybc, line 83: This means we could remove lots of `rlang::`.

R/build-manual.R
```diff
@@ -12,7 +12,7 @@ build_manual <- function(pkg = ".", path = NULL) {
   name <- paste0(pkg$package, "_", pkg$version, ".pdf", collapse = " ")
```
jennybc, line 12: Another place for `glue()`? I'm not too sure what the point of `collapse = " "` is here 🤔 it doesn't seem to be doing anything.

R/check-devtools.R
```diff
@@ -28,8 +28,8 @@ check_dev_versions <- function(pkg = ".") {
 
   parsed <- lapply(deps$version, function(x) unlist(numeric_version(x)))
```
jennybc, line 29: Not that it's a big win in this case, but this could become a `map()`.

R/check-win.R
```diff
@@ -86,8 +86,8 @@ check_win <- function(pkg = ".", version = c("R-devel", "R-release", "R-oldrelea
   )
   on.exit(file_delete(built_path), add = TRUE)
 
-  url <- paste0(
-    "ftp://win-builder.r-project.org/", version, "/",
+  url <- glue(
+    "ftp://win-builder.r-project.org/{version}/",
     path_file(built_path)
   )
```
jennybc, line 92: Does this fit on one line now?

jennybc (commented):
Looking good. I made a few comments, mostly at the level of nitpicks.

Once this advances one more click, we can tag Hadley in to discuss the number of dependencies.

On one hand, it would be nice to stay below the threshhold. But on the other hand, given the nature of this package, which is similar to that of the tidyverse meta-package, maybe this just isn't a sustainable position and we should admit it's a meta-package instead of doing weird gymnastics.

R/build-manual.R
ateucher: True enough!

R/check-win.R
ateucher: Well enough I think

NAMESPACE
ateucher: I did this in most places, but there are quite a few instances of:

```r
check_dots_used(action = getOption("devtools.ellipsis_action", rlang::warn))
```

Where I kind of like the clarity using `rlang::`

R/release.R
ateucher: Yup, thanks!

R/check-devtools.R
ateucher: Done, as well as a few others. I also switched to using `~` for anonymous functions in `map*()` calls

ateucher:
Thanks @jennybc! I think I've addressed your comments and suggestions.

Regarding the number of dependencies, I do think it feels a bit messy to sort of obfuscate by moving packages to Suggests to get around an arbitrary threshold. If there's a way to keep packages in Imports that should be there and not feel too much pain with CRAN submissions I think that would be ideal.

R/check-devtools.R
jennybc: I think `~` may not work in the `map_*()` variants offered by the standalone file; that was definitely the case in a previous purrr compat file. You may want to check this, in case the relevant code is not covered by tests.

And, zooming out a bit more, I think we're trying to not create new instances of the `~` style of anonymous function call, as part of the long-term plan to get behind base R's new anonymous function syntax (although we can't use it in packages for a long time).

So, overall, I think it's ok if we use a more verbose method for defining the function that we're mapping.

NAMESPACE
jennybc: Yeah, I agree with you on that specific pattern.

R/check-devtools.R
ateucher: It does work now - I switched to the `~` because I saw this: https://github.com/r-lib/devtools/blob/c9632c423c3e8499d3da94984d2c164702596d4b/R/import-standalone-purrr.R#L34 in the standalone file (and I did test it). But happy to go back to the verbose method!

jennybc: Ah, ok, that changed over time.

But, yeah I think we are trying to not create new usage of the `~`-style syntax. So I think we should do the verbose/explicit method, in the meantime.

ateucher: Sounds good - reverted back to function(x)...

ateucher:
Nice and simple, I like it. If we decide to go that route, should I look in Suggests and move to Imports those packages that are used unconditionally (i.e., without `if (!requireNamespace("pkg")) ...`

jennybc:
> If we decide to go that route, should I look in Suggests and move to Imports those packages that are used unconditionally (i.e., without if (!requireNamespace("pkg")) ...)?

That seems sensible. A good first step would be to determine which packages that applies to, if any, for concreteness.

ateucher:
There are three packages in Suggests that are called unconditionally in code: callr, httr, and rstudioapi.

All have quite strong indirect dependency relationships but I think are probably worth moving into Imports as they are widely used and pretty critical:

``` r
library(pak)

pkg_deps_explain("devtools", deps = "callr")
#> devtools -> pkgbuild -> callr
#> devtools -> pkgdown -> callr
#> devtools -> rcmdcheck -> callr
#> devtools -> rcmdcheck -> pkgbuild -> callr
#> devtools -> testthat -> callr
pkg_deps_explain("devtools", deps = "httr")
#> devtools -> pkgdown -> httr
pkg_deps_explain("devtools", deps = "rstudioapi")
#> devtools -> usethis -> gert -> rstudioapi
#> devtools -> usethis -> rstudioapi
```

<sup>Created on 2023-03-07 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

jennybc:
Yes, so these seem to be the packages that were sort of artificially moved to Suggests in #2411 to avoid the NOTE. I think putting them back in Imports brings this PR to an internally consistent state.

jennybc:
> Should I run use_latest_dependencies() or leave as is? I set glue to be the current CRAN version but haven't touched the others.

Yes, that seems the most appropriate thing to do in this PR where we embrace the meta-package nature of devtools.

> I think I should update NEWS.md with this changes, even though they're not really user-facing?

I'm not entirely sure what we can say that's user-facing, but sure go ahead of add such a bullet and we can decide how it feels.

ateucher:
Ok done - I added a few bullets to NEWS.md. Let me know if it's too much inside baseball!

omparmar23 (approved):

hadley:
With #2671 I think I've now incorporated all these changes (apart from glue, which doesn't seem too important).  Thanks for all your work on this @ateucher!

PR #2536, opened 2023-09-17
Don't pass the package name

### LINKED ISSUE #2535: Getting test coverage for the current package seems to require the package to be installed?
Observation from conf workshop:

Several folks, ~all on Windows I believe~ (~that could be a red herring~ yeah that is not important I can reproduce this), seem to need to install the current package before they can get test coverage for it.

jennybc:
Fixes #2535 

If you pass the package name, testthat tries to determine the edition from the *installed* package. Don't pass the name and testthat will consult `DESCRIPTION` in the *source* package.


jennybc:
I did this very quickly during the workshop and still need to check if there are other unsavory implications of this change.

jennybc:
I now think it's better to resolve this in testthat. I'm going to open an issue there.

PR #2537, opened 2023-09-27
Improve test coverage functionality

hadley:
This was never used correctly because we're passing the path to the package root, instead of the test directory. Additionally, based on https://github.com/hadley/testcoverage, they don't actually appear to be needed.

Fixes r-lib/testthat#1858

---

@jennybc my main scepticism here for the case of `test_coverage_active_file()`, I can't see how this can possibly work (even though it does). As far as I can tell, nothing in `test_coverage_active_file()` causes either `TESTTHAT_PKG` or the edition to be set correctly, but the tests still pass. 

...

Ooooh maybe the problem is that `test_coverage_active_file()` never actually reveals when the test fails.

...

Yes, that was the problem, and once I fixed that, it revealed that the snapshot paths weren't set up correctly, and it helped me understand why `local_test_directory()` is needed here, and why it works. I've added comments so hopefully this is less of a journey in the future.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,7 @@ Imports:
     rversions (>= 2.1.1),
     sessioninfo (>= 1.2.2),
     stats,
-    testthat (>= 3.1.5),
+    testthat (>= 3.1.10.9000),
     tools,
     urlchecker (>= 1.0.1),
     utils,
@@ -68,3 +68,5 @@ Language: en-US
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3
 Config/testthat/edition: 3
+Remotes:  
+    r-lib/testthat
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # devtools (development version)
 
+* `test_coverage()` now works if the package has not been installed.
+
+* `test_coverage_active_file()` now reports if any tests failed and does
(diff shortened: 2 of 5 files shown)


hadley:
@jennybc could you please interactively kick the tires of this PR with https://github.com/hadley/testcoverage and convince yourself that it all works as expected?

jennybc:
Yes, I feel like my tire-kicking of https://github.com/hadley/testcoverage has been successful. 

But as I've used `test_coverage_active_file()`, which apparently I don't do much of in real life, I see this warning, which is coming from covr, I think. I suppose this is a side observation, as I don't think it has anything to do with recent changes in testthat or in this PR.

```
> test_coverage_active_file()
Warning message:
In mapply(FUN = f, ..., SIMPLIFY = FALSE) :
  longer argument not a multiple of length of shorter
```

jennybc (approved):
So, I think all is well here and in testthat, (but suspect there may be some buglet in covr).

jennybc:
I made some other updates on `main` to eliminate some nuisance noise in `R CMD check`, so I merged that in here.

hadley:
🥳 

Do you have a reprex for the recycling warning? I know I saw it multiple times in the workshop, but now I don't.

jennybc:
I am also not seeing the `mapply()` warning any more 😐

PR #2551, opened 2024-01-09
show_news(): Support NEWS in Rd and md formats and under inst

### LINKED ISSUE #2499: `show_news()` only checks `{pkg}/NEWS`
The `show_news()` function is hard-coded to check only for `NEWS` in a package directory:

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Add any other context or about the feature request here.

arcresu:
Fixes #2499.

Exactly mirrors the search path documented in utils::news() for NEWS files rather than using the regex approach proposed in the issue.

The existing code could only handle plain NEWS files. In order to support the alternative formats, I had to mirror what `utils::news` was doing via internal `tools` machinery in `tools:::.build_news_db()`. 

I added some tests partly to convince myself it was working. Perhaps you don't want to take all of these though since it's effectively exercising the internal functions in `tools` more than `devtools` per se.


arcresu:
I discovered that the NEWS-parsing machinery in `tools` skips entries were the version string is malformed according to `.standard_regexps()$valid_package_version`. In particular, the de facto standard first entry of `packagename (development version)` is skipped because `"(development version)"` doesn't match that regex. This is unfortunate, since it's probably exactly the entry you'd want to look at.

If there are no valid entries, the `news_db` builder functions return `NULL` instead of a news database, and forwarding that on to `utils::news(db = NULL)` causes it to display the NEWS for R itself rather than the package. This was the cause of the earlier CI failures in this PR.

In combination, it means that in a new package with a `NEWS.md` generated by `usethis::use_news_md()`, if you call `devtools::show_news()`, you will see an error `Unable to parse NEWS`, since there are no entries with valid version strings. All in all, this PR still improves the state of affairs by supporting all formats and locations of NEWS that R supports.

PR #2558, opened 2024-02-16
Use `purrr::walk()` to yield better error message with `run_examples()`

olivroy:
I will test this to see if it works as expected.

purrr is imported by pkgdown, usethis, profvis, and roxygen2, so doesn't add a dep.


olivroy:
`run_examples()` could be improved, but this doesn't do much!

PR #2562, opened 2024-02-21
Fix small typo in check-mac.R

jgabry:
Adds a missing "for" in the sentence about checking results in 5-10 min.


jennybc:
Thanks!

PR #2574, opened 2024-07-28
Use httr conditionally

### LINKED ISSUE #2573: Use `httr` conditionally
`httr` package is in Suggests but for the function `submit_cran()` is not used conditionally.
Screenshot 2024-07-28 at 00 56 59
I will check if it happens in more functions and submit a PR if useful

catalamarti:
closes #2573


jennybc:
Thanks!

PR #2580, opened 2024-08-29
Fix typo with backticks in `check(vignette)` argument

peterdesmet:
Note: the sentence is 3 characters over 80, but it was already 1 character over initially, so I didn't wrap.


jennybc:
Thanks!

Issue #2563, opened 2024-02-22
Update README reg R Packages second edition


maelle:
:smile_cat:

jennybc:
Closed by 6ec478bbb26645c6fda6d1c518b27e7f478fc91e

PR #2582, opened 2024-09-19
update statements re. pkgs book, update url

luisDVA:
Addresses issue #2563

Updated the readme to reflect the completion of the 2nd edition and the online availability of the 1st edition to avoid confusion.
Also updates the reference to Posit Community.


jennybc:
Thanks!

PR #2598, opened 2025-04-02
Generate dSYM file in `install()` on macOS

lionel-:
This dSYM file contains the file/line information that allows debuggers to figure out the source location.

As noted in the argument doc, generating this file is only one half of the battle. Because the package is built from a temporary location, the source info points to dangling paths. It is possible to remap these paths from lldb though.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -69,4 +69,4 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.2
--- a/NEWS.md
+++ b/NEWS.md
@@ -149,7 +149,7 @@
 * `install_dev_deps()` now passes formal args onto `remotes::install_deps()`
 (@t-gibson, #2166)
 
-* `spell_check()` now checks if `spelling` is installed and prompts users to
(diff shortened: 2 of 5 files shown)


gaborcsardi (approved):
Thanks! I didn't look at the NEWS changes. (Maybe worth undoing the whitespace changes to reduce the noise?) 

Btw. this does not work for packages that use `src/install.libs.R`, unless the package author explicitly copies the debug files. (Which has its own advantages, e.g. avoiding binary packages that are hundreds of megabytes.)

lionel-:
@gaborcsardi Is there any way we could build the package in the original location and bundle it from there? A bit like we do with `load_all()`. This would ensure source info at R and library levels are correct. I looked into `no-staged-install` but that's not it.

lionel-:
> Btw. this does not work for packages that use src/install.libs.R, unless the package author explicitly copies the debug files. (Which has its own advantages, e.g. avoiding binary packages that are hundreds of megabytes.)

Good to know thanks.

I reverted the accidental whitespace changes.

gaborcsardi:
Compiling in the original location is the default for `R CMD INSTALL`, but for `devtools::install()` you need to use `build = FALSE` to get that. 

If you also want to run `R CMD build` first, that copies the while package, and it also creates a (temporary in this case) source R package, so not much you can do there.

PR #2631, opened 2025-12-04
Fix test_coverage_active_file() for testthat 3.3

### LINKED ISSUE #2630: test_coverage_active_file() fails with testthat 3.3
In testthat 3.3, [`local_snapshotter()` gained a new 1st argument, `reporter`](https://github.com/r-lib/testthat/pull/2199). The PR acknowledged that `local_snapshotter()` is used in devtools, but devtools doesn't appear to have an update to deal with the new 1st parameter (and it passes the path unnamed). Therefore, it no longer works:

```r
# I technically had to manually uninstall testthat before this.
devtools::install_version("testthat", "3.2.3")
devtools::test_coverage_active_file()
#> (no error)
```

```r
pak::pak("testthat")
devtools::test_coverage_active_file()
#> Error in reporter$new : $ operator is invalid for atomic vectors
```

I'll PR this momentarily.

jonthegeek:
Name `snap_dir` arg for `testthat::local_snapshotter()` in `test_coverage_active_file()`

Fixes #2630

### DIFF
--- a/R/test.R
+++ b/R/test.R
@@ -119,9 +119,9 @@ test_coverage_file <- function(file = find_active_file(), ...) {
 #' @rdname test
 #' @export
 test_coverage_active_file <- function(file = find_active_file(),
-  filter = TRUE,
-  show_report = interactive(),
-  export_all = TRUE,
+                                      filter = TRUE,
+                                      show_report = interactive(),
+                                      export_all = TRUE,
                                       ...) {
   rlang::check_installed(c("covr", "DT"))
   check_dots_used(action = getOption("devtools.ellipsis_action", rlang::warn))
@@ -143,8 +143,8 @@ test_coverage_active_file <- function(file = find_active_file(),
   )
   snap_reporter$start_file(basename(test_file))
   reporter <- testthat::MultiReporter$new(reporters = list(
-      testthat::StopReporter$new(praise = FALSE),
-      snap_reporter
+    testthat::StopReporter$new(praise = FALSE),
+    snap_reporter
   ))
 
   withr::local_envvar(r_env_vars())


jennybc (commented):
This is sort of nitpicky/meta but sort of legit and I know you are resilient enough @jonthegeek to take it 😁 but can you back out the Air-ish formatting changes, so this only has the meaningful change? I think it's just the addition of `snap_dir =`? 

(I need to do a holistic re-formatting of devtools as a whole. And it really does make it easier to see what's in this PR / what the problem is.)

jonthegeek:
> This is sort of nitpicky/meta but sort of legit and I know you are resilient enough @jonthegeek to take it 😁 but can you back out the Air-ish formatting changes, so this only has the meaningful change? I think it's just the addition of `snap_dir =`?
> 
> (I need to do a holistic re-formatting of devtools as a whole. And it really does make it easier to see what's in this PR / what the problem is.)

I almost did this earlier today because I didn't intentionally make them (air and/or "Strip trailing horizontal whitespace when saving" did). It's a little tricky to save WITHOUT applying those things but I can work it out!

jennybc:
> It's a little tricky to save WITHOUT applying those things but I can work it out!

Hmmm, does that mean you've configured Air formatting on save at the user level? That's kind of anti-recommended, even though it's tempting.

https://posit-dev.github.io/air/editor-vscode.html#user-vs-workspace-settings

> User level settings are automatically applied for all projects that you open. While this sounds nice, if you open an older project (or a project you don’t own) that doesn’t use Air, then you’ll have to remember to turn off your user level Air settings before committing to that project, otherwise you may create a large amount of format related diffs that the project may not want.

In any case, thanks for the fixup! (devtools is about to get Air formatted in a separate PR.)

PR #2649, opened 2026-01-22
Deprecate `install_*()` functions

### LINKED ISSUE #2571: Deprecate `install_*`
And point folks to `pak::pak()` instead

hadley:
Fixes #2571

@jennybc there were no existing tests of these install functions, so no easy way to add snapshot tests for the deprecation messages. So when reviewing I'd suggest reading the new docs (`?"install-deprecated")` and then trying out a couple of functions to check the messages look ok.


jennybc:
What do you think of moving remotes to Suggests? (And making the necessary changes to this PR.)

PR #2651, opened 2026-01-22
Use `pak::local_install_deps()` in `install()`

### LINKED ISSUE #2486: Why is `install()` so keen to update packages?
This requires more analysis, but my current thinking is that `install()` should not, by default, be so eager to update packages.

### LINKED ISSUE #2572: Import pak
I'd claim it's now a key part of the development workflow and we expect folks to have it installed.

hadley:
Includes some general updates to the docs of `install()`, and mild improvements to non-quiet output.

Fixes #2486. Fixes #2572.

@jennybc if you're testing `install()` interactively, it doesn't work on devtools itself, for probably obvious reasons.

### DIFF
--- a/R/install.R
+++ b/R/install.R
@@ -13,8 +13,8 @@
 #' @param quick if `TRUE`, skips some optional steps (e.g. help
 #'   pre-rendering and multi-arch builds) to make installation as fast
 #'   as possible.
-#' @param build if `TRUE` (the default), [pkgbuild::build()]s the package
-#'   first. This ensures that the installation is completely clean, and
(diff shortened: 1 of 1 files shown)


R/install.R
```diff
@@ -41,112 +36,123 @@
...
+    "--install-tests"
+  )
+  if (quick) {
+    opts <- c(opts, "--no-docs", "--no-multiarch", "--no-demo")
+  }
+  opts <- c(opts, args)
+
+  if (build) {
+    if (!quiet) {
+      cli::cat_rule(paste0("Building ", pkg$package), col = "cyan")
```
jennybc: I added this header so it would be more obvious whether the build step is happening or not.

PR #2654, opened 2026-01-22
Optionally report on lack of coverage

### LINKED ISSUE #2632: Zero coverage helper
I'm not sure what to call this, but it's a useful interface for code coverage for an LLM, focussing on what you usually care about: the absence of tests.

```R
zero_coverage <- function(file) {
  test_file <- devtools:::find_test_file(file)

  pkg <- devtools::as.package(".")
  withr::local_envvar(devtools::r_env_vars())

  testthat::local_test_directory(pkg$path, pkg$package)
  reporter <- testthat::local_snapshotter()
  reporter$start_file(test_file, "test")

  env <- devtools::load_all(pkg$path)$env
  testthat::with_reporter(reporter, {
    suppressWarnings(coverage <- covr::environment_coverage(env, test_file))
  })

  coverage_name <- devtools:::name_source(covr::display_name(coverage))
  local_name <- devtools:::name_test(file)
  coverage <- coverage[coverage_name %in% local_name]
  attr(coverage, "relative") <- TRUE
  attr(coverage, "package") <- pkg

  out <- covr::zero_coverage(coverage)
  out$filename <- file.path(
    basename(dirname(out$filename)),
    basename(out$filename)
  )
  out$value <- NULL
  out
}
```

hadley:
And use that display for LLMs and in non-interactive contexts.

It may have been lazy to not deprecate `show_report`, but I just can't imagine anyone is actually using it. Let me know if you think I'm wrong.

Fixes #2632

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,7 +12,7 @@
 * `load_all()` now errors if called recursively, i.e. if you accidentally include a `load_all()` call in one of your R source files (#2617).
 * `release()` is deprecated in favour of `usethis::use_release_issue()`.
 * `show_news()` now looks for NEWS files in the same locations as `utils::news()`: `inst/NEWS.Rd`, `NEWS.md`, `NEWS`, and `inst/NEWS` (@arcresu, #2499).
-* `test_coverage()` and `test_coverage_active_file()` gain a new `report` argument that can be set to `"html"` (the default, for an interactive browser report), `"zero"` (prints uncovered lines to the console, used for LLMs and non-interactive contexts), or `"silent"` (#2632).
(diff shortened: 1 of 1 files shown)


tests/testthat/test-test.R
```diff
@@ -59,3 +59,70 @@ test_that("stop_on_failure defaults to FALSE", {
...
+    report = "zero"
+  ))
+})
+
+test_that("report_default() does its job", {
+  withr::local_options(rlang_interactive = FALSE)
+  expect_equal(report_default(NULL), "zero")
+
+  withr::local_options(rlang_interactive = TRUE)
+  expect_equal(report_default(NULL), "html")
```
jennybc: This test fails when *claude* runs it, because the behaviour of `report_default()` actually changes when `is_llm()` is true 😬

hadley: 😆

PR #2658, opened 2026-01-30
Drop unusued dependencies

### LINKED ISSUE #2657: Why does devtools suggest gmailr?
Maybe we can remove gmailr from `Suggests`?

hadley:
Analysis performed by claude.

Fixes #2657


hadley:
Also pruned the imports a bit.

PR #2662, opened 2026-01-30
Actually drop superseded usethis functions

### LINKED ISSUE #2585: Stop using the deprecated `usethis::ui_*()` functions
Not sure if I want to use `:::` to use the new stuff in usethis (because I can, because I maintain both) or ... just inline / DIY the bits I would need?

hadley:
Fixes #2585

### DIFF
--- a/R/sitrep.R
+++ b/R/sitrep.R
@@ -111,7 +111,7 @@ dev_sitrep <- function(pkg = ".", debug = FALSE) {
     is_windows = is_windows,
     has_build_tools = has_build_tools,
     rtools_path = if (has_build_tools) pkgbuild::rtools_path(),
-    devtools_version = packageVersion("devtools"),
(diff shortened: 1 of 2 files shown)


R/sitrep.R
```diff
@@ -210,19 +228,12 @@ print.dev_sitrep <- function(x, ...) {
...
 # Helpers -----------------------------------------------------------------
 
-hd_line <- function(name) {
-  cat_rule(cli::style_bold(name))
-}
-
 kv_line <- function(key, value, path = FALSE) {
   if (is.null(value)) {
-    value <- cli::col_silver("<unset>")
+    cli::cli_inform(c("*" = "{key}: {.silver <unset>}"))
```
hadley: Also changed this to ensure that all of the output is now "message", not "output".

Issue #2664, opened 2026-01-31
Review `dev_sitrep()`


hadley:
We've had to touch it repeatedly in this round of clean ups, so I think it's worth taking a quick holistic look at it and confirming that it's still fit for purpose.

PR #2668, opened 2026-02-12
Move remotes to suggests

### LINKED ISSUE #2663: Move remotes to Suggests
Follow-on to #2649 (and discussed there). But, more than a bit of faffing around with the previously-re-exported and now-deprecated remotes functions, this really requires removing/guarding all internal use of remotes. Since #2651 makes the opening move towards pak, we'll let the dust settle on various PRs that are already in flight, then come back to this.

hadley:
Fixes #2663


R/sitrep.R
```diff
@@ -228,6 +228,31 @@ print.dev_sitrep <- function(x, ...) {
 
 # Helpers -----------------------------------------------------------------
 
+outdated_deps <- function(deps) {
```
hadley: Minimal code written by Claude. We can consider if anything is needed here in #2664.

PR #2675, opened 2026-02-26
Deprecate build_rmd()

### LINKED ISSUE #2510: `build_rmd()` fails with some output format
`devtools::build_rmd()` is supposed to be a generic function to build any rmd 

Currently it will error on some format like `rmardown::pdf_document()` (which could be used in vignettes too) because of 
https://github.com/r-lib/devtools/blob/daca0cf04413a9d251b3e55f80091afa07a60ab2/R/build-readme.R#L33-L34

````
> devtools::build_rmd("vignettes/demo.Rmd")
ℹ Installing dummyPkg in temporary library
ℹ Building C:/Users/chris/Documents/dummyPkg/vignettes/demo.Rmd
Error in rmarkdown::pdf_document(..., template = template) : 
  argument inutilisé (html_preview = FALSE)
Error:
! ! in callr subprocess.
Caused by error in `rmarkdown::pdf_document(..., template = template)`:
! argument inutilisé (html_preview = FALSE)
````

This is because `html_preview` is not part of `pdf_document()` arguments, and `pdf_document()` does not have an ignore `...` where those unused argument can go. 

We could add this `...` argument, like `rmarkdown::html_document_base()` has, but really this option `html_preview` is only to be used with `github_document()` IMO. So maybe setting this option was supposed to be part of `build_readme()` and not `build_rmd()` ?

### PR DESCRIPTION
**Pull request by jennybc**
Closes #2510

Acting on this comment specifically: https://github.com/r-lib/devtools/issues/2510#issuecomment-3781359906

(This also strangely unblocks me re: the changes in devtools and usethis to support `README.qmd`. By marking `build_rmd()` as "for internal use only", it frees me from having to implement an exported `build_qmd()`, for "symmetry". We just need `build_readme()` to work and the implementation is an internal matter.)

Tagging you @hadley mostly as an FYI.

### CODE REVIEW

File: R/build-readme.R [hunk 1]
```diff
@@ -1,22 +1,46 @@
-#' Build a Rmarkdown files package
```
jennybc: 🤪 this never made sense

R/build-readme.R
```diff
@@ -40,7 +64,9 @@ build_rmd <- function(
   output_options$html_preview <- FALSE
 
   for (path in paths) {
-    cli::cli_inform(c(i = "Building {.path {path}}"))
+    if (!quiet) {
```
jennybc: Unrelated to this PR but discovered while tinkering with tests.

hadley (approved):

PR #2682, opened 2026-03-13
Try to detect sudo calls

jennybc:
devtools's new use of pak can potentially indirectly execute `can_sudo_without_pw()`, which is basically doing some pre-checks in case sysreq work is needed. However testing and documenting devtools would never actually need to do such sysreq tasks. We should never tickle `sudo` this way. The new GHA workflow is intended to make sure this stays fixed.

https://github.com/r-lib/pak/blob/cda86c5ceea65d23f2facb70303347d8e09f4081/src/library/pkgdepends/R/sysreqs.R#L567

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -14,9 +14,12 @@ devtools_default_options <- list(
     options(devtools_default_options[toset])
   }
 
-  # Prevent pak from trying to install system requirements during R CMD check.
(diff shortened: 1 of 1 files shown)


hadley (approved):

jennybc:
Related: I also opened https://github.com/r-lib/pak/issues/850.

PR #2687, opened 2026-04-15
Better reporting around dependencies needed for task

### LINKED ISSUE #2683: `build_readme()` extremely slow due to `pak::local_install_deps()` installing full dependency tree in temp
After the release of devtools 2.5.0, `build_readme()` has become very slow for packages with moderate dependency trees.

`build_readme()` installs the package in a temporary directory https://github.com/r-lib/devtools/blob/ee2ae319579c45cc424fee761437fd161b7cfb41/R/install.R#L256
Because of this, `pak::local_install_deps()` installs the entire dependency tree in temp, even if all dependencies are already installed in the user's normal libraries. This is intended pak behavior, see
https://github.com/r-lib/pak/blob/63d646be4d2d59c29ea02cd99a68fb5b2c4164fd/R/package.R#L15-L17
(and is different from how remotes does it).

For packages with heavier dependency trees, this can add several minutes to `build_readme()`. E.g. for a package I maintain, on my machine it takes ~7 minutes to complete the `pak::local_install_deps()` call (~90 seconds if using Posit Package Manager binaries).

While this is happening, no useful progress information is printed, even with `quiet = FALSE`. The user sees no indication that a large dependency installation is occurring.

jennybc:
Fixes #2683 

The slowness in `build_readme()` was already fixed by never installing dependencies in `local_install()` (07292edc8d475aa43a89f4f21053d36becb68093). This PR is really about making that move safer or less mysterious by surfacing more details around package dependencies that are missing, behind, or ahead, when calling `build_readme() --> local_install()`. That makes use of a bunch of stuff related to `dev_sitrep()`, so `dev_sitrep()` also got some love.

### DIFF
--- a/R/sitrep.R
+++ b/R/sitrep.R
@@ -109,10 +109,12 @@ dev_sitrep <- function(pkg = ".", debug = FALSE) {
     has_build_tools = has_build_tools,
     rtools_path = if (has_build_tools) pkgbuild::rtools_path(),
     devtools_version = utils::packageVersion("devtools"),
-    devtools_deps = compare_deps(pak::pkg_deps("devtools", dependencies = NA)),
-    pkg_deps = if (!is.null(pkg)) {
-      compare_deps(pak::local_dev_deps(pkg$path, dependencies = TRUE))
-    },
+    devtools_cran_version = pak::pkg_deps(
+      "devtools",
+      dependencies = FALSE
+    )$version,
+    devtools_deps = pkg_dep_status("devtools", dependencies = NA),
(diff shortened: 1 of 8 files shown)


R/sitrep.R
```diff
@@ -191,26 +195,28 @@ print.dev_sitrep <- function(x, ...) {
   cli::cli_rule("devtools")
   kv_line("version", x$devtools_version)
 
+  devtools_version <- package_version(x$devtools_version)
+  devtools_cran_version <- package_version(x$devtools_cran_version)
```
jennybc, line 199: Previously we didn't inform folks that, eg, their devtools version was out-of-date. Now we do.

